PR #37, opened 2020-08-06
feature(deadline): Enable logging for DocDB

kozlove-aws:
This fix DocDB part of the issue discovered during the security audit

Was added class that can be used if customer want to switch audit log.

Description of how audit log can be enabled is [here](https://docs.aws.amazon.com/documentdb/latest/developerguide/event-auditing.html)

In repository constructor was added `databaseAuditLogging` parameter and according to value of this parameter enable audit logging for DocDB cluster.

Also was added example code in kitchen sink.

Was enabled parameter `audit_logs` and added `audit` to `enableCloudwatchLogsExports` construct option.

## Testing

Was deployed kitchen sink with this changes and checked that CloudWatch logs enabled 
image
 
Checking CloudWatch that audit log group was created and messages appeared.

### DIFF
--- a/examples/kitchen-sink/lib/index.ts
+++ b/examples/kitchen-sink/lib/index.ts
@@ -5,6 +5,8 @@
 
 import {
   DatabaseCluster,
+  ClusterParameterGroup,
+  CfnDBCluster,
 } from '@aws-cdk/aws-docdb';
 import {
   AmazonLinuxImage,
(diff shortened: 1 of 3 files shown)


ddneilson (changes requested):
> We use CDK construct for DocumentDB so we only can fix it in our example.

I do not believe this statement. Anything that can be done in the application code can be done in the Repository.  The ask, for the Repository, is to design an interface (similar to what Dave did for the ALB) that allows a customer to tell the Repository to enable audit logging when/if the Repository creates a DocDB database for the customer (i.e. if it's not being provided one).

Especially since the thinking is to remove the kitchen sink example from the repo once we have a solid set of official samples in place.

jericht (changes requested):

jericht (changes requested):
Just the comment on the parameter group descriptions then this looks good to me.

jericht (dismissed):
Looks good!

grbartel (commented):
I have 2 more comments for this one:
1. The merge request title needs to be updated since you are adding a feature to the Repository not adding to the example
2. This MR requires Tests for the new functionality that was added to the Repository.

ddneilson (dismissed):

grbartel (changes requested):

grbartel (changes requested):

grbartel (approved):

ddneilson (approved):

PR #77, opened 2020-08-24
fix(core): Remove AcmCertificateImporter race condition

grbartel:
Removes a race condition where certificates would attempt to be deleted while ACM thinks the certificates are still in use.

This bug and following fix mirrors the following merge request https://github.com/aws/aws-cdk/pull/4191 in CDK.

----


horsmand (approved):

ddneilson (approved):
Thanks for spotting and fixing this one, Grant.

PR #87, opened 2020-08-27
fix: allowing empty log group prefixes

ryyakobe:
In some RFDK constructs we allow setting a custom log group prefix. 

If no prefix is given, then we use a default prefix. This is done with the trenary operator `?:`:
`const prefix = props.logGroupProps?.logGroupPrefix ? props.logGroupProps.logGroupPrefix : SOME_DEFAULT_PREFIX;`

So if the prefix is empty `logGroupPrefix = ''`, we will still use the default prefix, instead of using no prefix at all. This PR fixes this issue by replacing the trenary operator `?:` with `??`:
`const prefix = props.logGroupProps?.logGroupPrefix ?? SOME_DEFAULT_PREFIX;`

Also, updated the unit-tests.

----

### DIFF
--- a/packages/aws-rfdk/lib/core/test/mongodb-instance.test.ts
+++ b/packages/aws-rfdk/lib/core/test/mongodb-instance.test.ts
@@ -671,13 +671,15 @@ describe('Test MongoDbInstance', () => {
     }));
   });
 
-  test('is created with correct LogGroup prefix', () => {
+  test.each([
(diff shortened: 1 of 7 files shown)


ddneilson (approved):
Thanks for putting this together, Roman

yashda (approved):
Looks good. Thanks Roman!

PR #90, opened 2020-08-27
feat(bin): Add simpler way to invoke stage-deadline

ddneilson:
This adds a new way to call the `stage-deadline` script. The current URIs are awkward to use, so this change introduces a positional argument, `<deadline_release_version>` that fills in the conventional Deadline release installer & docker URIs with the given release version number. Loosely copied from the command's new usage help:

```
Usage: stage-deadline [--output <output_dir>] [--verbose]
                      <deadline_release_version>

Arguments:
    <deadline_release_version>
        Specifies the official release of Deadline that should be staged. This must be of the form a.b.c.d.
        Both '-d' or '-c' arguments will be ignored if provided with this value.
```

----


packages/aws-rfdk/bin/stage-deadline.ts
```diff
@@ -34,6 +35,11 @@ while (n < args.length) {
       n++;
       dockerRecipesURI = args[n];
       break;
+    case '-r':
+    case '--release':
```
grbartel: release seems like a weird name for this. What do you think of --version instead?

ddneilson: Finding a good name for the option is tricky. `--version`/`-v` is a pretty typical standard option that means "what is the version of the thing that I'm running?" So, in this context, it doesn't fit.

`--release`/`-r` is also a little off because it almost implies that you're doing a release, not acquiring one.

jusiskin: Originally, I had envisioned a positional argument for the version once we had a way to discover and resolve version queries of Deadline.

```
stage-deadline [--output OUTPUT_DIR] [DEADLINE_VERSION]
```

I think it makes it simpler to use. Using it ends up looking like:

```bash
# get latest version
stage-deadline

# or a version specifier
stage-deadline 10.1.9

# or a fully-qualified version
stage-deadline 10.1.9.6
```

For now, since there is no way to resolve versions of Deadline other than fully-qualified patch releases (e.g. `10.1.9.6`), we could fail on things that require look-ups like `10.1.9`.

We could leave the existing usage so we don't break existing users' scripting.

What do you think?

yashda: Can we go with 'deadlineInstallerVersion' so as to match with 'deadlineInstallerURI' ?

ddneilson: 1) I can switch this to an optional positional argument easily enough, I'm not particularly tied to it being a named argument. For now it would just need the fully-qualified version, but that can be relaxed later.

2) How about `deadlineReleaseVersion` for the variable name? It really is more a release version; which happens to match the installer's version too.

yashda: `deadlineReleaseVersion` sounds good to me.

packages/aws-rfdk/bin/stage-deadline.ts
```diff
@@ -49,6 +55,22 @@ while (n < args.length) {
...
 }
 
+// Automatically populate the installer & recipe URI using the version, if it is provided.
+if (deadlineRelease !== '') {
+  if (!/^10\.\d+\.\d+\.\d+$/.test(deadlineRelease)) {
+    usage(1);
+  }
+
+  // installer URI only if not provided as an arg
+  if (deadlineInstallerURI === '') {
```
yashda: Here, both `deadlineInstallerURI` and `dockerRecipesURI` values are taking precedence over deadlineVersion. 
Shouldn't it be otherwise? If not, we should update the usage accordingly for the precedence if both values are provided.

horsmand: I believe the intention here is that you can provide `deadlineReleaseVersion`, but then also provide either of the URI's as well so that the URI's that that would get generated by using `deadlineReleaseVersion` can be overridden.

I do think that this adds some unnecessary complexity to understanding its usage. If we don't see a realistic use case where the user needs to override just one of these URI's, can we just make the usage either `stage-deadline --deadlineReleaseVersion <version>` or `stage-deadline --deadlineInstallerURI <uri> --dockerRecipesURI <uri>` with no combinations?

ddneilson: It's actually what I was going for. I could be clearer in the comments & command documentation, though. My thinking is that you can provide the version, but still use the other options to override the default URIs.

horsmand: I think that since Yash and I both stumbled on why it was like this and it's so unlikely that anyone other than RFDK devs would need to do it, that it would be better to remove this as an option. If you really do want to override one but not the other, it's simple enough to provide both URI's.

That said, if you do think it will be useful, I would also be fine with adding clarification to the documentation.

packages/aws-rfdk/bin/stage-deadline.ts
jusiskin: Thanks for incorporating the suggestion. Resolving

packages/aws-rfdk/bin/stage-deadline.ts
```diff
@@ -179,10 +212,20 @@ The AWS CLI must be configured to authenticate using your AWS account. This can
 See https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html for documentation on how to configure the AWS CLI.
 
 Usage: stage-deadline [--output <output_dir>] [--verbose]
+                      [<deadline_release_version>]
```
jusiskin: Since the version is currently a required argument for this calling pattern, we should remove the outer brackets with signify an optional argument.

jusiskin (changes requested):
All of the input handling looks good. I think there's just one minor fix to be made to the usage help output.

packages/aws-rfdk/bin/stage-deadline.ts
jusiskin: It looks like you've chosen to make these invocation patterns mutually-exclusive. Either choice would have been reasonable, but this simplification will make it easier to maintain this command as we don't have to be concerned about the combined use-case. The trade-off is that users will have to always specify both URIs, even if they only want to override one of them. Resolving.

packages/aws-rfdk/bin/stage-deadline.ts
yashda: fixed.

jusiskin (approved):
Looks great - approved!

jericht (approved):
LGTM

PR #100, opened 2020-08-28
fix(deadline): adding deadline version check for workers

yashda:
While configuring deadline workers if the installed deadline
client version is less than the minimum supported deadline version
then we now fail during deployment.

Also updating the deadline readme.

Verified by installing a higher and lower versions for both
windows and linux.

----


horsmand (dismissed):

packages/aws-rfdk/lib/deadline/README.md
```diff
@@ -6,6 +6,12 @@ The `aws-rfdk/deadline` sub-module contains Deadline-specific constructs that ca
 import * as deadline from 'aws-rfdk/deadline';
 ```
 
+---
+
+_**Note:** Current version of RFDK supports constructs for Deadline 10.1.9 and later._
```
jusiskin: Just a minor reword here:

```suggestion
_**Note:** RFDK constructs currently support Deadline 10.1.9 and later._
```

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
```diff
@@ -36,6 +36,26 @@ if [ ! -f "$DEADLINE_COMMAND" ]; then
     exit 1
 fi
 
+isVersionLessThan() {
+    local RESULT=$(python -c "import sys;print(tuple(map(int,\"$1\".split('.'))) < tuple(map(int, \"$2\".split('.'))))")
+    case $RESULT in
+        True) echo true ;;
+        False) echo false ;;
+        *) echo "ERROR: Unknown boolean value \"$RESULT\"" 1>&2; exit 1 ;;
+    esac
```
jusiskin: Suggestion:

1.  Rather than parsing stdout, could we just use the exit-code?
1.  Rather than inlining the versions operands into the python script, can we pass them as arguments? This will help avoid escaping issues.

Putting that together:

```suggestion
    python -c "import sys;sys.exit(0 if tuple(map(int, sys.argv[-2].split('.'))) < tuple(map(int, sys.argv[-1].split('.'))) else 1)" "$1" "$2"
```

using this function would become:

```bash
if isVersionLessThan 10.1.9 10.1.10; then
  echo "less than!"
fi
```

jusiskin (changes requested):
Just a few minor suggestions, but otherwise this will be great and could save a lot of potential headaches. Great stuff!

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
yashda: updated.

packages/aws-rfdk/lib/deadline/README.md
yashda: updated.

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
```diff
@@ -36,6 +36,21 @@ if [ ! -f "$DEADLINE_COMMAND" ]; then
...
+    python -c "import sys;sys.exit(0 if tuple(map(int, sys.argv[-2].split('.'))) < tuple(map(int, sys.argv[-1].split('.'))) else 1)" "$1" "$2"
+}
+
+DEADLINE_VERSION=$("$DEADLINE_COMMAND" -Version | grep -oP '[v]\K\d+\.\d+\.\d+\.\d+\b')
+if [ -z "$DEADLINE_VERSION" ]; then
+    echo "ERROR: Unable to identify the version of installed Deadline Client. Exiting..."
+    exit 1
+fi
+MINIMUM_SUPPORTED_DEADLINE_VERSION="10.1.9"
+if $(isVersionLessThan $DEADLINE_VERSION $MINIMUM_SUPPORTED_DEADLINE_VERSION); then
```
jusiskin: I believe [command substitution](https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html#Command-Substitution) or output expansion (`$(...)`) is not what we want here. Instead we want to check the exit code with:

```suggestion
if isVersionLessThan $DEADLINE_VERSION $MINIMUM_SUPPORTED_DEADLINE_VERSION; then
```

From [conditional constructs](https://www.gnu.org/software/bash/manual/html_node/Conditional-Constructs.html#Conditional-Constructs):

>   The syntax of the if command is:
>
>       if test-commands; then
>           consequent-commands;
>       [elif more-test-commands; then
>           more-consequents;]
>       [else alternate-consequents;]
>       fi
>
>   The _test-commands_ list is executed, and if its return status is zero, the _consequent-commands_ list is executed. If _test-commands_ returns a non-zero status, each `elif` list is executed in turn, and if its exit status is zero, the corresponding _more-consequents_ is executed and the command completes. If ‘`else alternate-consequents`’ is present, and the final command in the final `if` or `elif` clause has a non-zero exit status, then _alternate-consequents_ is executed. The return status is the exit status of the last command executed, or zero if no condition tested true.

jusiskin (changes requested):
Looking good. One last suggested change

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
yashda: fixed.

jusiskin: Thanks. resolving

packages/aws-rfdk/lib/deadline/lib/worker-fleet.ts
```diff
@@ -182,13 +183,17 @@ export interface WorkerInstanceFleetProps {
    * Health Monitor component to monitor the health of instances.
    *
    * @default - Health Monitoring is turned-off
+   *
+   * Note: The health-check feature is supported with Deadline Client v10.1.9 and later.
```
jusiskin: I'm not 100% positive, but it is possible this note might get parsed as part of the `@default` TSDoc block. Here's what I know:

Unfortunately, TSDoc has no official spec yet, but the [TSDoc GitHub README](https://github.com/microsoft/tsdoc#why-do-we-need-tsdoc) states:
>   a syntax that is loosely based on JSDoc

The [JSDoc spec](https://jsdoc.app/about-block-inline-tags.html) distinguishes two types of tags: inline and block. Inline uses the `{@tag ...}` syntax and block uses the start of a comment line like:

```ts
/**
 * @blockTag ...
 */
```

The JSDoc spec also states:

>    Each block tag must be followed by a line break, **with the exception of the last block tag in a JSDoc comment**.

There is a `@remarks` block tag that we could use for this. See the [TSDoc playground](https://microsoft.github.io/tsdoc/). This same feedback applies to the other "Note:"s added in this PR.

jusiskin (changes requested):
This is just about ready. Just a minor question about the TSDoc warning added.

packages/aws-rfdk/lib/deadline/lib/worker-fleet.ts
yashda: Nice catch! I tried @remarks tag, but that doesn't work. Verified that we use JSII docs to generate the docs website and remarks tag is not supported in JSII docs yet. Hence moving the note to the main comment block for better readability.

jusiskin: Great - thanks for investigating that. Resolved

jusiskin (approved):
Approved

horsmand (approved):

PR #101, opened 2020-08-28
chore(integ): rewrite end-to-end test to simplify deployment, use fal…

aws-painec:
…lback values, eliminate output file

----
This update makes several changes to the end-to-end integration test to simplify deployment. In one sense, not much changes; you still set variables in `test-config.sh` and then run the script. But the variables now have fallback values so the test will still run even if none of them are set. The logic breaks out like this:

* If DEADLINE_STAGING_PATH is provided by user, the Deadline version found at that path will be used for the test, overriding any value supplied for DEADLINE_VERSION
* If DEADLINE_VERSION is supplied by user, the script will automatically stage Deadline assets for that version in `integ` to use for the test
* Whichever way the version is supplied, the script will automatically pull AWS Portal AMI Ids for that version in the necessary region to use for Deadline workers, but this can be overridden by supplying the AMI ids in `test-config`
* The only value that doesn't have a fallback value is USER_ACCEPTS_SSPL_FOR_RFDK_TESTS, which still must affirmatively be set by the user

This commit also eliminates the output file created by the test and all outputs now go to STDOUT


integ/README.md
```diff
@@ -6,9 +6,9 @@ To run all test suites:
...
 1. Configure AWS credentials (tests will use the default AWS profile, so either set up a default profile in .aws/credentials or use temporary credentials).
 
-1. If your AWS profile has a different default region fromw here the test app will be deployed, set the AWS_REGION variable to the desired region.
+1. Set the environment variable CDK_DEFAULT_REGION to the region the test should be deployed in
 
-1. Configure test-config.sh. This script sets environment variables which are necessary for the tests. Currently this includes:
+1. Configure test-config.sh. This script configures which test modules will run and overrides certain default values. Currently these include:
     * Options required for all Deadline test components:
       * DEADLINE_VERSION - version of the Deadline repository installer used for the test
       * DEADLINE_STAGING_PATH - Complete path to local staging folder for Deadline assets (see `packages/aws-rfdk/docs/DockerImageRecipes.md` for more information)
```
ryyakobe, line 14: As now most of this options have default values, maybe we could move `USER_ACCEPTS_SSPL_FOR_RFDK_TESTS`  to be the first in the list, as it is the most important variable now.

integ/test-config.sh
```diff
@@ -3,16 +3,27 @@
...
 
+# Qualified path to staged Deadline assets. 
+#   - If not set here, assets will stage automatically to `aws-rfdk/integ/stage`
+#   - If set here, the version found in `manifest.json` at this path will override any value supplied for DEADLINE_VERSION
+export DEADLINE_STAGING_PATH
+
 # Options for Deadline Repository test component
+# Accept SSPL license to install MongoDB
+#   - If not set to `true` here, the Repository test component will fail immediately on execution
 export USER_ACCEPTS_SSPL_FOR_RFDK_TESTS=false
```
ryyakobe: I would recommend moving this variable to the very top, as all other variables are optional and have default values.

integ/README.md
ryyakobe, line 14: As now most of our variables have default values, maybe we could move `USER_ACCEPTS_SSPL_FOR_RFDK_TESTS`  to be the first in the list, as it is the most important variable now.
Maybe we could even split this list into `Required` and `Optional`.

integ/components/deadline/workerFleet/scripts/bash/e2e.sh
```diff
@@ -5,19 +5,26 @@
...
-  npm run test deadline-workerFleet >> "$OUTPUT_FILE" 2>&1
-  echo "Test suite complete. Destroying test app..." >> "$OUTPUT_FILE"
-  npx cdk destroy "*" -f
-  rm -f "./cdk.context.json"
-  rm -rf "./cdk.out"
-  echo "Test app destroyed." >> "$OUTPUT_FILE"
-  echo "Deadline Worker tests complete." >> "$OUTPUT_FILE"
+  if [[ $OPTION != '--destroy-only' ]]; then
+    deploy_component_stacks "Deadline WorkerInstanceFleet"
+    execute_component_test "deadline-workerFleet"
```
ryyakobe: Just an idea: maybe we could have `workerFleet`, `renderQueue`, and `repository` as variables, because we use them here but also in 

`if [ $(basename $(pwd)) != "workerFleet" ]; then`

so if we ever decide to rename the folder name or the test, we'll have to change it just in one place. 

What do you think?

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -11,55 +11,113 @@
...
+        cd "$INTEG_ROOT/$COMPONENT_ROOT" && "./scripts/bash/e2e.sh" || cleanup_on_failure
     fi
 done
 
 # Destroy the infrastructure stack on completion
-echo "Test suites completed. Destroying infrastructure stack..." >> "$OUTPUT_FILE"
-cd "$infrastructure"
+echo "Test suites completed. Destroying infrastructure stack..."
+cd "$INFRASTRUCTURE_APP"
 npx cdk destroy "*" -f
```
ryyakobe: I think we could also add here ` || cleanup_on_failure`, because destruction could also fail, and maybe on the second run it will be successful. At worst case we still get a failing destruction, but at least we did a best effort.

Or are there any risks I'm not aware of?

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -11,55 +11,113 @@
...
-echo "Starting RFDK-integ end-to-end tests" > "$OUTPUT_FILE"
+# Check if Deadline assets are staged at configured path
+if [ ! $(ls "$DEADLINE_STAGING_PATH/manifest.json" 2> /dev/null) ]; then
+    # Stage Deadline assets
+    npx --package=aws-rfdk@$RFDK_VERSION stage-deadline --deadlineInstallerURI "s3://thinkbox-installers/Deadline/$DEADLINE_VERSION/Linux/DeadlineClient-$DEADLINE_VERSION-linux-x64-installer.run" --dockerRecipesURI "s3://thinkbox-installers/DeadlineDocker/$DEADLINE_VERSION/DeadlineDocker-$DEADLINE_VERSION.tar.gz" --output "$DEADLINE_STAGING_PATH"
+fi
 
-# Run preflight checks to make sure necessary variables, etc. are set
-jest --passWithNoTests --silent "preflight"
+# If executing worker fleet tests, find Deadline AMIs based on supplied version
```
ryyakobe, line 45: General comment on this file:
This script is quite large as for shell script. Maybe we could split it into a few functions or even smaller scripts? For example, to have preparation (setting default values on environment variables, staging, etc.) separated from running the tests.

What do you think?

aws-painec:
Rebased to mainline and repushed. This adds an additional commit to:

- Make the output of the individual test components at the end of the test more orderly
- Add timers to measure how long the deploy, test, and teardown stages of each test take
- Renames the directories the test components are in so that the top-level test runner script runs them in order (repository, then renderQueue, then workerFleet). This renamed some directories so that creates a ton of file changes

integ/test-config.sh
aws-painec: Done, it's now the first thing in the file

integ/components/deadline/workerFleet/scripts/bash/e2e.sh
aws-painec: I agree. Since one of my updates is actually renaming those folders this is a good opportunity to add that fix.

integ/scripts/bash/rfdk-integ-e2e.sh
aws-painec: I don't think so. It's easy enough to add and hopefully it'll just never come up. Fixed.

integ/scripts/bash/rfdk-integ-e2e.sh
aws-painec: That's probably a good idea, it's gotten kind of unwieldy. I'll see what I can do.

integ/components/deadline/deadline_01_repository/scripts/bash/e2e.sh
```diff
@@ -0,0 +1,30 @@
...
+# Load utility functions
+source "../common/functions/deploy-utils.sh"
+
+if [ ! "${SKIP_DEADLINE_REPOSITORY_TEST-}" == true ]; then
+  
+  if [[ $OPTION != '--destroy-only' ]]; then
+    deploy_component_stacks "Deadline Repository"
+    execute_component_test "deadline_01_repository"
+  fi
+  if [[ $OPTION != '--deploy-only' ]]; then
```
grbartel: You have a flag here for "Deploy only" however this flag results in deploying and executing the tests.  which goes against the naming.

grbartel (changes requested):
Taking a step back on how these are all approached.  

Every e2e.sh script is the exact same with the exception of which directory they are being run in and the pretty names of the tests.  as such can we change these all into a shared script that uses the directory name for the test.  It would mean changing the validation that we are in the correct directory but it would reduce the duplicate code.

integ/components/deadline/deadline_01_repository/scripts/bash/e2e.sh
aws-painec: For making a single e2e.sh: yes, this is especially true now that I've reworked the component name to be the same variable passed into all those functions. I'll make that change.

Renamed that flag to "--deploy-and-test-only" for clarity

aws-painec:
OK, addressing your last feedback, I've turned the different component e2e scripts into a single generic script. It's also now possible to run the entire test package without needing to alter `test-config.sh` at all by running this command:
`USER_ACCEPTS_SSPL_FOR_RFDK_TESTS=true SKIP_TEST_CONFIG=true yarn e2e`

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -6,60 +6,105 @@
...
+export TEST_START_TIME="$(date +%s)"
+SECONDS=$TEST_START_TIME
 
-echo "Loading config..."
-source "${root}/test-config.sh"
+export INTEG_ROOT="$(pwd)"
+BASH_SCRIPTS="$INTEG_ROOT/scripts/bash"
+
+# RFDK version is passed in as first argument (taken from package.json)
+export RFDK_VERSION=${1-0.17}
```
ryyakobe: I wonder if it makes sense to default to some RFDK version if it's not provided? Maybe we should just exit here with error. 

As I understand, this may happen if the person manually called `rfdk-integ-e2e.sh` instead of using `yarn run e2e`. But if the latest RFDK version was `0.18`, that person will run integration tests for 1 hour and might not even understand, that tests were not on the latest version.

What do you think?

integ/components/deadline/common/scripts/bash/component_e2e.sh
```diff
@@ -0,0 +1,32 @@
...
+OPTION=${2:-undefined}
+
+if [[ $(basename $(pwd)) != $COMPONENT_NAME ]]; then
+  echo "ERROR: Script must be run from top directory of test component"
+  exit 1
+fi
+
+SKIP_TEST_CHECK=\$SKIP_${COMPONENT_NAME}_TEST
+SKIP_TEST_CHECK=$(eval "echo $SKIP_TEST_CHECK" 2> /dev/null) || SKIP_TEST_CHECK=false
+if [[ ! "${SKIP_TEST_CHECK}" == "true" ]]; then
```
ryyakobe: I guess `==` and `=` are equal in `bash`, but in other `POSIX` shells only `=` is guaranteed to work, so I think we should go with a safer `=` option. Please correct me if I'm wrong.

There are multiple places in this PR where this could be changed.

ryyakobe (commented):
The only little concern I still have is that it's possible to call `component_e2e.sh` with both `--deploy-and-test-only` and `--destroy-only` and then the order will define which is used.

But I also want all these amazing changes to be merged as soon as possible, so I think we can address this in the future separately if we ever need to.

integ/scripts/bash/rfdk-integ-e2e.sh
aws-painec: yeah that sounds right, it's not too much of a pain to expect the user to supply RFDK_VERSION if they are running the script directly (and there's the added benefit that running it through the node script guarantees the latest version). A very easy fix.

integ/components/deadline/common/scripts/bash/component_e2e.sh
aws-painec: I wasn't aware of that but as far as I know `=` should work just as well; I'll make a quick pass to fix these.

ryyakobe (dismissed):
Amazing work! Thanks, Claire!

ryyakobe (dismissed):
Approved. Thanks again!

aws-painec:
> The only little concern I still have is that it's possible to call `component_e2e.sh` with both `--deploy-and-test-only` and `--destroy-only` and then the order will define which is used.
> 

I think for the most part this isn't a huge issue right now, the component_e2e isn't really mean to be run on its own outside of the top-level `rfdk-integ-e2e` runner. Going forward I'll probably need to find a better way to configure which test components to run or run individual test components solo, but that's beyond the scope of this change I think.

integ/scripts/bash/deploy-infrastructure.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Deploy the infrastructure app, a cdk app containing only a VPC to be supplied to the following tests
+INFRASTRUCTURE_APP="$INTEG_ROOT/components/_infrastructure"
+cd "$INFRASTRUCTURE_APP"
+echo "Deploying RFDK-integ infrastructure..."
+npx cdk deploy "*" --require-approval=never || yarn run tear-down
+echo "RFDK-integ infrastructure deployed."
+cd "$INTEG_ROOT"
+
+exit 0
```
grbartel: Missing a new line here

grbartel (commented):
1 Minor nit otherwise it all looks good.

integ/scripts/bash/deploy-infrastructure.sh
aws-painec: There's always one! I've fixed it, thanks for catching :)

grbartel (approved):
Looks good. Thanks for the hard work Claire

ryyakobe (approved):

PR #104, opened 2020-08-28
chore(deps): Bulk update of all dependencies

ddneilson:
There's some tangling between our dependencies that make them tricky for Dependabot to update on its own. This is a bulk update of all of our deps; mostly to address the security issue with `dot-prop`.

The more notable updates are:
* Updating CDK to 1.61.1 -- this entailed some minor code changes to fix tests, a change away from the deprecated `Tag.add` interface (see: https://github.com/aws/aws-cdk/pull/9558 ), and implementing a new property of the `IResource` interface (see: https://github.com/aws/aws-cdk/pull/8280 ).
* Removing our awslint copy in favour of the official package.

----


grbartel (dismissed):
Looks good to me

examples/deadline/All-In-AWS-Infrastructure-Basic/python/setup.py
```diff
@@ -17,8 +17,8 @@
     packages=setuptools.find_packages(where="package"),
 
     install_requires=[
-        "aws-cdk.core==1.57.0",
-        "aws-rfdk==0.16.0"
+        "aws-cdk.core==1.61.1",
+        "aws-rfdk==0.17.0"
```
jusiskin: Not sure about this....

Version 0.17.0 of RFDK uses CDK 1.57.0, right?

jusiskin (changes requested):
Great job. Just one small thing I'm not sure about in the python sample app.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/setup.py
ddneilson: Matches the Typescript version of the example. Hrm... that's a pickle. I guess we'll have to leave the examples on the last release, and bump them (including CDK version) on release.

integ will have to be different. It's intended to be used with the tip-of-mainline.

grbartel (approved):

ddneilson:
Note: I had to remove building our example app from the `build.sh` script. There's a conflict in CDK dependency versions between release RFDK & development RFDK. We want the examples to match release RFDK so that customers can use them as-is.

jusiskin (approved):

PR #109, opened 2020-08-30
fix(deadline): adding version check for staging deadline

yashda:
Adding minimum supported deadline version check while staging
deadline installers. Now we fail fast if the provided version
is not supported by RFDK.

----


packages/aws-rfdk/lib/core/lib/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+   *
+   * @param versionFirst
+   * @param versionSecond
+   *
+   * @returns negative value if first version is smaller than second version;
+   * 0 if the versions matches, positive value if first version is smaller
+   * than second version.
+   */
+  public static versionCompare(versionFirst: string, versionSecond: string): number {
+    const regExStripZero = /(\.0+)+$/;
```
ryyakobe: I wonder if we should handle cases with a trailing dot: `10.1.0.`.
Not sure if we need it, but something like `(\.0+.+)+$` will do the trick.

What do you think?

packages/aws-rfdk/lib/core/lib/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export class Utils {
+  /**
+   * This method compares two version strings
+   *
+   * @param versionFirst
+   * @param versionSecond
+   *
+   * @returns negative value if first version is smaller than second version;
+   * 0 if the versions matches, positive value if first version is smaller
```
ryyakobe: There is an error in the description. These 2 statements contradict:

***negative*** value if first version is smaller than second version
***positive***  value if first version is smaller than second version.
```suggestion
   * 0 if the versions matches, positive value if first version is greater
```

packages/aws-rfdk/lib/core/lib/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+   * than second version.
+   */
+  public static versionCompare(versionFirst: string, versionSecond: string): number {
+    const regExStripZero = /(\.0+)+$/;
+    const versionArrayFirst = versionFirst.replace(regExStripZero, '').split('.');
+    const versionArraySecond = versionSecond.replace(regExStripZero, '').split('.');
+    const l = Math.min(versionArrayFirst.length, versionArraySecond.length);
+
+    for (let i = 0; i < l; i++) {
+      let diff = parseInt(versionArrayFirst[i], 10) - parseInt(versionArraySecond[i], 10);
```
ryyakobe: Neat: I guess you should be able to use const here.
```suggestion
      const diff = parseInt(versionArrayFirst[i], 10) - parseInt(versionArraySecond[i], 10);
```

packages/aws-rfdk/lib/deadline/lib/stage.ts
```diff
@@ -154,6 +157,12 @@ export class Stage {
       throw new TypeError(`Expected a string "version" but got: ${typeof version}`);
     }
 
+    // Do minimum supported deadline version check
+    const minimumSupportedVersion = '10.1.9';
```
ryyakobe: Souldn't we check for `10.1.9.2` here? We state in our documentation that we support this version, but this looks to work with `10.1.9.0`.

ryyakobe (changes requested):

packages/aws-rfdk/lib/core/lib/utils.ts
yashda: The method 'versionCompare' assumes that the version strings are already formatted. I added another validation for the version format. So, if the version string itself is incorrect, we will throw an error accordingly. This method will continue to just compare the formatted version strings.

packages/aws-rfdk/lib/deadline/lib/stage.ts
yashda: '2' is the only patch that was released under '10.1.9'. So its safe to compare it with '10.1.9' in the code.

packages/aws-rfdk/lib/core/lib/utils.ts
yashda: fixed. good catch.

packages/aws-rfdk/lib/core/lib/utils.ts
yashda: updated.

packages/aws-rfdk/lib/core/lib/utils.ts
```diff
@@ -0,0 +1,54 @@
+/**
```
jusiskin: Two suggestions:

1.  Rather than starting a `utils.ts` file, what would you think about renaming this file to `version.ts`? It would be good to avoid the common pitfall of having a growing utils module that becomes a collection of miscellaneous code.
1.  How would you feel about making this object-oriented? I'm thinking a `Version` class would be nice, where instances could be constructed in two ways - either by specifying the numeric components or parsed from a string:

    ```ts
    class Version {
    public readonly components: number[];

    constructor(...components: number[]) {
        this.components = components;
    }

    public isGreaterThan(other: Version): boolean {
        // code here
    }

    public static parse(versionText: string): Version {
        // code to parse a version string
        return new Version(...components);
    }
    }
    ```

Putting this together, it makes the API and internal code a bit more intuitive/readable:

```ts
import { Version } from './version';

const versionA = new Version(10, 1, 9, 2);
const versionB = Version.parse('10.1.9.3');
if (versionB.isGreaterThan(versionA)) {
  console.log('B is greater than A');
}
```

What do you think?

packages/aws-rfdk/lib/core/lib/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Regex: ^\d+(?:\.\d+){3}$
+     * Matches a sequence of '.' separated numbers with exactly 4 digits.
+     * - ^ asserts position at start of a line.
+     * - \d+ Matches one or more digits.
+     * - (?:\.\d+) Matches a dot and the following one or more digits.
+     * - * Matches previous pattern zero or more times.
+     * - $ asserts position at the end of a line
+     */
+    if (version.match(/^\d+(?:\.\d+)*$/g)) {
```
jusiskin: nit: we don't require the `/g` flag here, since we are anchoring to the beginning (`^`) and end (`$`) of the string.

packages/aws-rfdk/lib/core/lib/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+    const versionArraySecond = versionSecond.replace(regExStripZero, '').split('.');
+    const minLen = Math.min(versionArrayFirst.length, versionArraySecond.length);
+
+    for (let i = 0; i < minLen; i++) {
+      const diff = parseInt(versionArrayFirst[i], 10) - parseInt(versionArraySecond[i], 10);
+      if (diff) {
+        return diff;
+      }
+    }
+    return versionArrayFirst.length - versionArraySecond.length;
```
jusiskin: This function can currently be invoked as:

```ts
const result = versionCompare('10.1', '10.1.9');    // result will be -1
```

Is this the expected behavior? I suppose it depends on the intended semantic meaning when passing `"10.1"` as an argument to this function? Right now it is being treated as an implicit `"10.1.0"`, but I wonder if some users might consider `"10.1"` to mean the latest `10.1.x.y` release.

We cannot resolve versions (`10.1.x.y`) at deploy-time, so perhaps we should throw an exception if the two versions don't have the same number of components when being passed into this function. Callers can add `".0"` if need-be. What do you think?

jusiskin (changes requested):
This looks great. My main suggestion is around code organization, but there is another concern about input validation.

packages/aws-rfdk/lib/core/lib/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+export class Utils {
+  /**
+   * This method compares two version strings
+   *
+   * @param versionFirst
+   * @param versionSecond
+   *
+   * @returns negative value if first version is smaller than second version;
```
ryyakobe: Neat. Maybe:
```suggestion
   * @returns a negative value if the first version is smaller than the second version;
   * 0 if the versions match;
   * a positive value if the first version is greater than the second version.
```

ryyakobe (commented):
Looks good! Thanks, Yash! Do you think we could change a docstring just a bit more?

packages/aws-rfdk/lib/core/lib/utils.ts
jusiskin: +1. I'd also suggest replacing "smaller" with "less than"

packages/aws-rfdk/lib/core/lib/utils.ts
yashda: sounds good. I have modified it accordingly. Also removed a duplicate version parsing logic.

packages/aws-rfdk/lib/core/lib/utils.ts
yashda: Modified the code to only accept full, resolved version string. We can add another method to parse partial, unresoled version string once we add that logic.

packages/aws-rfdk/lib/core/lib/utils.ts
yashda: updated.

packages/aws-rfdk/lib/deadline/lib/stage.ts
```diff
@@ -154,6 +159,12 @@ export class Stage {
       throw new TypeError(`Expected a string "version" but got: ${typeof version}`);
     }
 
+    // Do minimum supported deadline version check
+    const minimumSupportedVersion = new Version([10, 1, 9, 2]);
+    if (minimumSupportedVersion.isGreaterThan(Version.parse(version))) {
```
ryyakobe, line 164: I feel it's a bit confusing to use a function called `isGreaterThan`, when the message says `is less than`.
You have to reread the code to understand what is really happening. 

Could we have both of these statements match? I guess `<` is used more often than `>`, so I would suggest changing the function to `isLessThan` and use
`if (manifestVersion.isLessThan(minimumSupportedVersion)) {`

What do you think?

ryyakobe: I feel it's a bit confusing to use a function called `isGreaterThan`, when the message says `is less than`.
You have to reread the code to understand what is really happening. 

Could we have both of these statements match? I guess `<` is used more often than `>`, so I would suggest changing the function to `isLessThan` and use
`if (manifestVersion.isLessThan(minimumSupportedVersion)) {`

What do you think?

yashda: Added another method 'isLessThan' to the Version class.

packages/aws-rfdk/lib/core/lib/utils.ts
yashda: fixed.

ryyakobe (dismissed):
Great work! Thanks Yash!

packages/aws-rfdk/lib/deadline/lib/stage.ts
```diff
@@ -154,6 +159,13 @@ export class Stage {
       throw new TypeError(`Expected a string "version" but got: ${typeof version}`);
     }
 
+    // Do minimum supported deadline version check
+    const minimumSupportedVersion = new Version([10, 1, 9, 2]);
```
jusiskin: #100 introduces a similar constant inline in code with this same minimum Deadline version. Are these minimums linked or logically separate?

Even if separate, I wonder if there's a single place we can specify this value that is easy to find later. This makes it easy if we ever need to change our minimum supported version later. What do you think? I'd also be open to creating a new issue for it for later.

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -72,6 +72,106 @@ abstract class VersionQueryBase extends Construct implements IVersion {
...
+    return false;
+  }
+
+  /**
+   * Numeric components of version.
+   */
+  public readonly components: number[];
+
+  constructor(components: number[]) {
+    this.components = components;
```
jusiskin, line 165: Should we add some input validation here as well and the corresponding tests? We are accepting arbitrary number of components and negative numbers.

packages/aws-rfdk/lib/deadline/test/version.test.ts
```diff
@@ -13,11 +13,153 @@ import {
...
+
+    // WHEN
+    const lhs = new Version([10, 0, 9, 2]);
+    const result = lhs.isLessThan(Version.parse('10.1.9.2'));
+
+    expect(result).toEqual(true);
+  });
+
+  describe('.parse', () => {
+    test.each<[string, { version: string, expectedValue: boolean }]>([
```
jusiskin: How would you feel about renaming `expectedValue` to `expectException` to be a bit more specific and readable? The values would need to be inverted.

packages/aws-rfdk/lib/deadline/test/version.test.ts
```diff
@@ -13,11 +13,153 @@ import {
...
+      // WHEN
+      let result = false;
+      try {
+        Version.parse(version);
+        result = true;
+      } catch (error) {
+      } finally {
+      }
+
+      expect(result).toEqual(expectedValue);
```
jusiskin: nit: while this test is technically valid, jest has some nice helpers for asserting exceptions being thrown that cleans up the `try`/`catch` block:

```ts
// Function to call
function parse() {
  Version.parse(version);
}

if (expectException) {
  expect(parse).toThrow();
}
else {
  expect(parse).not.toThrow();
}

```

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -72,6 +72,106 @@ abstract class VersionQueryBase extends Construct implements IVersion {
   }
 }
 
+/**
+  * This class is reposonsible to do basic operations on version format.
+  */
+export class Version {
```
jusiskin: It looks like this class could implement the `IPatchVersion` interface here. We could leverage [TypeScript accessors](https://www.typescriptlang.org/docs/handbook/classes.html#accessors) for implementing the individual components, e.g.:

```ts
export class Version {
  // ...
  public get majorVersion(): number {
    return this.components[0];
  }
}
```

This could simplify the code in `VersionQuery.parseVersionString`.

jusiskin (changes requested):
This is looking great. I have a few more suggestions. Let me know what you think.

packages/aws-rfdk/lib/deadline/lib/stage.ts
yashda: Agree. I have now modified the code such that the minimum supported deadline version will now be passed as a parameter to the UserData. This makes #100 dependent on this PR.

packages/aws-rfdk/lib/deadline/lib/version.ts
yashda: Added.

packages/aws-rfdk/lib/deadline/test/version.test.ts
yashda: Updated.

packages/aws-rfdk/lib/deadline/test/version.test.ts
yashda: updated.

packages/aws-rfdk/lib/deadline/lib/version.ts
yashda: updated.

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -116,9 +121,47 @@ export class Version {
...
+   * @inheritdoc
+   */
+  public get patchVersion(): number {
+    return this.components[3];
+  }
 
   constructor(components: number[]) {
+    // validations
+    if(components.length != 4) {
+      throw new TypeError('Invalid version format. Version should contain exactly 4 components.');
```
jusiskin: Minor suggestion here:

The TypeScript compiler and JSII enforce correct typing here, so rather than throwing a `TypeError`, should we just throw a generic `Error` class for this.

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -116,9 +121,47 @@ export class Version {
...
+    }
+    components.forEach((component) => {
+      if (component < 0) {
+        throw new TypeError('Invalid version format. None of the version components can be negative.');
```
jusiskin: Do you think [`RangeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError) fits better here?

packages/aws-rfdk/lib/deadline/lib/worker-fleet.ts
```diff
@@ -548,6 +549,7 @@ export class WorkerInstanceFleet extends WorkerInstanceFleetBase {
         `'${groups}'`,
         `'${pools}'`,
         `'${props.region || ''}'`,
+        `'${Version.MINIMUM_SUPPORTED_DEADLINE_VERSION.toString()}'`,
```
jusiskin: This and the changes in:

*   `packages/aws-rfdk/lib/deadline/scripts/powershell/configureWorker.ps1`
*   `packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts`

probably belong in #100 since they are not used yet.

jusiskin (changes requested):
Last few changes - but this is otherwise ready.

jusiskin (dismissed):
Approved!

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -72,6 +72,149 @@ abstract class VersionQueryBase extends Construct implements IVersion {
...
+
+  /**
+   * This method compares two version strings
+   *
+   * @param version
+   *
+   * @returns true if this version is less than the provided version;
+   * false if this version is greater than or equal to the provided verison.
+   */
+  public isLessThan(version: Version): boolean {
```
ryyakobe, line 188: There is some code duplication between `isLessThan` and `isGreaterthan`.
I think in some earlier iteration, you already used an integer as return type in `Utils.versionCompare`.
Maybe we can use a something like below to remove the code duplication and to have all 3 comparisons available:

```
private static compare(first: Version, second: Version): number {
  if (first.components.length != second.components.length) {
    throw new TypeError('Component count in both the versions should be same.');
  }

  for (let i = 0; i < first.components.length; i++) {
    const diff = first.components[i] - second.components[i];
    if (diff != 0) {
      return diff;
    }
  }
  return 0;
}


public isGreaterThan(version: Version): boolean {
  return compare(this, version) == 1;
}

public isGreaterThan(version: Version): boolean {
  return compare(this, version) == -1;
}

public isEqual(version: Version): boolean {
  return compare(this, version) == 0;
}
```

yashda: updated

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -72,6 +72,158 @@ abstract class VersionQueryBase extends Construct implements IVersion {
...
+  /**
+   * This method compares two version strings
+   *
+   * @param version
+   *
+   * @returns true if this version is greater than the provided version;
+   * false if this version is less than or equal to the provided verison.
+   */
+  public isGreaterThan(version: Version): boolean {
+    if (this.components.length != version.components.length) {
```
ryyakobe: Is there a reason why can't we move this check inside of the `compare` function?

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -72,6 +72,158 @@ abstract class VersionQueryBase extends Construct implements IVersion {
...
+  /**
+   * This method compares 2 versions.
+   *
+   * @param version version to compare
+   *
+   * @returns negative value if this version is less than the provided version;
+   * 0 if both the versions are equal;
+   * positive value if this version is greater than the provided verison.
+   */
+  private compare(version: Version): number {
```
ryyakobe, line 220: We still don't have an easy way to compare for version equality.
Could you please add `isEqualTo` method similar to `isGreaterThan` and `isLessThan`?

ryyakobe (changes requested):

packages/aws-rfdk/lib/deadline/lib/version.ts
yashda: added.

packages/aws-rfdk/lib/deadline/lib/version.ts
yashda: updated.

ryyakobe (approved):
Great work! Thanks again, Yash!

jusiskin (approved):
Approved! Looks great!

PR #118, opened 2020-09-04
chore(integ): refactor testing tier class hierarchy for component tests

aws-painec:
----
Removes the individual `testing-tier` files for each test component in favor on a single class definition in the `integ/lib` directory. The test bastions are now built using repeatable functions to make maintaining this code and adding new components for testing much easier. This results in no functional change to how the tests are run.

There is also a small functional fix to the `deploy-all` and `tear-down` bash scripts to make them run with the same modes and options as the end to end test. As a note, these scripts are for examining test stacks without running the full integration tests and are intended for troubleshooting and debug only.


integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,340 @@
...
+    });
+
+    this.configureBastionUserData({
+      installDeadlineClient: false,
+      fetchDocdbCert: true,
+      testingScriptPath: '../components/deadline/deadline_01_repository/scripts/bastion/testing',
+    });
+
+    this.configureMongodb();
+  }
```
jusiskin: It appears that this method and the two below it are test-specific. Instead of putting the logic in this common Stack, what if each test sub-classed this stack? This way we get to re-use the common code, but separate the test-specific code. The logic in this method could be moved into the constructor of the repository test's sub-class. To illustrate:

```ts
import {
  TestingTier,
  TestingTierProps,
} from '../../../../lib/testing-tier'

export interface RepositoryTestingTierProps extends TestingTierProps {
  readonly structs: Array<StorageStruct>;
}

export class RepositoryTestingTier extends TestingTeir {
  constructor(scope: Construct, id: string, props: RepositoryTestingTierProps) {
    super(scope, id, props);

    // Move configureRepositoryTest code into here
  }
}
```

jusiskin (changes requested):
Great work removing duplicate code here! I have a suggestion though to help improve the organization a little bit. I'll wait to see what you think before reviewing further.

integ/lib/testing-tier.ts
aws-painec: I think that makes a lot of sense, I'll take a look at revising this today.

aws-painec:
Pushed a new revision that re-adds files that for per-component testing tiers, which are now class extensions of the base testing tier in `integ/lib`

Just a warning that the tests are currently failing, due apparently to an unrelated issue with Cloudwatch Agent, which is still being investigated

aws-painec:
I've confirmed that all other tests except those affected by the cloudwatch issue run successfully. I did note that at least the WorkerFleet test can run up against jest's timeout limit and be in danger of failing. Since the tests make so many calls and stuff I decided to take care of it by adding one more update that ups the timeout limit on all of the jest tests to 60 seconds.

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,268 @@
...
+import { BastionHostLinux, InstanceType, Port, SubnetType, Vpc } from '@aws-cdk/aws-ec2';
+import { FileSystem } from '@aws-cdk/aws-efs';
+import { ILogGroup } from '@aws-cdk/aws-logs';
+import { Asset } from '@aws-cdk/aws-s3-assets';
+import { CfnOutput, Construct, Duration, Stack, StackProps } from '@aws-cdk/core';
+import { X509CertificatePem, MongoDbInstaller, MongoDbSsplLicenseAcceptance, MongoDbVersion } from 'aws-rfdk';
+import { IWorkerFleet, RenderQueue, Repository } from 'aws-rfdk/deadline';
+import { IRenderFarmDb } from './storage-struct';
+
+// Interface for choosing userData installation options
```
jusiskin: It appears `integ` directory is not subject to the same linting rules that enforce a common code style to the rest of the repository code.

While fixing that issue is definitely not in scope for this particular pull request, how would you feel about changing the comments (where applicable) in this file to conform to [TSDoc](https://github.com/microsoft/tsdoc) syntax? This is more important for common code files like this one containing a shared base class since it makes the documentation discoverable in code editors and IDEs that have TypeScript integration. For example, in VS Code, you can hover your mouse over a method and see its TSDoc string in a tooltip. So for example:

```suggestion
/**
 * Interface for choosing userData installation options
 */
```

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,268 @@
...
+  fetchDocdbCert: boolean;
+  testingScriptPath: string;
+}
+
+// Params object for TestingTier
+export interface TestingTierProps extends StackProps {
+  integStackTag: string;
+}
+
+// Class constructor
```
jusiskin: 🤔  This comment is not technically accurate nor would it be necessary if it were.

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,268 @@
...
+export class TestingTier extends Stack {
+
+  private testInstance: BastionHostLinux;
+  private deadlineVersion: string = process.env.DEADLINE_VERSION!.toString();
+  private stagePath: string = process.env.DEADLINE_STAGING_PATH!.toString();
+
+  constructor(scope: Construct, id: string, props: TestingTierProps) {
+    super(scope, id, props);
+
+    // Collect environment variables
```
jusiskin: Looks like this comment got left behind while refactoring.

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,268 @@
...
+  public configureDatabase(testSuiteId:string, database:IRenderFarmDb) {
+    const db = database.db;
+    const dbSecret = database.secret!;
+
+    this.testInstance.connections.allowTo(db, Port.tcp(27017));
+    dbSecret.grantRead(this.testInstance);
+
+    new CfnOutput(this, 'secretARN' + testSuiteId, {
+      value: dbSecret.secretArn,
+    });
```
jusiskin: Is `secretARN` descriptive enough? It looks like there are a few secrets being output by this stack. How much effort would it take to make this more explicit?

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  // Allows the bastion to connect to the repository's EFS file system
+  configureEfs(efs:FileSystem) {
+    this.testInstance.connections.allowToDefaultPort(efs);
+  }
+
+  // Configures the local instance of mongo DB on the bastion for reading from the repository
```
jusiskin: Unless I'm missing something, based on my understanding of the code below, this method only adds userdata commands to install MongoDB on the instance and doesn't point it to the repository.

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,268 @@
...
+    const logGroupName = logGroup.logGroupName;
+
+    repo.fileSystem.mountToLinuxInstance(this.testInstance.instance, {
+      location: '/mnt/efs/fs' + testSuiteId.toLowerCase(),
+    });
+
+    new CfnOutput(this, 'logGroupName' + testSuiteId, {
+      value: logGroupName,
+    });
+  }
```
jusiskin: This method is only used by the `RepositoryTestingTier`. What do you think about moving it to a private method in that class to keep the code closer to where it is being used? If/when its needed by another test in the future, we can move it up into the parent class.

Same applies to:
*   `configureMongodb` method above for the `RepositoryTestingTier`
*   `configureWorkerFleet` method below for the `WorkerFleetTestingTier`.

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    new CfnOutput(this, 'secretARN' + testSuiteId, {
+      value: dbSecret.secretArn,
+    });
+  }
+
+  // Allows the bastion to connect to the repository's EFS file system
+  configureEfs(efs:FileSystem) {
+    this.testInstance.connections.allowToDefaultPort(efs);
+  }
```
jusiskin: Since this is only called once by the `RepositoryTestingTier` and it only contains a line of code, I wonder if it should just be inlined instead. I believe it would make the calling code more readable/maintainable. What do you think?

jusiskin (changes requested):
Finished my second pass and again found a few mostly organizational changes. Let me know what you think.

integ/lib/testing-tier.ts
aws-painec: > It appears integ directory is not subject to the same linting rules that enforce a common code style to the rest of the repository code.

Is it? The integ files are included when I run the `build` scripts and it gives the linting errors for missing semicolons, misaligned whitespace, etc. What else is supposed to be included?

Either way I'm find making another pass to clean up those comments.

integ/lib/testing-tier.ts
jusiskin: Much better! Resolved.

integ/lib/testing-tier.ts
jusiskin: I believe we don't run `awslint` which means some linting is skipped. That linter has a rule to add TSDocs to all public exported classes/methods/interfaces/attribuites.

The [`integ/package.json` uses a different `"build"` script](https://github.com/aws/aws-rfdk/blob/75d0f9fedce52de9dcfe73482965de923fa8941d/integ/package.json#L44). Our main package at [`packages/aws-rfdk/package.json` uses `cdk-build`](https://github.com/aws/aws-rfdk/blob/75d0f9fedce52de9dcfe73482965de923fa8941d/packages/aws-rfdk/package.json#L46) which explains it.

integ/lib/testing-tier.ts
aws-painec: Removed as part of the comment cleanup

integ/lib/testing-tier.ts
aws-painec: Yup, removed

integ/lib/testing-tier.ts
aws-painec: It's a fairly minor change. Renamed to `DatabaseSecretARN`

integ/lib/testing-tier.ts
aws-painec: I'll edit the comment for clarity

integ/lib/testing-tier.ts
aws-painec: Yeah I kept them all here for convenience's sake if they're needed for other classes in the future, but if it's best practice to move them to the subclasses that's easy enough to implement. Will change.

integ/lib/testing-tier.ts
aws-painec: Yeah, probably, I don't like it when too many functions clog up the code. Will fix.

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,283 @@
...
+      'cd ~ec2-user',
+      'mkdir -p utilScripts',
+      'cd utilScripts',
+      `unzip ${utilZipPath}`,
+      'chmod +x *.sh',
+      `rm -f ${utilZipPath}`,
+    );
+
+    const testingScripts = new Asset(this, 'TestingScripts', {
+      path: path.join(__dirname, props.testingScriptPath),
```
jusiskin: It is slightly confusing to have the caller of this function have to specify relative paths with-respect-to the source code path of the called function. This is an artifact of my request to create the files with the sub-classed stacks deeper in the directory structure - sorry 🤷 .

What if instead, the caller provides the absolute path, or alternatively a pre-constructed `Asset` instance?

integ/lib/testing-tier.ts
```diff
@@ -0,0 +1,283 @@
...
+        bucketKey: clientInstaller.s3ObjectKey,
+      });
+
+      userDataCommands.push(
+        `cp ${installerPath} ./deadline-client-installer.run`,
+        'chmod +x *.run',
+        './install_deadline_client.sh',
+        `rm -f ${installerPath}`,
+      );
+    }
```
jusiskin: Rather than putting these behind props flags, what if we made this a separate method? That way they can be called by the subclasses that require them. So for example, this method could be `installDeadlineClient()`. Same goes for `fetchDocdbCert`, except that one is only used by `RepositoryTestingTier` and could be moved into a method in that class.

Once this is done, we have removed all branches from this method that is being called by all of the sub-classes. There remains a single argument `testingScriptPath`. At this point, we could make the `testingScriptPath`  (or an equivalent asset - see other PR discussion) a construtor prop of the `TestingTier` base class and have the `configureBastionUserData` method called by the constructor. This removes the need to call it in the sub-classes. So in summary, here's the proposed change:

```ts
export interface TestingTierProps extends StackProps {
  // ...

  /**
   * Local path to the framework directory containing the testing script to be copied to the Bastion
   */
  readonly testingScriptPath: string;

  // OR

  /**
   * Asset of a ZIP archive containing the directory of testing scripts to be copied to the Bastion
   */
  readonly testingScripts: Asset;
}

export abstract class TestingTier extends Stack {
  constructor(scope: Construct, id: string, props: TestingTierProps) {
    this.configureBastionUserData(props.testingScriptPath);

    // OR

    this.configureBastionUserData(props.testingScripts);
  }

  protected installDeadlineClient(): void {
    // ...
  }

  private configureBastionUserData(testingScriptPath: string) {
    // ...
  }

  // OR

  private configureBastionUserData(testingScripts: Asset) {
    // ...
  }
}
```

and in `integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts`:

```ts
export class RepositoryTestingTier extends TestingTier {
  constructor(scope: Construct, id: string, props: RepositoryTestingTierProps) {
    // ...

    this.fetchDocdbCert();
  }

  private fetchDocdbCert(): void {
    // ...
  }
```

integ/lib/testing-tier.ts
jusiskin: Much better. Thanks!

integ/lib/testing-tier.ts
jusiskin: thanks! resolved

integ/lib/testing-tier.ts
jusiskin: Looks good now - resolved

integ/lib/testing-tier.ts
jusiskin: Looks like this one is still around

aws-painec: Fixed, thanks for catching.

integ/lib/testing-tier.ts
aws-painec: Agreed. Changed it so the call takes in the full path instead of relative.

integ/lib/testing-tier.ts
aws-painec: Rerunning the tests to confirm but I have this worked out with separate functions for configure the cert and deadline user data. `configureBastionUserData` now takes the full path to the scripts directory.

aws-painec:
Repushed to address Josh's feedback. I'm running the full test to confirm everything's still working as expected and will update when it finishes.

aws-painec:
I've confirmed all tests are still passing with these changes.

jusiskin (dismissed):
Thanks for your patience as always @aws-painec. This code is an order of magnitude clearer and more maintainable than before this PR. Great work!

integ/components/deadline/deadline_01_repository/bin/deadline_01_repository.ts
```diff
@@ -16,20 +16,25 @@ const env = {
...
-  useMongoDB: false,
-});
-const storage3 = new StorageStruct(componentTier, 'StorageStruct3', {
-  integStackTag,
-  provideDocdbEfs: false,
-  useMongoDB: true,
-});
 
-new TestingTier(app, 'RFDKInteg-DL-TestingTier' + integStackTag, { env, integStackTag, structs: [storage1, storage2, storage3] });
+const structs: Array<StorageStruct> = [];
```
ryyakobe: I think we can initialize this array in-play, instead of using `push`:
```suggestion
const structs: Array<StorageStruct> = [
  new StorageStruct(componentTier, 'StorageStruct1', {
    integStackTag,
    provideDocdbEfs: false,
    useMongoDB: false,
  }),
  new StorageStruct(componentTier, 'StorageStruct2', {
    integStackTag,
    provideDocdbEfs: true,
    useMongoDB: false,
  }),
  new StorageStruct(componentTier, 'StorageStruct3', {
    integStackTag,
    provideDocdbEfs: false,
    useMongoDB: true,
  }),
];
```

integ/components/deadline/deadline_01_repository/bin/deadline_01_repository.ts
```diff
@@ -16,20 +16,25 @@ const env = {
...
+structs.push(
+  new StorageStruct(componentTier, 'StorageStruct1', {
+    integStackTag,
+    provideDocdbEfs: false,
+    useMongoDB: false,
+  }),
+  new StorageStruct(componentTier, 'StorageStruct2', {
+    integStackTag,
+    provideDocdbEfs: true,
+    useMongoDB: false,
```
ryyakobe: It seems weird that we are not testing all possible combinations (missing `provideDocdbEfs: true` and `useMongoDB: true`).

If I understand correctly, we can only have one of them used. So maybe instead of using 2 separate variables, we could usean `enum`. So in `storage-struct.ts` we could have:

```
export enum DatabaseType {
  DocDB = 1,
  MongoBD = 2
}

export interface StorageStructProps {
  readonly integStackTag: string;
  readonly databaseType?: DatabaseType;
}
...
if (props.databaseType == DatabaseType.DocDB) {
...
else if (props.databaseType == DatabaseType.MongoBD) {
```
And here, we will use that `enum`:
```
import { StorageStruct, DatabaseType } from '../../../../lib/storage-struct';
...
  new StorageStruct(componentTier, 'StorageStruct1', {
    integStackTag,
    databaseType: DatabaseType.MongoBD,
  }),
```

I think it will make our intention to use only one database more clear.

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  }
+
+  /**
+   * Mounts the Deadline repository's file system to the bastion and outputs the name of its log group
+   *
+   * @param testSuiteId Test case to configure the repository for
+   * @param repo Repository object to connect to the test Bastion
+   */
+  private configureRepo(testSuiteId:string, repo:Repository) {
```
ryyakobe, line 75: Nit: In constructor you use spaces between a variable and a type  `scope: Construct, id: string, props: RepositoryTestingTierProps)`
but here you don't: `testSuiteId:string, repo:Repository`.

I've noticed it in other files as well, that you don't use spaces for function parameters. So not sure if that's intentional.

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
```diff
@@ -0,0 +1,115 @@
...
+    const logGroup = repo.node.findChild('RepositoryLogGroup') as ILogGroup;
```
ryyakobe: Is it guaranteed that `repo` will have a child `RepositoryLogGroup`?
In other places I've noticed that we often use `tryFindChild` and the check the result:
```
const logGroup = repo.node.tryFindChild(RepositoryLogGroup) as ILogGroup);
if (logGroup)
```

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
```diff
@@ -0,0 +1,115 @@
...
+      value: logGroupName,
+    });
+  }
+
+  /**
+   * Installs MongoDB on the test instance
+   */
+  protected installMongodb() {
+    const userAcceptsSSPL = process.env.USER_ACCEPTS_SSPL_FOR_RFDK_TESTS;
+    if (userAcceptsSSPL){
```
ryyakobe: Nit: usually you have space before `{`

integ/components/deadline/deadline_01_repository/test/deadline_01_repository.test.ts
```diff
@@ -12,15 +12,15 @@ import awaitSsmCommand from '../../common/functions/awaitSsmCommand';
 const testingStackName = 'RFDKInteg-DL-TestingTier' + process.env.INTEG_STACK_TAG?.toString();
 const deadlineVersion = process.env.DEADLINE_VERSION?.toString();
 
-jest.setTimeout(10000);
+jest.setTimeout(60000);
```
ryyakobe: You made this comment in the PR itself:
> I did note that at least the WorkerFleet test can run up against jest's timeout limit and be in danger of failing. Since the tests make so many calls and stuff I decided to take care of it by adding one more update that ups the timeout limit on all of the jest tests to 60 seconds.

Maybe we could even add it as a comment here, so we don't remove/decrease this timeout in the future.

integ/components/deadline/deadline_02_renderQueue/bin/deadline_02_renderQueue.ts
```diff
@@ -27,17 +27,20 @@ const storage = new StorageStruct(componentTier, 'StorageStruct', {
...
-  repository: storage.repo,
-  protocol: 'http',
-});
-//Create test struct for Render Queue in https mode
-const render2 = new RenderStruct(componentTier, 'RenderStructRQ2', {
-  integStackTag,
-  repository: storage.repo,
-  protocol: 'https',
-});
+const structs: Array<RenderStruct> = [];
```
ryyakobe: Again, I think we can just initialize without `push`.

ryyakobe (commented):
Looks really great! Just a few small questions/suggestions:

integ/components/deadline/deadline_01_repository/test/deadline_01_repository.test.ts
aws-painec: That makes sense. The other option is to code the timeout into the `jest.config` file, then it wouldn't have to be set on a per-test basis but could still be overridden if necessary. I think I'll just do that.

integ/components/deadline/deadline_01_repository/bin/deadline_01_repository.ts
aws-painec: Agreed. Change made.

integ/components/deadline/deadline_01_repository/bin/deadline_01_repository.ts
aws-painec: Agreed, changed.

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
aws-painec: Fixed in all the cases I could find. If I missed any I can fix them in an upcoming commit that will revise the commenting in a bunch of these files.

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
aws-painec: Fixed.

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
aws-painec: It should definitely always exist but for better resilience/error-handling I've fixed this to use the convention you outlined. There are two cases in `storage-struct` where `findChild` is used similarly that I've left in place for now; they check for nodes that are guaranteed to be on the repository, and in fact they are nodes which should probably be directly exposed as properties, so there may be another issue to file there.

integ/components/deadline/deadline_02_renderQueue/bin/deadline_02_renderQueue.ts
aws-painec: Fixed.

integ/components/deadline/deadline_01_repository/test/deadline_01_repository.test.ts
ryyakobe: I can't find the place where we are setting this timeout now. Am I missing something?

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
ryyakobe: If it's guaranteed to always exist, then I'm fine with the original implementation.
And if we use `tryFindChild`, we should probably add some error reporting, otherwise we now silently skip an issue (although it's unlikely to happen):
```
if (!logGroup) {
   throw new Error(...);
}
```

aws-painec: By my read of the code it definitely should always be present, but there's a small chance that the name might not match up if something weird is happening. I've added the error handling so it will hopefully be obvious if something happens to break the function.

integ/components/deadline/deadline_01_repository/test/deadline_01_repository.test.ts
aws-painec: Accidentally didn't include it in the commit, which I will now correct. It's set in `jest.config.js`.

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
jusiskin: If we want to match the style of the rest of the codebase ([one example](https://github.com/aws/aws-rfdk/blob/cb68992de1c72f7997de6ff81f1b0c09e88dacaf/packages/aws-rfdk/lib/deadline/lib/render-queue.ts#L234)), we'd go with a keeping a space between the `:` and the type, so:

```ts
private configureRepo(testSuiteId: string, repo: Repository) {
```

in this case. It looks like you've pushed a change to remove the space everywhere 🤦 . If you do want to do this, we don't need to run another round of end-to-end testing. It would be sufficient to make sure the code builds properly. If you want to defer this clean-up to get this merged, I'm also okay with that. What do you think?

**UPDATE:** As a follow-up, we should probably have our linting enforce the code style so it is caught before code review.

jusiskin (approved):

ryyakobe (approved):
Great work! Thanks, Claire!

integ/components/deadline/deadline_01_repository/lib/repository-testing-tier.ts
jusiskin: Added #132 to track this

PR #139, opened 2020-09-29
feat: Update stage-deadline script with new version index 

kozlove-aws:
## Problem:

We need to update our stage-deadline script  to use index file with installers URLs, so customers can specify a version, such as 10.1.9 or 10.1.9.2, or latest and not have to enter the URL's themselves, or have the script attempt to form them on its own. The updated script should access the index, parse the JSON, and use that to determine where to download the customer's requested version.

## Solution

1) Was created lambda script that consume index file and return URLs for required product and platform.
2) Was modified stage-script to use lambda for getting URLs to Deadline client and Docker recipe.

## Testing

Were created unit tests for lambda script that covered 100% of functionality.
Was built project and tested that stage-script correctly download deadline client and Docker recipe in next cases:
- latest versions without any parameters.
- correct version when version is specified 
- latest possible version when specified partial version (like '10.1' or '10.1.8')

----


packages/aws-rfdk/bin/stage-deadline.ts
```diff
@@ -56,88 +57,135 @@ while (n < args.length) {
...
 }
 
-const deadlineInstallerURL = new url.URL(deadlineInstallerURI);
-const dockerRecipeURL = new url.URL(dockerRecipesURI);
 
-if (deadlineInstallerURL.protocol !== 's3:') {
-  console.error(`ERROR: Invalid URI protocol ("${deadlineInstallerURL.protocol}") for --deadlineInstallerURI. Only "s3:" URIs are supported.`);
-  usage(1);
+// populate docker recipe URI
+if (dockerRecipesURI === '') {
```
grbartel, line 97: This entire block is a copy & paste of the above (including variable names) with the exception of the product.
Can you convert this into a method.

packages/aws-rfdk/bin/stage-deadline.ts
```diff
@@ -350,3 +411,17 @@ function getAndExtractArchive(props: GetExtractArchiveProps) {
...
+    const version = Version.parse(versionString);
+    if(version.isLessThan(Version.MINIMUM_SUPPORTED_DEADLINE_VERSION)) {
+      console.error(`ERROR: Unsupported Deadline Version ${version.toString()}. Minimum supported version is ${Version.MINIMUM_SUPPORTED_DEADLINE_VERSION} \n`);
+      usage(1);
+    }
+  } catch(e) {
+    console.error(`ERROR: ${(e as Error).message} \n`);
+    usage(1);
+  }
+}
```
grbartel: Missing a new line at the end of this file.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -0,0 +1,357 @@
+/**
```
grbartel, line 1: This entire file has to do directly with Deadline so why is it in the Core library of the rfdk instead of the Deadline section.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -0,0 +1,357 @@
...
+  /* istanbul ignore next */ // @ts-ignore
+  public async doDelete(physicalId: string, resourceProperties: IVersionProviderProperties): Promise<void> {
+    // Nothing to do -- we don't modify anything.
+    return;
+  }
+
+  private implementsIVersionProviderProperties(value: any): boolean {
+    if (!value || typeof(value) !== 'object') { return false; }
+
+    // TODO: Ignore case
```
grbartel: Can you either implement these TODO's or are you going to create issues to fix them.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    return true;
+  }
+
+  /* istanbul ignore next */ // @ts-ignore
+  private async downloadInstallerIndex() {
+    const productionInfoURL = 'https://downloads.thinkboxsoftware.com/version_info.json';
+
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const parsedUrl = require('url').parse(productionInfoURL);
```
grbartel: do we have to use require imports here? or can we just do a normal import at the top of the file.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    const options = {
+      host: parsedUrl.hostname,
+      path: parsedUrl.path,
+    };
+
+    return new Promise((resolve, reject) => {
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      require('https').get(options, (res: IncomingMessage) => {
```
grbartel: Same question for imports.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -0,0 +1,357 @@
...
+          }
+        });
+      }).on('error', (err: Error) => {
+        reject(err);
+      });
+    });
+  }
+
+  /**
+   * This method reads index file.
```
grbartel: This docstring is not very useful since all it says is the name of the function.

What do we do with the read file return it ? throw it out? what do we return it as? a string? json blob?

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -0,0 +1,357 @@
...
+    }
+    const data = fs.readFileSync(this.indexFilePath, 'utf8');
+
+    // convert the response to a json object and return.
+    const json = JSON.parse(data);
+    return json;
+  }
+
+  private parseVersionString(versionString: string): RegExpExecArray | null {
+    const VALID_VERSION_REGEX = /^(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?$/;
```
grbartel: Since this REGEX is constant could we make this a class level constant.

Also could we use /^(0|[1-9]\d*)(?:\.(0|[1-9]\d*)){0,3}$/ instead. the {0,3} means 0-3 copies of the preceeding so we can use this instead of repeating the same section 3 times with the ? quantifier.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -0,0 +1,357 @@
...
+}
+
+/**
+ * The handler used to provide the installer links for the requested version
+ */
+/* istanbul ignore next */
+export async function handler(event: CfnRequestEvent, context: LambdaContext): Promise<string> {
+  const versionProvider = new VersionProvider();
+  return await versionProvider.handler(event, context);
+}
```
grbartel: missing a new line at the end of the file.

grbartel (changes requested):

packages/aws-rfdk/bin/stage-deadline.ts
```diff
@@ -56,88 +57,135 @@ while (n < args.length) {
...
+  console.error('The target directory is not empty.');
+  process.exit(1);
+}
 
-  // populate installer URI
-  if (deadlineInstallerURI && deadlineInstallerURI.length > 0) {
-    console.info('INFO: Since deadline release version is provided, "deadlineInstallerURI" will be ignored.');
-  }
-  deadlineInstallerURI = `s3://thinkbox-installers/Deadline/${deadlineReleaseVersion}/Linux/DeadlineClient-${deadlineReleaseVersion}-linux-x64-installer.run`;
+const handler = new VersionProvider();
```
horsmand, line 67: It's really weird for us to stand up a `SimpleCustomResource` here in this script, outside of a CDK construct. Could we refactor the code so that the methods like `getUrisForPlatform` and `getLatestVersion` are outside of the `VersionProvider` class? That way you can use them here without standing up the full `VersionProvider` and having to call its `doCreate` method.

I think it should be able to be accomplished by moving the methods from `VersionProvider` in `handler.ts` into another file inside the `version-provider` directory and then move their tests to a separate file as well. Then in this script you'd do something similar to what `doCreate()` is doing.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
kozlove-aws: Your expression work great for validation but it return only first and last part as result of `exec`

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
horsmand: This is because it requires classes in our Lambda library. I've created an issue to move the Lambdas out of `core` and will follow up on it after this PR is complete.
https://github.com/aws/aws-rfdk/issues/146

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  expect(result).not.toBeNull();
+
+  if (result === null) { return; }
+  expect(result[0]).toEqual('10.1.9.2');
+  expect(result[1]).toEqual('10');
+  expect(result[2]).toEqual('1');
+  expect(result[3]).toEqual('9');
+  expect(result[4]).toEqual('2');
+});
```
horsmand: This should enumerate the different types of valid and invalid strings to parse. For example:
10.1.9.2
10.1.9
10.1
10

10.9.1.2.1
10.
10.-1

Maybe just leave this test as-is (it has expectations for the full string and each individual part) and then have a second similar one that uses `test.each()` with the other valid values and just the `not.toBeNull()` and `result[0]).toEqual()` expectations. The invalid values could be a separate `test.each()` type of test.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+test('latest version', () => {
+  const result = versionProvider['getLatestVersion'](Platform.linux, productSection);
+
+  expect(result).toEqual('10.1.9.2');
+});
```
horsmand: Should test for all Platforms and Products.

packages/aws-rfdk/bin/stage-deadline.ts
```diff
@@ -221,8 +269,7 @@ Usage: stage-deadline [--output <output_dir>] [--verbose]
 
 Arguments:
     <deadline_release_version>
-        Specifies the official release of Deadline that should be staged. This must be of the form a.b.c.d.
-        Both '-d' or '-c' arguments will be ignored if provided with this value.
+        Specifies the official release of Deadline that should be staged. This must be of the form a.b.c.d, a.b.c, a.b or a.
```
horsmand: What do you think about putting quotes around the forms like "a.b.c.d"? This avoids the confusion of just the major version being `"a."` or `"a".`

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  expect(versionProvider.implementsIVersionProviderProperties({
+    product: Product.deadline,
+    versionString: 'version',
+  })).toEqual(false);
+
+  expect(versionProvider['implementsIVersionProviderProperties']({
+    product: Product.deadline,
+    platform: 'test',
+  })).toEqual(false);
+});
```
horsmand: These should be separate tests in a `describe()` block rather than expectations in a single test.

I've made the suggestion to split a few tests up into individual tests below as well. One reason I think this is important is because a test will stop running once an expectation fails. That means if all these expectations were to fail, you'd see the first one, fix it, re-run the build and tests, see the next one, fix it, run the build+test, and so on. It's a much more time consuming process than seeing all the failures and fixing them at once.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  expect(installerVersion?.PatchVersion).toEqual('2');
+});
+
+test('product is not in file', async () => {
+  await expect(versionProvider.getVersionUris({
+    product: Product.deadlineDocker,
+  })).rejects.toThrowError(/Information about product DeadlineDocker can't be found/);
+});
+
+test('get deadline version for all platforms', async () => {
```
horsmand, line 116: This looks like it's missing expectations for `Platform.mac`

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    versionString: 'version',
+  })).toEqual(false);
+
+  expect(versionProvider['implementsIVersionProviderProperties']({
+    product: Product.deadline,
+    platform: 'test',
+  })).toEqual(false);
+});
+
+test('invalide file path', () => {
```
horsmand: invalide -> invalid

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  expect(() => (new VersionProvider())['readInstallersIndex']()).toThrowError(/File path should be defined./);
+  expect(() => (new VersionProvider('test.txt'))['readInstallersIndex']()).toThrowError(/File test.txt was not found/);
```
horsmand: These should also be in their own tests rather than in a single test.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,187 @@
...
+});
+
+test('get latest version', () => {
+  expect(() => versionProvider['getLatestVersion']('linux',{})).toThrowError(/Information about latest version can not be found/);
+  expect(() => versionProvider['getLatestVersion']('linux',{latest: {}})).toThrowError(/Information about latest version for platform linux can not be found/);
```
horsmand: Split into 2 tests.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,187 @@
...
+});
+
+test('get requested Uri version', () => {
```
horsmand: I would change this to use `test.each()` with the version string and expected output as inputs.

horsmand (changes requested):

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    let uriIndex: IDeadlineInstallerUris | IDeadlineDockerRecipeUris | undefined;
+    if ((platform in versionMap)) {
+      if (product == Product.deadline) {
+        uriIndex = {
+          bundle: versionMap[platform].bundle,
+          clientInstaller: versionMap[platform].clientInstaller,
+          repositoryInstaller: versionMap[platform].repositoryInstaller,
+          certificateInstaller: versionMap[platform].certificateInstaller,
```
horsmand: I don't think this will work for the Mac section, it does not have repository or client installers, only the bundle. Our interface does not have repository and client as optional. We should write a unit test that covers this case, confirm it fails, and then update the interface to use optionals here.

packages/aws-rfdk/bin/index-test.json
```diff
@@ -0,0 +1,49 @@
...
+              "1": {
+                "8": {
+                  "5": {
+                      "linux": {
+                          "bundle": "s3://thinkbox-installers/Deadline/10.1.8.5/Linux/Deadline-10.1.8.5-linux-installers.tar",
+                          "clientInstaller": "s3://thinkbox-installers/Deadline/10.1.8.5/Linux/DeadlineClient-10.1.8.5-linux-x64-installer.run",
+                          "repositoryInstaller": "s3://thinkbox-installers/Deadline/10.1.8.5/Linux/DeadlineRepository-10.1.8.5-linux-x64-installer.run"
+                      },
+                      "mac": {
+                          "installer": "s3://thinkbox-installers/Deadline/10.1.8.5/Mac/Deadline-10.1.8.5-osx-installers.dmg"
```
horsmand: This is incorrectly named, it will be "bundle" in the real index file.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
```diff
@@ -0,0 +1,323 @@
...
+}
+
+export interface IDeadlineInstallerUris extends IUris {
+  readonly bundle: string;
+
+  readonly clientInstaller: string;
+
+  readonly repositoryInstaller: string;
+
+  readonly certificateInstaller?: string;
```
horsmand: If we have to make the client and repository optional here to account for how our Mac installers work anyways, what do you think about getting rid of `IDeadlineDockerRecipeUris` and `IDeadlineInstallerUris` then moving these fields to `IUris`?

None of it is ideal, I still kind of think we should have a separate class for retrieving values from the DeadlineDocker section, but this would at least allow us to avoid having to make unsafe casts for the return types while leaving the logic in a single class.

packages/aws-rfdk/bin/stage-deadline.ts
kozlove-aws: Good point.

packages/aws-rfdk/bin/index-test.json
horsmand: Still needs to be addressed

horsmand:
Added the do not merge label because this is dependent on the version index file being made available on the downloads website first.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/test/version-provider.test.ts
```diff
@@ -0,0 +1,243 @@
...
+    10: {
+      1: {
+        9: {
+          2: {
+            linux: 's3://thinkbox-installers/DeadlineDocker/10.1.9.2/DeadlineDocker-10.1.9.2.tar.gz',
+          },
+        },
+      },
+    }}, Platform.windows, Product.deadlineDocker )).toEqual(undefined);
+});
```
horsmand: Missing new line

horsmand: Looks like this is still missing.

horsmand (dismissed):

horsmand:
The index file has been published, I've removed the do not merge label. Everything looks good, approved.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    return true;
+  }
+
+  /* istanbul ignore next */ // @ts-ignore
+  private async downloadInstallerIndex() {
+    const productionInfoURL = 'https://downloads.thinkboxsoftware.com/version_info.json';
+
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const parsedUrl = require('url').parse(productionInfoURL);
```
grbartel: Is there a reason we can't do this import at the top like normal?

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    const options = {
+      host: parsedUrl.hostname,
+      path: parsedUrl.path,
+    };
+
+    return new Promise((resolve, reject) => {
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      require('https').get(options, (res: IncomingMessage) => {
```
grbartel: Same question for imports

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
```diff
@@ -0,0 +1,318 @@
...
+  ): IVersionedUris | undefined {
+
+    let versionMap = indexedVersionInfo;
+    const versionArray: string[] = [];
+
+    // iterate through all 4 major, minor, release and patch numbers,
+    // and get the matching version from the indexed version map.
+    for (let versionIndex = 0; versionIndex < 4; versionIndex++) {
+      let version: string;
+      if (null === requestedVersion
```
grbartel: can this be changed to using optional chaining something along the lines of

requestedVersion?.[versionIndex + 1] == null

grbartel (changes requested):

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
```diff
@@ -0,0 +1,315 @@
...
+
+export class VersionProvider {
+  private readonly indexFilePath: string|undefined;
+  private readonly VALID_VERSION_REGEX = /^(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?$/;
+  constructor(indexFilePath?: string) {
+    this.indexFilePath = indexFilePath;
+  }
+
+  /**
+   * @inheritdoc
```
ddneilson: inheritdoc from what? This is a base class

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
```diff
@@ -0,0 +1,315 @@
...
+  /**
+   * The URLs to installers
+   */
+  readonly Uris: IUris;
+}
+
+export class VersionProvider {
+  private readonly indexFilePath: string|undefined;
+  private readonly VALID_VERSION_REGEX = /^(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?$/;
+  constructor(indexFilePath?: string) {
```
ddneilson, line 80: A docstring explaining the interface for this class would be lovely. It looks like there's some default behavior based on whether `indexFilePath` has been defined or not. Would be great to have that explained in a docstring.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public async getVersionUris(resourceProperties: IVersionProviderProperties): Promise<Map<Platform, IVersionedUris>> {
+    /* istanbul ignore next */
+    const indexJson = this.indexFilePath ? this.readInstallersIndex() : await this.downloadInstallerIndex();
+
+    const productSection = indexJson[resourceProperties.product];
```
ddneilson, line 93: Would be nice to see some defensive programming here. If the format of the index file changes (though unlikely) or the product code isn't in the json then this just blows up as it is.

ddneilson (changes requested):

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/version-provider.ts
kozlove-aws: I can't see any undefensive code here. I even have test with not existing product and it returns correct exception.

ddneilson (approved):

horsmand (approved):
Grant's comments are still unresolved, but I can see they've been addressed. Looks good to me!

grbartel (approved):

PR #155, opened 2020-10-06
feat(integ): add ability to use hook function before each component test

kozlove-aws:
## Problem
In the main integ-rfdk-e2e.sh we need to add the ability to run an extra function if it is designed.

## Solution
Added functionality to invoke script in case when environment variable `PRE_COMPONENT_HOOK` is defined and function with such name was exported.

## Testing
Script was tested locally with all cases like: correct function, undefined variable or function name is incorrect.

----

### DIFF
--- a/integ/scripts/bash/rfdk-integ-e2e.sh
+++ b/integ/scripts/bash/rfdk-integ-e2e.sh
@@ -73,6 +73,12 @@ export INFRASTRUCTURE_DEPLOY_FINISH_TIME=$SECONDS
 for COMPONENT in **/cdk.json; do
     COMPONENT_ROOT="$(dirname "$COMPONENT")"
     COMPONENT_NAME=$(basename "$COMPONENT_ROOT")
+    # Invoke hook function if it is exported and name is defined in PRE_COMPONENT_HOOK variable
+    if [ -n "$PRE_COMPONENT_HOOK" ]  && [ "$(type -t $PRE_COMPONENT_HOOK)" == "function" ]
+    then
+      $PRE_COMPONENT_HOOK $COMPONENT_NAME
+    fi
+
     # Use a pattern match to exclude the infrastructure app from the results
     export ${COMPONENT_NAME}_START_TIME=$SECONDS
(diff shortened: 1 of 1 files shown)


integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -71,6 +71,12 @@ export INFRASTRUCTURE_DEPLOY_FINISH_TIME=$SECONDS
 
 # Pull the top level directory for each cdk app in the components directory
 for COMPONENT in **/cdk.json; do
+    # Invoke hook function if it is exported and name is defined in HOOK_FUNCTION variable
+    if [ -n "$HOOK_FUNCTION" ]  && [ "$(type -t $HOOK_FUNCTION)" == "function" ]
```
jusiskin: How would you feel about using a more descriptive variable name? Say `PRE_COMPONENT_HOOK`?

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -71,6 +71,12 @@ export INFRASTRUCTURE_DEPLOY_FINISH_TIME=$SECONDS
...
+    then
+      $HOOK_FUNCTION
```
jusiskin: Is there any argument/parameters that might be useful to pass to the hook function? I'm trying to think in the general case if someone would need a hook function that is triggered before each component test, they'd likely be interested in knowing which component is being tested. Does it make sense to include the component name?

jusiskin (changes requested):
Looks pretty straight-forward - just a few minor suggestions.

jusiskin:
Can we scope the conventional commit title to `feat(integ):`? This is how we have scoped prior commits to the integration tests.

kozlove-aws:
> Can we scope the conventional commit title to `feat(integ):`? This is how we have scoped prior commits to the integration tests.

jusiskin (approved):

ddneilson (approved):

PR #160, opened 2020-10-08
feat(integ): use configurable version of Deadline for integration tests

kozlove-aws:
## Problem
The integration tests for RFDK need to be updated to always find the installers for the latest version of Deadline.

## Solution
After stage script received ability to take URIs for last versions by default we need just invoke it with specified output folder.

## Testing
Were run integration test (including automation) and validate that was used the latest version and latest AMIs.

----


integ/scripts/bash/stage-deadline.sh
```diff
@@ -8,7 +8,7 @@ set -euo pipefail
 # Check if Deadline assets are staged at configured path
 if [ ! $(ls "$DEADLINE_STAGING_PATH/manifest.json" 2> /dev/null) ]; then
     # Stage Deadline assets
-    npx ../packages/aws-rfdk/bin/stage-deadline --deadlineInstallerURI "s3://thinkbox-installers/Deadline/$DEADLINE_VERSION/Linux/DeadlineClient-$DEADLINE_VERSION-linux-x64-installer.run" --dockerRecipesURI "s3://thinkbox-installers/DeadlineDocker/$DEADLINE_VERSION/DeadlineDocker-$DEADLINE_VERSION.tar.gz" --output "$DEADLINE_STAGING_PATH"
+    npx ../packages/aws-rfdk/bin/stage-deadline --output "$DEADLINE_STAGING_PATH"
```
jusiskin: Are we sure we want to always test with the latest version of Deadline? What if we instead make the default behavior to use the latest version of Deadline, but still allow `DEADLINE_VERSION` to be specified? I know there is an existing way to do this with `DEADLINE_STAGING_PATH`, but this would remove that extra manual step.

Regardless of that, there are other references to the `${DEADLINE_VERSION}` variable that would need to be updated/removed.

*   In [`integ/scripts/bash/set-test-variables.sh`](https://github.com/aws/aws-rfdk/blob/58410d21cea1c2ba66a96ba5099ca50ed7da7fcc/integ/scripts/bash/set-test-variables.sh#L30-L33)
*   In `integ/test-config.sh`:
    *   [L10-12](https://github.com/aws/aws-rfdk/blob/0b42aa73418d468ad82b0bf519244589bc835293/integ/test-config.sh#L10-L12)
    *   [L16](https://github.com/aws/aws-rfdk/blob/0b42aa73418d468ad82b0bf519244589bc835293/integ/test-config.sh#L16)
*   In [`integ/scripts/bash/fetch-worker-amis.sh`](https://github.com/aws/aws-rfdk/blob/0b42aa73418d468ad82b0bf519244589bc835293/integ/scripts/bash/fetch-worker-amis.sh#L10)
    *   This one may be harder to remove.
*   In [`integ/README.md`](https://github.com/aws/aws-rfdk/blame/0b42aa73418d468ad82b0bf519244589bc835293/integ/README.md#L13)

jusiskin (changes requested):

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -40,7 +40,7 @@ export INTEG_TEMP_DIR="$INTEG_ROOT/.e2etemp"
 mkdir -p $INTEG_TEMP_DIR
 
 # Stage deadline from script
-$BASH_SCRIPTS/stage-deadline.sh
+$BASH_SCRIPTS/stage-deadline.sh $DEADLINE_VERSION
```
jusiskin: I believe you intended to make this change in `integ/scripts/bash/stage-deadline.sh` (line 11) instead, right? We're calling that script here, and it does not appear to accept any command-line arguments.

jusiskin (changes requested):

integ/scripts/bash/stage-deadline.sh
jusiskin: Looks like you've opted to keep the variable, so I'll resolve this discussion.

integ/scripts/bash/rfdk-integ-e2e.sh
kozlove-aws: Yes, you are right.

jusiskin: Looks good. resolving

jusiskin (changes requested):
Just recalled there was one outstanding issue on my first discussion. The `integ/scripts/bash/fetch-worker-amis.sh` file makes use of the `DEADLINE_VERSION` environment variable here:

https://github.com/aws/aws-rfdk/blob/3a94e588fd816e441bff3f62bd691c2d9435f6a8/integ/scripts/bash/fetch-worker-amis.sh#L10

When this is the empty string or an unresolved version string, the URL will not work. This script is called here:

https://github.com/aws/aws-rfdk/blob/3a94e588fd816e441bff3f62bd691c2d9435f6a8/integ/scripts/bash/rfdk-integ-e2e.sh#L42-L48

Notice how `stage-deadline` is called just above? We can use the `"version"` field in the `manifest.json` that is staged by the ingration tests to obtain the fully-resolved version number for obtaining the Worker AMIs. What do you think?

horsmand (changes requested):
I believe Josh's previous comment about `fetch-worker-amis.sh` still stands, and we also now have the script [run-e2e-automated.sh](https://github.com/aws/aws-rfdk/blob/mainline/integ/scripts/bash/run-e2e-automated.sh) that is used as an entry-point into the integration tests from CodeBuild, to run them in our pipeline. We'll need to change the way that script is setting `DEADLINE_VERSION` as well.

integ/scripts/bash/set-test-variables.sh
```diff
@@ -27,7 +27,7 @@ elif [ $(ls "$DEADLINE_STAGING_PATH/manifest.json" 2> /dev/null) ]; then
     export DEADLINE_VERSION=$(node -e $'const json = require(process.argv[1] + \'/manifest.json\'); console.log(json.version)' "$DEADLINE_STAGING_PATH")
 fi
 
-# Set Deadline version to default if not overridden
+# Set Deadline version to empty if not overridden
 if [ -z ${DEADLINE_VERSION+x} ]; then
-    export DEADLINE_VERSION="10.1.9.2"
+    export DEADLINE_VERSION=""
 fi
```
horsmand: Can we just remove this altogether? It seems like the same check we're doing in `rfdk-integ-e2e.sh`

integ/scripts/bash/run-e2e-automated.sh
```diff
@@ -32,7 +32,7 @@ refreshcreds() {
 
 # Basic integ test configuration
 export SKIP_TEST_CONFIG=true
-export DEADLINE_VERSION="10.1.11.5"
+export DEADLINE_VERSION
```
horsmand: I would suggest removing this line from here and modifying the comment at the top of the file to say that if the `DEADLINE_VERSION` environment variable isn't set before running this script, it will default to the latest version. Maybe add it to the list of expected env vars like:
```
# DEADLINE_VERSION - Optional, will default to the latest Deadline version if not set, otherwise the specified version will be used.
```

integ/scripts/bash/set-test-variables.sh
kozlove-aws: This code was moved to `stage-deadline.sh` - we still need to have defined variable to run stage script.

integ/scripts/bash/run-e2e-automated.sh
kozlove-aws: I do not like when such variable is hidden in comments.
What do you say if we left it like in `test-config.sh` with the same comment?

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -42,6 +42,9 @@ mkdir -p $INTEG_TEMP_DIR
 # Stage deadline from script
 $BASH_SCRIPTS/stage-deadline.sh
 
+# Extract the Deadline version to use for Deadline installations on the farm. This will override any other Deadline version provided.
+export DEADLINE_VERSION=$(node -e $'const json = require(process.argv[1] + \'/manifest.json\'); console.log(json.version)' "$DEADLINE_STAGING_PATH")
```
jusiskin: My understanding of the purpose of this is code is to fully-resolve the `DEADLINE_VERSION` after it has been staged. We need the fully-resolved version to determine the AWS Portal AMI IDs programatically. To do this, we are getting the resolved version from the stage manifest. So if someone running the tests set `DEADLINE_VERSION` initially to `10.1.11`, we can look at the manifest and get the full version (e.g. `10.1.11.5`) and use that to find the corresponding AMI IDs.

If this is correct, can we update the comment to explain this? Otherwise, it's not really clear why we are doing this.

jusiskin (changes requested):
Just needs a clearer code comment, but otherwise looks good

integ/scripts/bash/run-e2e-automated.sh
horsmand: I don't think we should leave it like `test-config.sh` because the two scripts serve different purposes.

`test-config.sh` is meant to be a config file that someone would manually edit before manually kicking off the integration tests.

`run-e2e-automated.sh` is meant to be the entry point for the tests when running them in an automated fashion. It gets executed from our pipeline, where you don't have the opportunity to manually edit this value before the tests get executed, but we do have the opportunity to configure environment variables to be set before the execution.

If we remove this line, we could set the `DEADLINE_VESION` environment variable before running the script, if we leave this line here, it will reset the env var set by the user, forcing the tests to always use the latest version.

integ/scripts/bash/rfdk-integ-e2e.sh
kozlove-aws: You are right. 
Also if we have not ran stage script because manifest file is already in stage folder we just take correct version from this file.

integ/scripts/bash/set-test-variables.sh
horsmand: Ah yes, you're right, I overlooked the fact that the script would fail if it wasn't being set to the empty string like this.

horsmand (approved):

integ/scripts/bash/rfdk-integ-e2e.sh
jusiskin: right. looks good now. resolved

jusiskin (approved):

PR #171, opened 2020-10-13
chore(examples): Update Python example deps

### LINKED ISSUE #172: Upgrade the "All in AWS" python example
## :question: General Issue

After bumping to `0.18.0` we noticed that our bump script doesn't bump the python example. The RFDK version used in `examples/deadline/All-In-AWS_Infrastructure-Basic/python` should be manually bumped to `0.18.0`. The CDK version should be bumped to `1.66.0` as well, to match the version used across the rest of the repository.

horsmand:
Fixes https://github.com/aws/aws-rfdk/issues/172

Upgrades the AWS CDK and RFDK versions used by the All in AWS infrastructure Python example app.

This was tested by deploying a render farm using this upgrade and Deadline 10.1.10.6.

----

### DIFF
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/python/.gitignore
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/python/.gitignore
@@ -8,4 +8,5 @@ __pycache__
 # CDK asset staging directory
 .cdk.staging
 cdk.out
+cdk.context.json
 stage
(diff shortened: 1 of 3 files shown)


horsmand:
I've added in one other small change to fix the default value for the `key_pair_name` as I found that using the string `''` caused failures when trying to deploy the worker fleet. It didn't seem like a big enough change to warrant its own bug fix line in our change log.

jusiskin (approved):

ddneilson (approved):

PR #175, opened 2020-10-14
feat(deadline): versionquery construct

### LINKED ISSUE #176: Create a Construct to supply Deadline installers
We would like a construct which determines the S3 URI's for the Deadline installers and Deadline Docker recipes. The version of Deadline for these can be either specified by the user or defaulted to the latest version.

For now, this will be integrated with the Repository construct to get the repository installer for it to use. In the future we can extend it out to be used by other constructs that require Deadline installers or the Docker recipes as well.

### Use Case

Currently to get the repository installer, we use hard-coded values that make assumptions about where the installer should be found in S3:
https://github.com/aws/aws-rfdk/blob/75d0f9fedce52de9dcfe73482965de923fa8941d/packages/aws-rfdk/lib/deadline/lib/version.ts#L293-L310

This would provide a more robust way of determining the installer locations

### Proposed Solution

A VersionQuery Construct that contains a CustomResource that downloads the [version index file](https://downloads.thinkboxsoftware.com/version_info.json) and parses it, then returns the required information about the installers for a specific Deadline version.

The Repository construct will then take the versioned installers supplied by the VersionQuery construct, and use the repo installer.

* [x] :warning: This feature might incur a breaking change

---

horsmand:
Fixes https://github.com/aws/aws-rfdk/issues/176

This was tested by successfully deploying the TypeScript and Python example apps, and running the integration tests.

----


examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
```diff
@@ -117,11 +118,13 @@ def __init__(self, scope: Construct, stack_id: str, *, props: ServiceTierProps,
             stage=Stage.from_directory(props.docker_recipes_stage_path)
         )
 
+        version_query = VersionQuery(self, 'VersionQuery')
```
horsmand: I'm not sure how to handle this. If you pip install from PyPI (which is how the example is set up to work), this won't work because VersionQuery hasn't been released yet. If you package locally and pip install the result of that, it will work.

If we remove this now to keep the example working, it will eventually break when we do our next release, and we'll need to remember to come back and update it then.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -7,52 +7,253 @@
...
+      product: Product.deadlineDocker,
+      platform: Platform.linux,
+    });
+
+    const deadlineDockerLinux = deadlineDockerPlatFormVersionedUris.get(Platform.linux)!;
+    const deadlineDockerLinuxUris = deadlineDockerLinux.Uris;
+
+    const s3Bucket = this.parseS3BucketName(deadlineLinuxUris.bundle);
+
+    return {
```
ddneilson, line 105: This doesn't need to return *everything* that is in the metadata file -- it only needs to return the values that would/could be used in an RFDK application. More importantly, we don't want to be breaking the `IVersion` interface -- it's a breaking change, and that would require a major version bump of the RFDK. 

There is a way to do this that won't entail a breaking change.
1. Change the return value of this function to match a flattened `IVersion`. ex:
```
return {
  MajorVersion: ... , // -> IVersion.majorVersion
  MinorVersion: ... , // -> IVersion.minorVersion
  ReleaseVersion: ... , // -> IVersion.releaseVersion
  LinuxPatchVersion: ... , // -> IVersion.linuxInstallers.patchVersion
  LinuxRepositoryS3Bucket: ..., // -> IVersion.linuxInstallers.repository.s3Bucket
  LinuxRepositoryObjectKey: ..., // -> IVersion.linuxInstallers.repository.objectKey
```

We don't currently need the docker bundle information, or the client installer information, or the mac information, or the Windows information. We can always add those later if/when the need arises and we extend the `IVersion` interface to contain that information.

2. Undo all of the breaking changes to `IVersion` and `VersionQuery`.

3. In the implementation of the `VersionQuery` constructor, just invoke this Custom Resource and put the returned values directly into the appropriate spots in the `IVersion`.

ddneilson (changes requested):
In hindsight, we should have coordinated how this was being implemented better before it got this far. A breaking change is not required to get this to work.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
ddneilson: You should be able to `pip install` the self-built Python package that contains your change into the example.

horsmand: Yes there's no problem doing that, I mentioned it works in my original comment. The issue is what should our expectations be on the customer for them to get the example to work. I guess recommending they only work off the release branch should be fine. I'll update the readme to state that and also add step-by-step instructions on how to get it to work with the latest mainline version by running `pack.sh` and then `pip install` for the packaged python version.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
horsmand: This is fine, but if we ever need to add the Mac, Windows, or Deadline Docker recipes in the future, we'd likely need to make a breaking change since the patch versions aren't guaranteed to be the same. I think it makes more sense to have a breaking change now that allows this to be extensible to support those in the future, rather than breaking at a later time. 

Maybe most of the other Platform installers and the recipes are unlikely to be needed, so it's fine not to worry about supporting them, but do we see this construct also holding the Docker containers when we begin publishing those? We may want to consider that use case now.

horsmand: Actually I see how my last comment isn't correct. I was thinking `IVersion` extends `IPatchVersion`, but it extends `IReleaseVersion`. We could add additional fields to the `IVersion` interface to support the other use cases without breaking backwards compatibility.

It doesn't make sense for `linuxFullVersionString()` to be a method though, since the full value needs to be returned from the CustomResource rather than combining the version parts into a string. The version parts will be Tokens, so trying to detokenize them in `linuxFullVersionString()` wouldn't work. I believe I can make this change without breaking backwards compatibility.

ddneilson: > we'd likely need to make a breaking change since the patch versions aren't guaranteed to be the same.

Patch versions are allowed to vary, but you caught on to that with the structuring of `IReleaseVersion` and `IPatchVersion` -- a difference in major/minor/release versions between Windows, Linux, and macOS should be treated as an error. Realistically, it'll never happen unless Deadline discontinues support for an operating system.

> It doesn't make sense for linuxFullVersionString() to be a method though, since the full value needs to be returned from the CustomResource rather than combining the version parts into a string. 

AFAIK, this isn't true. CDK lets you combine tokens in a string. We do this with our UserData strings when embedding Secret ARNs for our TLS certificate data. For example, this should work:
```ts
const major = Token.asString(deadlineResource.getAtt('MajorVersion'));
const minor = Token.asString(deadlineResource.getAtt('MinorVersion'));
const release = Token.asString(deadlineResource.getAtt('ReleaseVersion'));
const patch = Token.asString(deadlineResource.getAtt('LinuxPatchVersion'));

const linuxVersionString = `${major}.${minor}.${release}.${patch}`;
```

packages/aws-rfdk/lib/deadline/lib/version-query.ts
```diff
@@ -0,0 +1,156 @@
...
+      serviceToken: lambdaFunc.functionArn,
+      properties: deadlineProperties,
+      resourceType: 'Custom::RFDK_DEADLINE_INSTALLERS',
+    });
+
+    this.majorVersion = Token.asNumber(deadlineResource.getAtt('MajorVersion'));
+    this.minorVersion = Token.asNumber(deadlineResource.getAtt('MinorVersion'));
+    this.releaseVersion = Token.asNumber(deadlineResource.getAtt('ReleaseVersion'));
+
+    this.linuxFullVersionString = Token.asString(deadlineResource.getAtt('LinuxFullVersionString'));
```
ddneilson: Do we actually need to get the `linuxFullVersionString` as an output of the Custom Resource?

AFAIK, we should be able to just get major, minor, release, and patch version values separately and then keep `linuxFullVersionString()` as a function defined as:
```ts
public linuxFullVersionString(): string {
  return `${this.majorVersion}.${this.minorVersion}.${this.releaseVersion}.${this.linuxInstallers.patchVersion}`;
}
```

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
```diff
@@ -8,51 +8,175 @@
...
+  public async doDelete(physicalId: string, resourceProperties: IVersionProviderResourceProperties): Promise<void> {
     // Nothing to do -- we don't modify anything.
     return;
   }
+
+  private implementsIVersionProviderResourceProperties(value: any): boolean {
+    if (!value || typeof(value) !== 'object') { return false; }
+
+    if (value.versionString) {
+      if (!Version.validateVersionString(value.versionString))  { return false; }
```
ddneilson: It's probably worth adding a `console.log()` here if the validation of the version string fails just to provide some breadcrumbs in the lambda log.

If validation fails, then the customer will see "Input did not pass validation check" ( https://github.com/aws/aws-rfdk/blob/79e0d1f86389b289596974b15bcf49c735d72e56/packages/aws-rfdk/lib/core/lambdas/nodejs/lib/custom-resource/simple-resource.ts#L88 ) as the failure reason on the Custom Resource, but understanding why it failed the validation won't be obvious. Might actually be worth extending the error message in the base handler to point the customer to the Lambda log as well (the LambdaContext has the log group & stream name in it -- https://github.com/aws/aws-rfdk/blob/79e0d1f86389b289596974b15bcf49c735d72e56/packages/aws-rfdk/lib/core/lambdas/nodejs/lib/aws-lambda/types.ts#L13 ).

packages/aws-rfdk/lib/core/lambdas/nodejs/lib/version-provider/index.ts
```diff
@@ -3,4 +3,5 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export * from './version-provider';
+export * from './version';
+export * from './version-provider';
```
ddneilson: nit: Missing newline at end of file.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
```diff
@@ -44,6 +44,7 @@
     ThinkboxDockerRecipes,
     UsageBasedLicense,
     UsageBasedLicensing,
+    VersionQuery,
```
ddneilson: Looks like this isn't used in the file.

integ/components/deadline/deadline_01_repository/bin/deadline_01_repository.ts
```diff
@@ -17,17 +19,23 @@ const integStackTag = process.env.INTEG_STACK_TAG!.toString();
 
 const componentTier = new Stack(app, 'RFDKInteg-DL-ComponentTier' + integStackTag, {env});
 
+// This will get the installers for the latest version of Deadline
+const version = new VersionQuery(componentTier, 'VersionQuery');
```
ddneilson: The integ tests currently allow for testing arbitrary versions of Deadline (not just latest) through definition of the `DEADLINE_VERSION` environment variable. I'm thinking that we should probably use/reference that in these additions as well. Reason being tha it's possible that the container images will be built from a different version of Deadline than latest (they're based on the staged version), and we'd want the version that we're passing to the constructs to match that version.

I do like the idea of getting the Repository's version from the VersionQuery in our tests -- makes sure that it works. 

What do you think?

integ/lib/render-struct.ts
```diff
@@ -79,11 +86,11 @@ export class RenderStruct extends Construct {
       vpc,
       repository: props.repository,
       images: recipes.renderQueueImages,
-      version: recipes.version,
       logGroupProps: {
         logGroupPrefix: Stack.of(this).stackName + '-' + id,
       },
       hostname,
+      version: props.version,
```
ddneilson: I'm thinking that, for now, we should keep the RenderQueue's version as the version from the recipe for now. Reason being that we want the RenderQueue's version to match the staged version.

What do you think?

packages/aws-rfdk/lib/deadline/lib/thinkbox-docker-recipes.ts
```diff
@@ -106,6 +106,8 @@ export class ThinkboxDockerRecipes extends Construct {
 
     this.version  = props.stage.getVersion(this, 'Version');
 
+    // this.versionQuery  = props.stage.getVersion(this, 'Version');
```
ddneilson: Looks like this was here for testing, and can be removed.

packages/aws-rfdk/lib/deadline/test/thinkbox-docker-recipes.test.ts
```diff
@@ -129,40 +132,13 @@ describe('ThinkboxDockerRecipes', () => {
...
 
-  describe('provides the Deadline version with', () => {
-    let version: IVersion;
-    beforeEach(() => {
-      // WHEN
-      const recipes = new ThinkboxDockerRecipes(stack, 'Recipes', {
-        stage,
-      });
-      version = recipes.version;
+  test('provides the Deadline version', () => {
```
ddneilson, line 137: These changes don't look right. There weren't, and shouldn't be, any changes to the `ThinkboxDockerRecipes` construct, so the test shouldn't change.

ddneilson (changes requested):
Thanks for the changes, David. This is looking much better.

packages/aws-rfdk/lib/deadline/lib/version-query.ts
horsmand: I tried this and got output like `-1.888154589708867e+289.-1.8881545897088673e+289.-1.8881545897088675e+289.-1.8881545897088678e+289`

integ/components/deadline/deadline_01_repository/bin/deadline_01_repository.ts
horsmand: I was looking at my changes to the integ tests and I was thinking that rather than doing the VersionQuery at this point and injecting it into the other structs, it would make more sense to pull the ThinkboxDockerRecipes up to this level and inject that instead. That way the version used in the VersionQuery would match the version used by the recipes, which is defined with `DEADLINE_VERSION`. This is how the example apps were changed to work.

integ/lib/render-struct.ts
horsmand: Ah yes, I just made a similar reply to your previous comment

packages/aws-rfdk/lib/deadline/test/thinkbox-docker-recipes.test.ts
horsmand: The version belonging to `ThinkboxDockerRecipes` comes from `Stage` and is an instance of a `VersionQuery` now. Since these values are outputs of the CustomResource in `VersionQuery`, we can't validate them in a unit test. I think I can improve the `expectCDK` statement to validate that the `Custom::RFDK_DEADLINE_INSTALLERS` has `FULL_VERSION_STRING` in its input.

packages/aws-rfdk/lib/deadline/lib/version-query.ts
jusiskin: I recall having this issue embedding [number-encoded CDK tokens](https://docs.aws.amazon.com/cdk/latest/guide/tokens.html#tokens_number) in a string and this won't work naively. The CDK doesn't replace numeric tokens within a string. I suppose this may be because there is loss of precision in the round-trip encoding and decoding from floating point numbers to and from strings. It also would be hard to differentiate when such a number is used adjacent to a numeric string character that is not part of the token.

There is a work-around that involves first checking the number to see if it is unresolved (in this case we know it is an unresolved token since it comes from a resource output attribute), and if so, wrap that numeric token with a string token. You can see an example of this in our [`Endpoint.portAsString()`](https://github.com/aws/aws-rfdk/blob/79e0d1f86389b289596974b15bcf49c735d72e56/packages/aws-rfdk/lib/core/lib/endpoint.ts#L118-L133) implementation.

horsmand: Updated to use Josh's suggestion.

packages/aws-rfdk/lib/core/lambdas/nodejs/version-provider/handler.ts
horsmand: Updated

packages/aws-rfdk/lib/core/lambdas/nodejs/lib/version-provider/index.ts
horsmand: Fixed

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
horsmand: Removed

packages/aws-rfdk/lib/deadline/lib/thinkbox-docker-recipes.ts
horsmand: Removed

examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
```diff
@@ -12,30 +12,43 @@ These instructions assume that your working directory is `examples/deadline/All-
...
+1.  This sample app on the `mainline` branch may contain features that have not yet been officially released, and may not be available in the `aws-rfdk` package installed through pip from PyPI. To work from an example of the latest release, please switch to the `release` branch. If you would like to try out unreleased features, you can stay on `mainline` and follow the instructions for building, packing, and installing the `aws-rfdk` from your local repository.
+2.  Install the dependencies of the sample app:
 
     ```bash
     pip install -r requirements.txt
     ```
-2.  Change the value in the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker.
+3.  If working on the `release` branch, this step can be skipped. If working on `mainline`, navigate to the base directory where the build and packaging scripts are, then run them and install the result over top of the `aws-rfdk` version that was installed in the previous step:
+    ```bash
+    cd ../../../..
```
yashda: We should either explicitly mention the current directory to which this command is relative to, or avoid the relative path completely.

How about doing this:
```
cd <aws-rfdk root directory>
...
cd <aws-rfdk root directory>/examples/deadline/All-In-AWS-Infrastructure-Basic-Tiered/python
```

examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
```diff
@@ -12,30 +12,43 @@ These instructions assume that your working directory is `examples/deadline/All-
...
+    # Enter the Docker container to run the build and pack scripts
+    ./scripts/rfdk_build_environment.sh
+    ./build.sh
+    ./pack.sh
+    # Exit the Docker container
+    exit
+    cd ./examples/deadline/All-In-AWS-Infrastructure-Basic-Tiered/python
```
yashda: How about using `pushd` & `popd` for jumping back to this directory?

integ/components/deadline/deadline_01_repository/bin/deadline_01_repository.ts
yashda: I think using the staged version here by using ThinkboxDockerRecipes here makes more sense. This way there is a consistency between repository and RCS versions. Also, we can run integration tests by specifying a single version.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
horsmand: We already have this note in the readme (and a similar in the TS readme) above the step-by-step instructions, so I was just working with that:
```
**NOTE**

These instructions assume that your working directory is `examples/deadline/All-In-AWS-Infrastructure-Basic-Tiered/python/` relative to the root of the AWS-RFDK package.
```

examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
horsmand: Updated

examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
yashda: I missed that. Thanks for pointing that out. Resolving.

packages/aws-rfdk/lib/deadline/lib/version-query.ts
```diff
@@ -0,0 +1,186 @@
...
+      handler: 'version-provider.handler',
+      timeout: Duration.seconds(30),
+      logRetention: RetentionDays.ONE_WEEK,
+    });
+
+    const deadlineProperties: IVersionProviderResourceProperties = {
+      versionString: props?.version,
+      // If we don't have a full static version string, create a random string that will force the Lambda to always
+      // run on redeploys, effectively checking for version updates.
+      forceRun: this.forceRun(props?.version),
```
ddneilson, line 133: I like the way you did this!

ddneilson (dismissed):

yashda (dismissed):
Looks good, approving! Thanks David!

yashda (dismissed):

ddneilson (approved):

yashda (approved):

PR #180, opened 2020-10-16
feat(examples):  Added examples for Spot Event Plugin Deployment

### LINKED ISSUE #152: Add Example Application for configuring a farm to use Deadline's SEP
Currently there isn't an example in the RFDK for configuring the Spot event Plugin.  While long term this is not our recommended way of auto scaling for now this is a stepping stone.

This will involve creating a new example farm and a writeup on the additional manual steps that are required to configure a render farm.

---

This is a 📕 documentation issue

grbartel:
Adds python and TS examples that create a farm that has the requirements to use Deadline's Spot Event Plugin.

Resolves #152 

----


examples/deadline/All-In-AWS-Infrastructure-SEP/python/package/lib/service_tier.py
```diff
@@ -0,0 +1,182 @@
...
+        :param stack_id: The ID of this construct.
+        :param props: The properties for this construct.
+        :param kwargs: Any kwargs that need to be passed on to the parent class.
+        """
+        super().__init__(scope, stack_id, **kwargs)
+
+        # A bastion host to connect to the render farm with.
+        # The bastion host is for convenience (e.g. SSH into RenderQueue and WorkerFleet instances).
+        # This is not a critical component of the farm, so can safely be removed.
+        self.bastion = BastionHostLinux(
```
ddneilson: imo, it would be better if we strip the example down to just what's needed for the SEP. It makes the delta for adding SEP components clearer. The example can be framed as demonstrating just the SEP parts, and that the user could add the parts to the basic example.

Also reduces the amount of copy/paste that exists between the examples.

examples/deadline/All-In-AWS-Infrastructure-SEP/python/package/lib/storage_tier.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+class StorageTier(Stack):
+    """
+    The storage tier of the render farm.
+    This stack contains all constructs that persist data which would be useful to keep between deployments.
+    There should little to no "business-logic" constructs in this stack.
+    """
+
+    def __init__(self, scope: Construct, stack_id: str, *, props: StorageTierProps, **kwargs):
```
ddneilson: Since the storage tier isn't relevant to the SEP we could either:
1. Leave it in, but empty with a comment that directs the customer to the basic example for a demonstration of the storage tier.
2. Remove the storage tier entirely, and leave a comment in the app that says that the storage tier was removed to focus the example and direct the customer to the basic example for the storage tier.

I.e. in either case, just use the self-contained Repository in the service tier.

ddneilson (changes requested):
This is incomplete, unless I'm missing something. 

Don't you think that we should also demonstrate creating the resources that the SEP will need? ex: A security group for the worker fleet? A LaunchConfiguration that is set up to connect to their RenderQueue, that they could then copy the json configuration from to copy into the SEP plugin settings?

examples/deadline/All-In-AWS-Infrastructure-SEP/README.md
```diff
@@ -0,0 +1,106 @@
+# RFDK Sample Application - Deadline Spot Event Plugin
+
+This is a sample RFDK application that deploys the basic infrastructure for a Deadline render farm. This application is structured in tiers, each representing a CloudFormation stack. The tiers are:
```
ddneilson: A whole bunch of this isn't accurate anymore.

examples/deadline/All-In-AWS-Infrastructure-SEP/python/README.md
```diff
@@ -0,0 +1,53 @@
...
+2. Stage the Docker recipes for `RenderQueue`:
+
+    ```bash
+    # Set this value to the version of RFDK your application targets
+    RFDK_VERSION=<version_of_RFDK>
+
+    # Set this value to the version of AWS Thinkbox Deadline you'd like to deploy to your farm. Deadline 10.1.9 and up are supported.
+    RFDK_DEADLINE_VERSION=<version_of_deadline>
+
+    npx --package=aws-rfdk@${RFDK_VERSION} stage-deadline \
```
ddneilson: There's a newer form for `stage-deadline` that doesn't require providing the full S3 URLs. Would probably be better to use that.

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/.gitignore
```diff
@@ -0,0 +1,20 @@
...
+# Exclude package artifacts
+dist
+*.tgz
+.LAST_PACKAGE
+
+#Defines license that must be present
+!license-header.js
+
+# The staged files for Deadline
+stage
```
ddneilson: nit: File should end with a newline.

There are a couple of files like this. I'll save you the copy/paste by not commenting on all of them.

ddneilson (changes requested):

examples/deadline/All-In-AWS-Infrastructure-SEP/python/setup.py
```diff
@@ -0,0 +1,25 @@
...
+
+    description="RFDK All In AWS Infrastructure SEP",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+
+    package_dir={"": "package"},
+    packages=setuptools.find_packages(where="package"),
+
+    install_requires=[
+        "aws-cdk.core==1.66.0",
```
ddneilson: These are now out of date. :-(

We're on CDK 1.70.0 & RFDK 0.19.0

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "watch": "tsc -w"
+  },
+  "devDependencies": {
+    "@types/node": "^14.6.1",
+    "aws-cdk": "1.66.0",
+    "ts-node": "^9.0.0",
+    "typescript": "~3.9.7"
+  },
+  "dependencies": {
+    "@aws-cdk/core": "1.66.0",
```
ddneilson: Same versioning issue as in the Python example.

examples/deadline/All-In-AWS-Infrastructure-SEP/python/setup.py
grbartel: Updated

ddneilson (dismissed):

examples/deadline/All-In-AWS-Infrastructure-SEP/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This sample application deploys a basic Deadline Render farm that is configured to use Deadlines [Spot Event Plugin](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/event-spot.html). 
+
+### Components
+
+All components in the render farm live within a [VPC](https://aws.amazon.com/vpc/).
+
+#### Repository
+
+The Repository component contains the database and file system that store persistent data used by Deadline. These resources are initialized by the Deadline Repository installer. The database in this example uses [Amazon DocumentDB](https://aws.amazon.com/documentdb/).
```
horsmand, line 21: Should link to the Deadline Repository documentation.

examples/deadline/All-In-AWS-Infrastructure-SEP/python/README.md
```diff
@@ -0,0 +1,50 @@
+# RFDK Sample Application - Deadline Spot Event Plugin - Python
+
+## Overview
+[Back to overview](../README.md)
+
+## Instructions
```
horsmand, line 6: We can get into a weird state on the mainline branch where examples can be updated to use features that haven't been officially released yet. For the python example, we need a note [like this](https://github.com/aws/aws-rfdk/blob/caccc0cafdb8b9bbb3d041019f320a112fa285bd/examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md) to explain this and recommend either switching to the release branch or packaging their local build of the repo and pip installing it over the pypi version.

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+    ```
+    yarn install
+    ```
+2. Modify the `deadline_ver` field in the `config` block of `package.json` as desired (Deadline 10.1.9 and up are supported), then stage the Docker recipes for `RenderQueue`:
+
+    ```
+    yarn stage
+    ```
+3. Build the sample app:
```
horsmand: This build won't succeed unless the build of the `aws-rfdk` package is done first. We should explain that before telling them to run the build of the example.

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/tsconfig.json
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  "include": [
+    "**/*.ts"
+  ],
+  "files": [
+    "package.json"
+  ],
+  "references": [
+    {
+      "path": "../../../../packages/aws-rfdk"
```
horsmand, line 32: Just pointing out that this reference here is what will cause the build of the example to fail if the build from the base directory hasn't run yet. We have an issue open to fix this https://github.com/aws/aws-rfdk/issues/218

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/README.md
```diff
@@ -0,0 +1,49 @@
+# RFDK Sample Application - Deadline - Typescript
+
+## Overview
+[Back to overview](../README.md)
+
+## Instructions
```
horsmand, line 6: We may want a note or step at the start to explain the possible differences between the example on the mainline branch and our releases [like this](https://github.com/aws/aws-rfdk/blob/2acb7c4b48d09f211efeb06127b9f3ecea7478d1/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/README.md)

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/lib/sep-stack.ts
```diff
@@ -0,0 +1,112 @@
...
+   * Initializes a new instance of {@link NetworkTier}.
+   * @param scope The scope of this construct.
+   * @param id The ID of this construct.
+   * @param props The stack properties.
+   */
+  constructor(scope: Construct, id: string, props: SEPStackProps) {
+    super(scope, id, props);
+    
+    //
+    // Network Tier
```
horsmand: We call out the network tier here, but none of the other tiers, which could make it confusing as to what this is actually referring to.

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "all-in-farm-sep",
+  "version": "0.19.0",
+  "bin": {
+    "app": "bin/app.js"
+  },
+  "config": {
+    "deadline_ver": "10.1.9.2",
```
horsmand: This can be shortened to `10.1` to always use the latest version matching that

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "stage_path": "stage"
+  },
+  "scripts": {
+    "build": "tsc",
+    "build+test": "yarn build && yarn test",
+    "cdk": "cdk",
+    "clean": "tsc --build --clean && bash ./clean.sh",
+    "stage": "stage-deadline --deadlineInstallerURI s3://thinkbox-installers/Deadline/${npm_package_config_deadline_ver}/Linux/DeadlineClient-${npm_package_config_deadline_ver}-linux-x64-installer.run --dockerRecipesURI s3://thinkbox-installers/DeadlineDocker/${npm_package_config_deadline_ver}/DeadlineDocker-${npm_package_config_deadline_ver}.tar.gz --output ${npm_package_config_stage_path}",
```
horsmand: This can be updated to 
```
stage-deadline ${npm_package_config_deadline_ver} --output ${npm_package_config_stage_path}
```

examples/deadline/All-In-AWS-Infrastructure-SEP/README.md
grbartel: The repository doesn't really have it's own documentation.  This also matches the other example.

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/package.json
grbartel: updated

grbartel:
@horsmand This has all been updated

examples/deadline/All-In-AWS-Infrastructure-SEP/python/README.md
```diff
@@ -0,0 +1,64 @@
...
+---
+
+1.  This sample app on the `mainline` branch may contain features that have not yet been officially released, and may not be available in the `aws-rfdk` package installed through pip from PyPI. To work from an example of the latest release, please switch to the `release` branch. If you would like to try out unreleased features, you can stay on `mainline` and follow the instructions for building, packing, and installing the `aws-rfdk` from your local repository.
+2.  Install the dependencies of the sample app:
+
+    ```bash
+    pip install -r requirements.txt
+    ```
+3.  If working on the `release` branch, this step can be skipped. If working on `mainline`, navigate to the base directory where the build and packaging scripts are, then run them and install the result over top of the `aws-rfdk` version that was installed in the previous step:
+    ```bash
```
horsmand, line 22: This code block is missing the closing ` ``` `

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/README.md
```diff
@@ -0,0 +1,62 @@
...
+These instructions assume that your working directory is `examples/deadline/All-In-AWS-Infrastructure-SEP/ts/` relative to the root of the RFDK package.
+
+---
+1. This sample app on the mainline branch may contain features that have not yet been officially released, and may not be available in the aws-rfdk package installed through npm from npmjs. To work from an example of the latest release, please switch to the release branch. If you would like to try out unreleased features, you can stay on mainline and follow the instructions for building, packing, and installing the aws-rfdk from your local repository.
+2. Install the dependencies of the sample app:
+
+    ```
+    yarn install
+    ```
+3.  If working on the `release` branch, this step can be skipped. If working on `mainline`, navigate to the base directory where the build and packaging scripts are, then run them and install the result over top of the `aws-rfdk` version that was installed in the previous step:
```
horsmand: We can simplify this down like in here: https://github.com/aws/aws-rfdk/blob/2acb7c4b48d09f211efeb06127b9f3ecea7478d1/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/README.md

We don't need to npm install the local package since we point to the local package location as a reference in the `tsconfig.json` already. Running `build.sh` before running `npm run build` in the example will be enough to get it working and we have an issue to avoid having to do that too https://github.com/aws/aws-rfdk/issues/218

Sorry if  I gave you bad instructions on my last review, I thought I linked you to the readme example I shared here, but maybe it was the old instructions that did suggest to do the `npm install` like this.

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Instructions
+
+---
+**NOTE**
+
+These instructions assume that your working directory is `examples/deadline/All-In-AWS-Infrastructure-SEP/ts/` relative to the root of the RFDK package.
+
+---
+1. This sample app on the mainline branch may contain features that have not yet been officially released, and may not be available in the aws-rfdk package installed through npm from npmjs. To work from an example of the latest release, please switch to the release branch. If you would like to try out unreleased features, you can stay on mainline and follow the instructions for building, packing, and installing the aws-rfdk from your local repository.
```
horsmand: This still mentions needing to pack and install the local version, which isn't required.

examples/deadline/All-In-AWS-Infrastructure-SEP/ts/README.md
```diff
@@ -0,0 +1,49 @@
...
+3. Modify the `deadline_ver` field in the `config` block of `package.json` as desired (Deadline 10.1.9 and up are supported), then stage the Docker recipes for `RenderQueue`:
+
+    ```
+    yarn stage
+    ```
+4. Build the sample app:
+
+    ```
+    yarn build
+    ```
```
horsmand: This step won't work without running the base's `build.sh` first.

examples/deadline/All-In-AWS-Infrastructure-SEP/python/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Instructions
+
+---
+**NOTE**
+
+These instructions assume that your working directory is `examples/deadline/All-In-AWS-Infrastructure-SEP/python/` relative to the root of the AWS-RFDK package.
+
+---
+
+1.  This sample app on the `mainline` branch may contain features that have not yet been officially released, and may not be available in the `aws-rfdk` package installed through pip from PyPI. To work from an example of the latest release, please switch to the `release` branch. If you would like to try out unreleased features, you can stay on `mainline` and follow the instructions for building, packing, and installing the `aws-rfdk` from your local repository.
```
horsmand, line 15: This still mentions needing to build, pack, and install the local aws-rfdk if working from mainline, which is correct, but the instructions for doing that aren't here now.

horsmand (changes requested):
It seems you had misunderstood my last comment and removed all the additional steps for building and deploying both python and npm packages. The comment was specific to the ts example and suggested to keep the part about how to run build.sh before running the example's build.

Python needs the step you removed, where we explain how to install the locally built and packaged version of aws-rfdk if working on mainline. This is because python will always pull the package from pypi and while the release branch example will always match the package in pypi, the mainline branch will be out of sync when any unreleased features are added.
An example of these instructions are in step 3 [here](https://github.com/aws/aws-rfdk/tree/2acb7c4b48d09f211efeb06127b9f3ecea7478d1/examples/deadline/All-In-AWS-Infrastructure-Basic/python).

TypeScript needs the locally built aws-rfdk package, regardless of which branch is being worked off. This is because we refer to the local package from the example's tsconfig.json, always using it and never trying to fetch anything from npmjs.
An example of these instructions are in step 12 [here](https://github.com/aws/aws-rfdk/tree/2acb7c4b48d09f211efeb06127b9f3ecea7478d1/examples/deadline/All-In-AWS-Infrastructure-Basic/ts).

examples/deadline/All-In-AWS-Infrastructure-SEP/python/setup.py
```diff
@@ -0,0 +1,25 @@
...
+    description="RFDK All In AWS Infrastructure SEP",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+
+    package_dir={"": "package"},
+    packages=setuptools.find_packages(where="package"),
+
+    install_requires=[
+        "aws-cdk.core==1.70.0",
+        "aws-rfdk==0.19.0"
```
horsmand: We're on CDK 1.72.0 and RFDK 1.20.0 now. Needs to be updated here and in the ts example's package.json.

horsmand (approved):

ddneilson (approved):

Issue #200, opened 2020-10-30
Update version bump script to add CDK version to changelog


horsmand:
Running the version bump script should update the changelog to include a section that displays the version of CDK that the new version of RFDK is depending on.

### Use Case

Currently the process of adding the version of CDK is done manually after the bump script is ran. We would like to automate this process and combine it into the bump script.

### Proposed Solution

This can be accomplished by pulling the CDK version from the `package.json` using `node` and then using `sed` in `bash.sh` to find the correct spot in the changelog and add in the new section.

* [x] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

PR #201, opened 2020-10-30
chore: updates to the bump script

### LINKED ISSUE #173: Update version bump script to bump python example too
Currently our [bump script](https://github.com/aws/aws-rfdk/blob/b479e89cfdb0fe5b00af426198fc68dff8c22dae/bump.sh) can bump all of the lerna packages contained in our repository, but not the python package contained in the `examples` directory.

### Use Case

When doing a version bump, we need to be able to bump the python version at the same time. This will save time when doing releases and make it impossible to forget to do, which can happen when it's a manual process.

### Proposed Solution

Update the bump script so it updates the RFDK version used in `examples/deadline/All-In-AWS-Infrastructure-Basic/python`. We should take the time to evaluate if we can do this in a way that will also cover any other python examples that get added to the `examples` directory in the future.

### Other

Pull request https://github.com/aws/aws-rfdk/pull/171 can be used as a reference to see where the version bump needs to be done. The AWS CDK version bump in this PR can be ignored. We are currently doing CDK version upgrades manually, and that was a separate miss that it didn't get updated when the rest of the repository had its CDK version updated.

---

horsmand:
Fixes #173 and #200

* Added automation for bumping the RFDK version in the python example
* Added automation for adding a section to the changelog that lists the
version of CDK that this version of RFDK depends on

I tested this by running the bump script using both `minor` and `patch` to test both kinds of version bumps.

----

### DIFF
--- a/bump.sh
+++ b/bump.sh
@@ -43,8 +43,11 @@ npx standard-version --skip.tag=true --commit-all
 # Get the new version number to do some manual find and replaces
 new_version=$(node -p "require('./package.json').version")
 
-# Update the version of RFDK used in the python example
-sed -i "s/\"aws-rfdk==[0-9]*\.[0-9]*\.[0-9]*\"/\"aws-rfdk==$new_version\"/" ./examples/deadline/All-In-AWS-Infrastructure-Basic/python/setup.py
+# Update the version of RFDK used in the python examples
+for exampleSetupPy in $(find ./examples/ -name 'setup.py')
+do
+  sed -i "s/\"aws-rfdk==[0-9]*\.[0-9]*\.[0-9]*\"/\"aws-rfdk==$new_version\"/" "$exampleSetupPy"
+done
 
 # When standard-version adds a patch release to the changelog, it makes it a smaller header size. This undoes that.
 if [[ $version == "patch" ]]; then
--- a/bump.sh
+++ b/bump.sh
@@ -22,13 +22,13 @@ version=${1:-minor}
 
 cd "$(dirname "$0")"
 
-echo "Starting ${version} version bump"
+echo "Starting $version version bump"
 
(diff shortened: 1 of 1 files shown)


bump.sh
```diff
@@ -39,3 +39,24 @@ npx lerna version ${version} --yes --exact --no-git-tag-version --no-push
 
 # Generate CHANGELOG and create a commit
 npx standard-version --skip.tag=true --commit-all
+
+# Get the new version number to do some manual find and replaces
+new_version=$(node -p "require('./package.json').version")
+
+# Update the version of RFDK used in the python example
+sed -i "s/\"aws-rfdk==[0-9]*\.[0-9]*\.[0-9]*\"/\"aws-rfdk==$new_version\"/" ./examples/deadline/All-In-AWS-Infrastructure-Basic/python/setup.py
```
ddneilson: This is good for the single example, but we could probably make it more general:

```bash
for exampleSetupPy in $(find ./examples/ -name 'setup.py')
do
  sed -i "s/\"aws-rfdk==[0-9]*\.[0-9]*\.[0-9]*\"/\"aws-rfdk==$new_version\"/" "$exampleSetupPy"
done
```

ddneilson (changes requested):

jericht (dismissed):

bump.sh
horsmand: Fixed

jusiskin (approved):
Checked out and tested with both a minor and patch bumps and confirmed it works as expected. Great work, @horsmand. Approved.

jericht (approved):

bump.sh
jusiskin: Resolved

ddneilson (approved):

PR #205, opened 2020-10-30
chore(integ): configure tests to run in pipeline

### LINKED ISSUE #199: Enable integration tests to be run in a pipeline
Currently the integration tests are configured in a way that allows people with the aws-rfdk repository checked out locally to manually configure and run them. However, if someone wants to automate them to be run in the testing step of a pipeline, it isn't currently possible. We need some tooling around allowing that to be possible.

### Use Case

I want to be able to set a pipeline that pulls the aws-rfdk repository, runs the build and package, and then starts the integration tests with the artifacts creating from the packaging step.

### Proposed Solution

To accomplish this we need to:
1. Package the integration tests and anything they require to be used during the build/package step so they are accessible alongside the aws-rfdk package in any following pipeline steps.
1. Write a script that can be used as the entry point for the testing step that will do any necessary configuration and then execute `rfdk-integ-e2e.sh` to run the tests.
1. Make sure that if anything fails, the tests clean up after themselves.

---

horsmand:
Fixes #199

Tested by setting up a pipeline with the integ tests in a testing stage and getting a successful run, and also running the tests manually, from the command line like we have been doing for previous releases, to make sure the changes didn't affect that workflow.

----


integ/scripts/bash/run-e2e-from-pipeline.sh
```diff
@@ -0,0 +1,36 @@
...
+scriptdir=$(cd $(dirname $0) && pwd)
+
+cd "$scriptdir"/../..
+
+yarn
+
+# Basic integ test configuration
+export SKIP_TEST_CONFIG=true
+export DEADLINE_VERSION="10.1.10.6"
+export USER_ACCEPTS_SSPL_FOR_RFDK_TESTS=true
```
jusiskin: We need to remove this since we don't want to accept Mongo SSPL license on anyone's behalf if they checkout this repository and try running this.

Instead, those building a pipeline that runs these tests should decide whether to accept the SSPL for themselves and configure this environment variable in their pipeline.

We could put a sanity check that happens here and fails fast if the environment variable is not set to accept the SSPL and display a user-friendly error message.

integ/components/deadline/common/scripts/bash/component_e2e.sh
```diff
@@ -19,12 +27,15 @@ if [[ ! "${SKIP_TEST_CHECK}" = "true" ]]; then
...
-    
+
   if [[ $OPTION != '--destroy-only' ]]; then
+    run_hook
     deploy_component_stacks $COMPONENT_NAME
+    run_hook
     execute_component_test $COMPONENT_NAME
   fi
   if [[ $OPTION != '--deploy-and-test-only' ]]; then
+    run_hook
```
jusiskin: nit: the re-use of the `PRE_COMPONENT_HOOK` to happen before each phase changes the semantics of the name of the hook. Should we either:

1.  Rename the hook to something like `PRE_COMPONENT_PHASE_HOOK`?
1.  Make separate hooks for each phase?
    *   `PRE_COMPONENT_DEPLOY_HOOK`
    *   `PRE_COMPONENT_TEST_HOOK`
    *   `PRE_COMPONENT_DESTROY_HOOK`

    In the case of our pipeline, we'd point them to the same bash function to assume the role.


Same feedback applies to the other scripts where we use  the hook:
*   `integ/scripts/bash/cleanup.sh`
*   `integ/scripts/bash/tear-down.sh`
*   `integ/scripts/bash/teardown-infrastructure.sh`

integ/components/deadline/common/scripts/bash/component_e2e.sh
```diff
@@ -5,6 +5,14 @@
 
 set -euo pipefail
 
+run_hook() {
+    # Invoke hook function if it is exported and name is defined in PRE_COMPONENT_HOOK variable
+    if [ ! -z "${PRE_COMPONENT_HOOK+x}" ]  && [ "$(type -t $PRE_COMPONENT_HOOK)" == "function" ]
+    then
+      $PRE_COMPONENT_HOOK
```
jusiskin: In the original place we introduced this hook

https://github.com/aws/aws-rfdk/blob/b4bfd4ec475c3e94002ac961c3b2ce6c8abd65ee/integ/scripts/bash/rfdk-integ-e2e.sh#L79,

we call the hook function with the name of the component being tested as the first argument. We should probably make this consistent and either:

1.  supply the component name here
1.  remove the component name from being passed in `integ/scripts/bash/rfdk-integ-e2e.sh`

Same feedback applies to other places you've added calls to `PRE_COMPONENT_HOOK`

integ/scripts/bash/run-against-dist
```diff
@@ -0,0 +1,42 @@
...
+# - Install the CLI from that Verdaccio fixture somewhere and put it
+#   on the PATH.
+# - Prepare the various package managers to take their packages from dist/ as well.
+#
+# Parameter: DIST_ROOT, if different than the current directory.
+set -euo pipefail
+scriptdir=$(cd $(dirname $0) && pwd)
+source $scriptdir/run-against-dist.bash
+
+# If DIST_ROOT is not given, go up two levels out of examples/kitchen-sink and check if the
```
jusiskin: nit: looks like a copy/paste leftover from the kitchen sink

integ/components/deadline/deadline_04_workerFleetHttps/bin/deadline_04_workerFleetHttps.ts
```diff
@@ -0,0 +1,50 @@
+/**
```
jusiskin, line 1: There is a lot of duplication from the HTTP Worker fleet tests. Is there any way we can make that code common and only vary the specifics to HTTP(S) to remove the duplication?

integ/jest.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   roots: ['<rootDir>'],
-  testMatch: ['**/*.test.ts'],
+  testMatch: ['**/*.test.[jt]s'],
   transform: {
     '^.+\\.tsx?$': 'ts-jest'
   },
```
jusiskin: Does this make both the `.ts` and `.js` tests run? It looks like this pull request is trying to standardize on running only the javascript files except in this one instance. Unless I'm missing something, we can probably restrict this to `**/*.test.js`.

integ/pack.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+dir=$PWD
+filename=$(npm pack)
+
+echo "Copying $filename to $dir/../dist/js"
+cp "$filename" "$dir/../dist/js/"
```
jusiskin: As discussed offline, there may have been concern that by adding packaging to the `integ` package, there may be side-effects on the publishing and API reference doc generation phases of our pipeline.

Adding this comment here to continue this discussion. The same feedback applies to `pkglint`.

integ/components/deadline/deadline_04_workerFleetHttps/lib/workerFleetHttps-testing-tier.ts
```diff
@@ -0,0 +1,74 @@
...
+ *   At execution the tests retrieve the value of secrets for the authentication cert.
+ */
+export class WorkerFleetTestingTier extends TestingTier {
+  constructor(scope: Construct, id: string, props: WorkerFleetTestingTierProps) {
+    super(scope, id, props);
+
+    const structs = props.structs;
+    structs.forEach( workerStruct => {
+
+      const testSuiteId = 'WFS' + (structs.indexOf(workerStruct) + 1).toString();
```
jusiskin, line 45: This file is almost an exact copy of `integ/components/deadline/deadline_03_workerFleetHttp/lib/workerFleetHttp-testing-tier.ts`. The only functional change is this `testSuiteId`. Can we make this a shared file and take that as a prop?

integ/scripts/bash/run-against-dist
```diff
@@ -0,0 +1,42 @@
...
+dist_root=$(cd ${DIST_ROOT:-.} && pwd)
+
+if [[ ! -f $dist_root/build.json ]]; then
+  echo "$dist_root does not seem to be a built RFDK distribution (change directory or use DIST_ROOT)" >&2
+  exit 1
+fi
+
+tempdir=$(mktempdir)
+
+serve_npm_packages $tempdir
```
jusiskin: Do we actually require verdaccio? Wondering if we can avoid all of that by simply installing the built packages with `npm`:

```bash
# install dependencies
npm install
# overwrite production ones with packaged ones
npm install path/to/built/package.tgz
```

jusiskin (changes requested):
The main feedback is to:

*   investigate downstream impacts of packaging `integ` and `pkglint`
*   remove the duplication between the HTTP/HTTPS Worker fleet tests
*   see if we can eliminate the need for verdaccio

integ/scripts/bash/run-against-dist
horsmand: I believe the first `npm install` will fail due to not finding all the required packages (i.e. pkglint). This is irrelevant now that I switched away from using packaging.

integ/pack.sh
horsmand: No longer packing these packages.

integ/jest.config.js
horsmand: Since I moved away from packaging the tests I changed it back to just use ts.

integ/scripts/bash/run-against-dist
horsmand: Removed this script.

integ/components/deadline/common/scripts/bash/component_e2e.sh
horsmand: I'm not sure what the component name would need to be supplied for, so I've removed it.

integ/components/deadline/common/scripts/bash/component_e2e.sh
horsmand: I don't see a need for a separate hook for each phase so I'll change the name as you suggested. I'm thinking about calling it `PRE_AWS_INTERACTION_HOOK` since it is meant to run before any interaction with AWS.

integ/scripts/bash/run-e2e-from-pipeline.sh
horsmand: I've moved this SSPL acceptance out of here and injected it as an env var from the pipeline side.

integ/components/deadline/deadline_04_workerFleetHttps/bin/deadline_04_workerFleetHttps.ts
horsmand: I'm hesitant to do any refactoring to remove duplicate code at this time. 
* There are differences in the requirement for certificates that would need to be added into the HTTPS app (I had left these in my original PR but my next update will remove them from the HTTP app).
* We've got 2 copies, the rule of 3 would say to wait for another file that requires similar code. When we make these run in parallel, it might make sense to split the platforms out into their own CDK apps too to minimize the run time. I'd rather look at how to refactor once at that point than doing it now and then again later.
* It's also unclear to me if these two apps will always need to be kept in sync. If changes don't need to be applied to both of them, refactoring out the duplicate code would just make more difficult.

I don't have very strong feelings against doing this, so if you still think it's a good idea I'll do the refactoring now.

integ/components/deadline/common/scripts/bash/component_e2e.sh
jusiskin: Sounds good. Let's go with this

integ/components/deadline/common/scripts/bash/component_e2e.sh
jusiskin: This is fine by me - especially in light of the name change to `PRE_AWS_INTERACTION_HOOK`.

integ/components/deadline/deadline_04_workerFleetHttps/bin/deadline_04_workerFleetHttps.ts
jusiskin: These tests will rarely require modifications so long as our API remains stable and since you mentioned we plan to revisit this code soon for parallel execution, we can defer refactoring until then. Resolved.

integ/components/deadline/deadline_03_workerFleetHttp/bin/deadline_03_workerFleetHttp.ts
```diff
@@ -0,0 +1,65 @@
...
+
+// Get unique tag for this integration test from environment variable
+const integStackTag = process.env.INTEG_STACK_TAG!.toString();
+
+// Worker fleets with their own repository and render queue will be created for each permutation of OS and protocol
+const oss = ['Linux','Windows'];
+
+let structs: Array<WorkerStruct> = [];
+let i = 1;
+oss.forEach( os => {
```
jusiskin: nit: to avoid manually incrementing the counter variable (`i`), the callback to `forEach` can take an additional positional argument that contains the 0-based index. See [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#Parameters).

Same feedback applies to other places we are doing this.

integ/components/deadline/deadline_04_workerFleetHttps/test/deadline_04_workerFleetHttps.test.ts
```diff
@@ -84,6 +85,7 @@ describe.each(testCases)('Deadline WorkerFleet tests (%s)', (_, id) => {
       return awaitSsmCommand(bastionId, params);
     }
     else {
+      console.error(`Did not find a secrect ARN for ${testingStackName}`);
       return;
```
jusiskin: It probably warrants throwing an exception here since these tests are only run for HTTPS, right?

integ/lib/render-struct.ts
```diff
@@ -45,7 +45,7 @@ export class RenderStruct extends Construct {
     }
 
     const host = 'renderqueue';
-    const zoneName = Stack.of(this).stackName + '.local';
+    const zoneName = Stack.of(this).stackName.slice(0, 45) + '.local';
```
jusiskin: I know this gave you headaches to root-cause. Can we add a comment here so these efforts are documented? Same applies to the other occurrence in `integ/lib/testing-tier.ts`

integ/scripts/bash/run-e2e-automated.sh
```diff
@@ -0,0 +1,44 @@
...
+    unset AWS_SECRET_ACCESS_KEY
+    unset AWS_SESSION_TOKEN
+    creds=$(mktemp -d)/creds.json
+    AWS_STS_REGIONAL_ENDPOINTS=regional aws sts assume-role \
+        --role-arn $ROLE_ARN \
+        --role-session-name $ROLE_SESSION_NAME \
+        --external-id $ROLE_EXTERNAL_ID > $creds
+    export AWS_ACCESS_KEY_ID="$(cat ${creds} | grep "AccessKeyId" | cut -d'"' -f 4)"
+    export AWS_SECRET_ACCESS_KEY="$(cat ${creds} | grep "SecretAccessKey" | cut -d'"' -f 4)"
+    export AWS_SESSION_TOKEN="$(cat ${creds} | grep "SessionToken" | cut -d'"' -f 4)"
```
jusiskin: Instead of writing to a file (`$(mktemp -d)/creds.json`), it is safer to just keep this in memory. Aside from other processes/agents being able to read the file, if the EBS volume is unencrypted and the underlying block is reused for another instance, a motivated attacker may be able to recover the credentials. The credentials are time-limited, but it is still good practice and doesn't hurt.

I believe you should be able to do this with:

```bash
export CREDS="$(AWS_STS_REGIONAL_ENDPOINTS=regional aws sts assume-role \
        --role-arn $ROLE_ARN \
        --role-session-name $ROLE_SESSION_NAME \
        --external-id $ROLE_EXTERNAL_ID)"
export AWS_ACCESS_KEY_ID="$(printenv CREDS | grep "AccessKeyId" | cut -d'"' -f 4)"
export AWS_SECRET_ACCESS_KEY="$(printenv CREDS | grep "SecretAccessKey" | cut -d'"' -f 4)"
export AWS_SESSION_TOKEN="$(printenv CREDS | grep "SessionToken" | cut -d'"' -f 4)"
# Clean up the env var
unset CREDS
```

We do this already when [configuring the database connection for the repo installer](https://github.com/aws/aws-rfdk/blob/f65591f1974f603de01fae20e3566b11c77cffa9/packages/aws-rfdk/lib/deadline/lib/database-connection.ts#L184-L187). The `printenv` is used to avoid having the credentials logged in shell history and observable in procfs.

jusiskin (changes requested):
Looking great. Just a few minor things left.

integ/scripts/bash/run-e2e-automated.sh
horsmand: Updated

integ/lib/render-struct.ts
horsmand: Added

integ/components/deadline/deadline_04_workerFleetHttps/test/deadline_04_workerFleetHttps.test.ts
horsmand: Yeah, makes sense, will change it.

integ/components/deadline/deadline_03_workerFleetHttp/bin/deadline_03_workerFleetHttp.ts
horsmand: Fixed.

integ/components/deadline/deadline_04_workerFleetHttps/test/deadline_04_workerFleetHttps.test.ts
jusiskin: Looks good

integ/lib/render-struct.ts
jusiskin: Thanks!

jusiskin (approved):
Great work! Approved!

integ/scripts/bash/run-e2e-automated.sh
```diff
@@ -0,0 +1,48 @@
...
+    export AWS_ACCESS_KEY_ID="$(printenv CREDS | grep "AccessKeyId" | cut -d'"' -f 4)"
+    export AWS_SECRET_ACCESS_KEY="$(printenv CREDS | grep "SecretAccessKey" | cut -d'"' -f 4)"
+    export AWS_SESSION_TOKEN="$(printenv CREDS | grep "SessionToken" | cut -d'"' -f 4)"
+    # Clean up the env var
+    unset CREDS
+}
+
+# Basic integ test configuration
+export SKIP_TEST_CONFIG=true
+export DEADLINE_VERSION="10.1.11.5"
```
ddneilson, line 35: Could/should we always be using the latest version of Deadline? ex: If we set this to "10.1" then staging and such should pull the latest Deadline.

integ/scripts/bash/stage-deadline.sh
```diff
@@ -8,7 +8,7 @@ set -euo pipefail
 # Check if Deadline assets are staged at configured path
 if [ ! $(ls "$DEADLINE_STAGING_PATH/manifest.json" 2> /dev/null) ]; then
     # Stage Deadline assets
-    npx ../packages/aws-rfdk/bin/stage-deadline --deadlineInstallerURI "s3://thinkbox-installers/Deadline/$DEADLINE_VERSION/Linux/DeadlineClient-$DEADLINE_VERSION-linux-x64-installer.run" --dockerRecipesURI "s3://thinkbox-installers/DeadlineDocker/$DEADLINE_VERSION/DeadlineDocker-$DEADLINE_VERSION.tar.gz" --output "$DEADLINE_STAGING_PATH"
+    npx stage-deadline --deadlineInstallerURI "s3://thinkbox-installers/Deadline/$DEADLINE_VERSION/Linux/DeadlineClient-$DEADLINE_VERSION-linux-x64-installer.run" --dockerRecipesURI "s3://thinkbox-installers/DeadlineDocker/$DEADLINE_VERSION/DeadlineDocker-$DEADLINE_VERSION.tar.gz" --output "$DEADLINE_STAGING_PATH"
```
ddneilson, line 11: Could we use the newer arg form of stage-deadline here that doesn't require the full URIs? Or is there a use-case for this code that requires using unreleased versions of Deadline?

integ/scripts/bash/run-e2e-automated.sh
jusiskin: This is being tackled in #160. It looks like this one is lined up to be merged first. Once this is merged lets make a comment on that one to make sure the change is applied here.

integ/scripts/bash/stage-deadline.sh
jusiskin: Also part of #160.

EDIT: This one is easy to apply here as well though without any conflict.

ddneilson (approved):

PR #209, opened 2020-11-04
feat(deadline): add WorkerInstanceConfiguration construct

### LINKED ISSUE #208: Construct for configuring Deadline Workers
We've had an ask from an internal team for a convenient way to configure an arbitrary Deadline Worker in the same way that we configure the workers within a WorkerInstanceFleet.

### Use Case

Creating a Launch Template for use with Deadline's Spot Event Plugin

### Proposed Solution

Refactoring the CloudWatch log stream & Worker settings steps out of the WorkerInstanceFleet code into a self-contained WorkerConfiguration construct/class.

### Other

N/A

---

ddneilson:
Fixes: https://github.com/aws/aws-rfdk/issues/208

This is a construct that provides helper methods for configuring a Deadline Worker in the RFDK context.

## Testing

This was tested by deploying both a WorkerInstanceFleet and a single instance using the WorkerConfiguration construct. Deployments were in the context of the simple app that's in the RFDK docs here -- https://docs.aws.amazon.com/rfdk/latest/guide/what-is-rfdk.html#why-use-the-rfdk . After deployment, I ensured that the setup was correct via the generated CloudWatch logs.

Python test code for the test was:

```python
import os
import jsii
import aws_cdk.core as core
import aws_cdk.aws_ec2 as ec2
import aws_cdk.aws_iam as iam
import aws_cdk.aws_efs as efs
import aws_rfdk as rfdk_core
import aws_rfdk.deadline as rfdk_deadline

@jsii.implements(rfdk_deadline.IHost)
class WorkerInstance(core.Construct):
  def __init__(self, scope: core.Construct, id: str, *, vpc: ec2.Vpc, render_queue: rfdk_deadline.RenderQueue) -> None:
    super().__init__(scope, id)

    deployment_region = core.Stack.of(self).region

    machine_image=ec2.MachineImage.generic_linux({
      deployment_region: "ami-05d4887175201bde8" # Deadline 10.1.10.6 AWSPortal Worker for Linux, us-west-2
    })
    self.os_type = machine_image.get_image(self).os_type
    self.grant_principal = iam.Role(self, 'WorkerRole', assumed_by=iam.ServicePrincipal('ec2.amazonaws.com'))
    self.user_data = ec2.UserData.for_linux()
    security_group = ec2.SecurityGroup(self, 'WorkerSecGrp',
      vpc=vpc,
      description='Security group for Fleet [fleet identifier]'
    )
    self.connections = ec2.Connections(security_groups=[security_group])

    # Note: Must configure userData before we render it
    config = rfdk_deadline.WorkerInstanceConfiguration(self, 'Config', 
      worker=self,
      render_queue=render_queue,
      cloudwatch_log_settings=rfdk_core.LogGroupFactoryProps(
        log_group_prefix='/test-worker/'
      ),
      worker_settings=rfdk_deadline.WorkerSettings(
        groups= ['g1'],
      )
    )

    worker_subnet = vpc.select_subnets(subnet_type=ec2.SubnetType.PRIVATE).subnets[0]
    worker_instance = ec2.CfnInstance(self, 'Worker', 
      image_id=machine_image.get_image(self).image_id,
      instance_type='t3.small',
      security_group_ids=[security_group.security_group_id],
      iam_instance_profile=iam.CfnInstanceProfile(self, 'WorkerProfile', 
        roles=[self.grant_principal.role_name],
      ).ref,
      user_data=core.Fn.base64(self.user_data.render()),
      subnet_id=worker_subnet.subnet_id,
      availability_zone=worker_subnet.availability_zone,
    )

  # =====================================
  # Properties for the IHost interface = (IConnection, IGrantable, IScriptHost)

  # --------
  #  ec2.IConnection
  @property
  def connections(self):
    return self._connections

  @connections.setter
  def connections(self, value):
    self._connections = value

  # -------
  #  iam.IGrantable
  @property
  def grant_principal(self):
    return self._grant_principal

  @grant_principal.setter
  def grant_principal(self, value):
    self._grant_principal = value

  # -----
  #  rfdk.IScriptHost
  @property
  def os_type(self):
    return self._os_type

  @os_type.setter
  def os_type(self, value):
    self._os_type = value

  @property
  def user_data(self):
    return self._user_data

  @user_data.setter
  def user_data(self, value):
    self._user_data = value

class InfrastructureStack(core.Stack):

  def __init__(self, scope: core.Construct, id: str, **kwargs) -> None:
    super().__init__(scope, id, **kwargs)

    self.vpc = ec2.Vpc(self, "Vpc", max_azs=2)

    script_dir = os.path.dirname(os.path.abspath(__file__))
    stage_dir = os.path.abspath(os.path.join(script_dir, '..', 'stage'))
    local_recipe_stage = rfdk_deadline.Stage.from_directory(stage_dir)
    server_recipes = rfdk_deadline.ThinkboxDockerRecipes(self, 'ServerImages',
      stage=local_recipe_stage
    )

    self.repository = rfdk_deadline.Repository(self, 'Repository',
      vpc=self.vpc,
      version=server_recipes.version,
      # Allow resources to be deleted when we delete the sample
      removal_policy=rfdk_deadline.RepositoryRemovalPolicies(
        database=core.RemovalPolicy.DESTROY,
        filesystem=core.RemovalPolicy.DESTROY
      )
    )
    
    self.render_queue = rfdk_deadline.RenderQueue(self, 'RenderQueue',
      vpc=self.vpc, 
      version=server_recipes.version,
      images=server_recipes.render_queue_images,
      repository=self.repository,
      # Allow the load-balancer to be deleted when we delete the sample
      deletion_protection=False,
    )



class WorkersStack(core.Stack):

  def __init__(self, scope: core.Construct, id: str, *, render_queue, vpc, **kwargs) -> None:
    super().__init__(scope, id, **kwargs)

    worker = WorkerInstance(self, 'Worker', vpc=vpc, render_queue=render_queue)

    health_monitor = rfdk_core.HealthMonitor(self, 'HealthMonitor',
      vpc=vpc,
      deletion_protection=False,
    )

    fleet = rfdk_deadline.WorkerInstanceFleet(self, 'LinWorkers',
      vpc=vpc, 
      render_queue=render_queue,
      worker_machine_image=ec2.MachineImage.generic_linux({
        # Fill in your AMI id here
        core.Stack.of(self).region: "ami-05d4887175201bde8" # Deadline 10.1.10.6 AWSPortal Worker for Linux, us-west-2
      }),
      min_capacity=5,
      instance_type=ec2.InstanceType("c5.large"),
      spot_price=0.15,
      health_monitor=health_monitor
    )

    fleet2 = rfdk_deadline.WorkerInstanceFleet(self, 'WinWorkers',
      vpc=vpc, 
      render_queue=render_queue,
      worker_machine_image=ec2.MachineImage.generic_windows({
        # Fill in your AMI id here
        core.Stack.of(self).region: "ami-09c712180218564f2" # Deadline 10.1.10.6 AWSPortal Worker for Windows, us-west-2
      }),
      min_capacity=5,
      instance_type=ec2.InstanceType("c5.large"),
      spot_price=0.25,
      health_monitor=health_monitor
    )
```

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
@@ -1,0 +1,162 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as path from 'path';
+
+import {
+  OperatingSystemType,
+} from '@aws-cdk/aws-ec2';
+import {
+  IGrantable,
+} from '@aws-cdk/aws-iam';
+import {
+  Construct,
+  Duration,
+} from '@aws-cdk/core';
+import {
+  CloudWatchAgent,
+  CloudWatchConfigBuilder,
+  IScriptHost,
+  LogGroupFactory,
+  LogGroupFactoryProps,
+  ScriptAsset,
+} from '../../core';
+import { Version } from './version';
+
+/**
+ * Interface for Deadline clients that can be configured via the ClientConfiguration
+ * helper class.
+ */
+export interface IConfigurableWorker extends IScriptHost, IGrantable {
+}
+
+/**
+ * The Deadline Worker settings that can be configured via the configureWorkerSettings method
+ * of the WorkerConfiguration helper class.
+ */
+export interface WorkerSettings {
+  /**
+   * Deadline groups these workers needs to be assigned to. The group is
+   * created if it does not already exist.
+   *
+   * @default - Worker is not assigned to any group
+   */
+  readonly groups?: string[];
+
+  /**
+   * Deadline pools these workers needs to be assigned to. The pool is created
+   * if it does not already exist.
+   *
+   * @default - Worker is not assigned to any pool.
+   */
+  readonly pools?: string[];
+
+  /**
+   * Deadline region these workers needs to be assigned to.
+   *
+   * @default - Worker is not assigned to any region
+   */
+  readonly region?: string;
+}
+
+/**
+ * This is a helper construct for configuring Deadline Workers to connect to a RenderQueue, send their
+ * log files to CloudWatch, and similar common actions.
+ *
+ * Security Considerations
+ * ------------------------
+ * - The instances configured by this construct will download and run scripts from your CDK bootstrap bucket when that instance
+ *   is launched. You must limit write access to your CDK bootstrap bucket to prevent an attacker from modifying the actions
+ *   performed by these scripts. We strongly recommend that you either enable Amazon S3 server access logging on your CDK
+ *   bootstrap bucket, or enable AWS CloudTrail on your account to assist in post-incident analysis of compromised production
+ *   environments.
+ */
+export class WorkerConfiguration extends Construct {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+  }
+
+  /**
+   * This method can be used to configure a Deadline Worker instance to stream its logs to the AWS CloudWatch
+   * service. The logs that this configures to stream are:
+   * - EC2 Instance UserData execution; this is the startup scripting that is run when the instance launches
+   *   for the first time.
+   * - Deadline Worker logs.
+   * - Deadline Launcher logs.
+   *
+   * @param worker The worker to configure. This can be an instance, auto scaling group, launch template, etc.
+   * @param id Identifier to disambiguate the resources that are created.
+   * @param logGroupProps Configuration for the log group in CloudWatch.
+   */
+  public configureCloudWatchLogStream(worker: IConfigurableWorker, id: string, logGroupProps?: LogGroupFactoryProps): void {
+    const logGroup = LogGroupFactory.createOrFetch(this, `${id}LogGroupWrapper`, `${id}`, logGroupProps);
+
+    logGroup.grantWrite(worker);
+
+    const cloudWatchConfigurationBuilder = new CloudWatchConfigBuilder(Duration.seconds(15));
+
+    if (worker.osType === OperatingSystemType.WINDOWS) {
+      cloudWatchConfigurationBuilder.addLogsCollectList(logGroup.logGroupName,
+        'UserdataExecution',
+        'C:\\ProgramData\\Amazon\\EC2-Windows\\Launch\\Log\\UserdataExecution.log');
+      cloudWatchConfigurationBuilder.addLogsCollectList(logGroup.logGroupName,
+        'WorkerLogs',
+        'C:\\ProgramData\\Thinkbox\\Deadline10\\logs\\deadlineslave*.log');
+      cloudWatchConfigurationBuilder.addLogsCollectList(logGroup.logGroupName,
+        'LauncherLogs',
+        'C:\\ProgramData\\Thinkbox\\Deadline10\\logs\\deadlinelauncher*.log');
+    } else {
+      cloudWatchConfigurationBuilder.addLogsCollectList(logGroup.logGroupName,
+        'cloud-init-output',
+        '/var/log/cloud-init-output.log');
+      cloudWatchConfigurationBuilder.addLogsCollectList(logGroup.logGroupName,
+        'WorkerLogs',
+        '/var/log/Thinkbox/Deadline10/deadlineslave*.log');
+      cloudWatchConfigurationBuilder.addLogsCollectList(logGroup.logGroupName,
+        'LauncherLogs',
+        '/var/log/Thinkbox/Deadline10/deadlinelauncher*.log');
+    }
+
+    new CloudWatchAgent(this, `${id}WorkerFleetLogsConfig`, {
+      cloudWatchConfig: cloudWatchConfigurationBuilder.generateCloudWatchConfiguration(),
+      host: worker,
+    });
+  }
+
+  /**
+   * This method can be used to set up the Deadline Worker application on an EC2 instance. From a practical
+   * perspective, this is executing the script found in aws-rfdk/lib/deadline/scripts/[bash,powershell]/configureWorker.[sh,ps1]
+   * to configure the Deadline Worker application.
+   *
+   * @param worker The worker to configure. This can be an instance, auto scaling group, launch template, etc.
+   * @param id Identifier to disambiguate the resources that are created.
+   * @param settings The Deadline Worker settings to apply.
+   */
+  public configureWorkerSettings(worker: IConfigurableWorker, id: string, settings?: WorkerSettings): void {
+    const configureWorkerScriptAsset = ScriptAsset.fromPathConvention(this, `${id}WorkerConfigurationScript`, {
+      osType: worker.osType,
+      baseName: 'configureWorker',
+      rootDir: path.join(
+        __dirname,
+        '..',
+        'scripts/',
+      ),
+    });
+
+    // Converting to lower case, as groups and pools are all stored in lower case in deadline.
+    const groups = settings?.groups?.map(val => val.toLowerCase()).join(',') ?? ''; // props.groups ? props.groups.map(val => val.toLowerCase()).join(',') : '';
+    const pools = settings?.pools?.map(val => val.toLowerCase()).join(',') ?? ''; // props.pools ? props.pools.map(val => val.toLowerCase()).join(',') : '';
+
+    configureWorkerScriptAsset.executeOn({
+      host: worker,
+      args: [
+        `'${groups}'`,
+        `'${pools}'`,
+        `'${settings?.region ?? ''}'`,
+        `'${Version.MINIMUM_SUPPORTED_DEADLINE_VERSION.toString()}'`,
+      ],
+    });
+  }
+}
--- a/packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
@@ -9,28 +9,25 @@ import {
   OperatingSystemType,
 } from '@aws-cdk/aws-ec2';
 import {
-  IGrantable,
-} from '@aws-cdk/aws-iam';
-import {
   Construct,
   Duration,
 } from '@aws-cdk/core';
 import {
   CloudWatchAgent,
   CloudWatchConfigBuilder,
-  IScriptHost,
   LogGroupFactory,
   LogGroupFactoryProps,
   ScriptAsset,
 } from '../../core';
-import { Version } from './version';
-
-/**
- * Interface for Deadline clients that can be configured via the ClientConfiguration
- * helper class.
- */
-export interface IConfigurableWorker extends IScriptHost, IGrantable {
-}
+import {
+  IHost,
+} from './host-ref';
+import {
+  IRenderQueue,
+} from './render-queue';
+import {
+  Version,
+} from './version';
 
 /**
  * The Deadline Worker settings that can be configured via the configureWorkerSettings method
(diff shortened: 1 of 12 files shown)


packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -0,0 +1,162 @@
...
+  LogGroupFactoryProps,
+  ScriptAsset,
+} from '../../core';
+import { Version } from './version';
+
+/**
+ * Interface for Deadline clients that can be configured via the ClientConfiguration
+ * helper class.
+ */
+export interface IConfigurableWorker extends IScriptHost, IGrantable {
```
jusiskin: nit: unless I'm missing something, I believe this is redundant since `IScriptHost` already extends `IGrantable`.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -0,0 +1,162 @@
...
+      rootDir: path.join(
+        __dirname,
+        '..',
+        'scripts/',
+      ),
+    });
+
+    // Converting to lower case, as groups and pools are all stored in lower case in deadline.
+    const groups = settings?.groups?.map(val => val.toLowerCase()).join(',') ?? ''; // props.groups ? props.groups.map(val => val.toLowerCase()).join(',') : '';
+    const pools = settings?.pools?.map(val => val.toLowerCase()).join(',') ?? ''; // props.pools ? props.pools.map(val => val.toLowerCase()).join(',') : '';
```
jusiskin: nit: these appear to be leftover comments from debugging

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -0,0 +1,162 @@
...
+ * - The instances configured by this construct will download and run scripts from your CDK bootstrap bucket when that instance
+ *   is launched. You must limit write access to your CDK bootstrap bucket to prevent an attacker from modifying the actions
+ *   performed by these scripts. We strongly recommend that you either enable Amazon S3 server access logging on your CDK
+ *   bootstrap bucket, or enable AWS CloudTrail on your account to assist in post-incident analysis of compromised production
+ *   environments.
+ */
+export class WorkerConfiguration extends Construct {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+  }
```
jusiskin: This construct has no logic in its constructor nor any member variables. All of the functionality is in instance methods that generate construct nodes under this construct sub-tree and the user has to call these in the correct order and provide disambiguating IDs.

I notice that in `WorkerInstanceFleet`, we have [this sequence of calls](https://github.com/ddneilson/aws-rfdk/blob/021ac8f2a0c4aa93eee24038c3fb7be91794fb05/packages/aws-rfdk/lib/deadline/lib/worker-fleet.ts#L444-L455):

```ts
    // Updating the user data with installation logs stream.
    workerConfig.configureCloudWatchLogStream(this.fleet, id, {
      logGroupPrefix: WorkerInstanceFleet.DEFAULT_LOG_GROUP_PREFIX,
      ...props.logGroupProps,
    });

    props.renderQueue.configureClientInstance({
      host: this.fleet,
    });

    // Updating the user data with deadline repository installation commands.
    workerConfig.configureWorkerSettings(this.fleet, id, props);
```

That configuration order should be preserved so errors configuring the Worker to connect to the render queue are streamed and the render queue connection is configured before trying to configure the worker, but I wonder if it makes makes sense to configure the Worker to use the render queue as part of this new `WorkerConfiguration` construct? It feels like an important missing piece of Worker configuration and we want to abstract knowledge of the correct calling order.

I suspect the decision to use methods was chosen because the `IRenderQueue` interface has two configuration methods that rely on knowing whether the Workers are instances or containers and the intention was to keep `WorkerConfiguration` agnostic to that. The current implementation is instance-specific anyways since Deadline configuration is done through user data which is not available in containers. You've already provided an example in the PR description of how to implement the `IHost` interface, so....

Allow me propose a slight change to bring this all together:

```ts
import { IHost } from './host-ref';
import { IRenderQueue } from './render-queue';

interface WorkerInstanceConfigurationProps extends WorkerSettings {
  readonly logGroup?: LogGroupFactoryProps;
  readonly renderQueue?: IRenderQueue;
  readonly workerHost: IHost;
}

export class WorkerInstanceConfiguration extends Construct {
  constructor(scope: Construct, id: string, props: WorkerInstanceConfigurationProps) {
    super(scope, id);

    if (props.logGroup) {
      this.configureCloudWatchLogStream(props.workerHost, props.logGroup);
    }
    if (props.renderQueue) {
      props.renderQueue.configureClientInstance({ host: props.workerHost });
    }
    this.configureWorkerSettings(props.workerHost, props);
  }
}
```

this makes the usage of this construct 1:1 with a Worker, but eliminates the need to specify child IDs, know the best-practice for the order of the method calls, makes the calling code cleaner. It goes from:

```ts
const worker: IConfigurableWorker = ...;
const renderQueue: IRenderQueue = ...;
const workerConfiguration = new WorkerConfiguration(this, 'WorkerConfiguration');
workerConfiguration.configureCloudWatchLogStream(worker, 'WorkerCloudWatch', { ... });
renderQueue.configureClientInstance({ host: props.worker });
workerConfiguration.configureWorkerSettings(worker, 'WorkerSettings', { ... });
```

to:

```ts
const workerHost: IHost = ...;
const renderQueue: IRenderQueue = ...;
const workerConfiguration = new WorkerInstanceConfiguration(this, 'WorkerConfiguration', {
  workerHost,
  renderQueue,
  logGroup: ...,
  groups: [...],
  pools: [...],
  region: 'myregion',
});
```

What do you think?

packages/aws-rfdk/lib/deadline/test/worker-configuration.test.ts
```diff
@@ -0,0 +1,503 @@
...
+        'Fn::Join': [
+          '',
+          [
+            '{\"logs\":{\"logs_collected\":{\"files\":{\"collect_list\":[{\"log_group_name\":\"',
+            {
+              'Fn::GetAtt': [
+                'ConfigWorkerLogGroupWrapperDC3AF2E7',
+                'LogGroupName',
+              ],
+            },
```
jusiskin: There is a technique to avoid baking these dynamically generated logical IDs into our tests. It involves the use of the [`stack.resolve(...)`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html#resolveobj). In this case, you could use:

```ts
const logGroupName = stack.resolve((config.node.findChild('WorkerLogGroup') as ILogGroup).logGroupName);
```

same feedback applies below.

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorkerHealthCheck.sh
```diff
@@ -0,0 +1,64 @@
...
+"$DEADLINE_COMMAND" -SetIniFileSetting LauncherHealthCheckPort $HEALTH_CHECK_PORT
+
+# Restart service, if it exists, else restart application
+if service --status-all | grep -q 'Deadline 10 Launcher'; then
+  service deadline10launcher restart
+else
+  DEADLINE_LAUNCHER="$DEADLINE_PATH/deadlinelauncher"
+  "$DEADLINE_LAUNCHER" -shutdownall
+  "$DEADLINE_LAUNCHER"
+fi
```
jusiskin: In the `WorkerInstanceFleet` case, we will now be restarting the launcher and Worker twice at startup. I understand it is necessary to ensure the changes take effect when running this script in isolation, but to avoid this extra restart, we could use OO-inheritance of `WorkerConfiguration`, like this:

```ts
class WorkerConfiguration {
  private readonly applyConfigScript: ScriptAsset;
  private readonly setConfigScript: ScriptAsset;

  constructor(...) {
    super(...);

    // Script to restart the launcher
    this.applyConfigScript = ScriptAsset.fromPathConvention(...);

    this.setConfiguration();
    this.applyConfiguration();
  }

  protected setConfiguration() {
    setConfigScript.executeOn(...);
  }

  protected applyConfiguration() {
    // Calls a separate ScriptAsset to restart the launcher
    applyConfigScript.executeOn(...);
  }
}
```

and for the worker instance fleet we override the `setConfiguration` method:

```ts
class WorkerInstanceFleetConfiguration extends WorkerConfiguration {
  private readonly configureHealthMonitor: ScriptAsset;

  constructor(...) {
    super(...);
    this.configureHealthMonitor = ScriptAsset.fromPathConvention(...);
  }

  protected setConfiguration() {
    super.setConfiguration();
    this.configureHealthMonitor.executeOn(...);
  }
}
```

This assumes my other proposed change of applying the configuration in the constructor. It's a bit of rework, but do you think it's worth it?

jusiskin (changes requested):
Fantastic job factoring out some re-usable logic here without changing any public APIs. I just have  some suggestions to refine the new APIs we are introducing here - let me know what you think.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -0,0 +1,162 @@
...
+      baseName: 'configureWorker',
+      rootDir: path.join(
+        __dirname,
+        '..',
+        'scripts/',
+      ),
+    });
+
+    // Converting to lower case, as groups and pools are all stored in lower case in deadline.
+    const groups = settings?.groups?.map(val => val.toLowerCase()).join(',') ?? ''; // props.groups ? props.groups.map(val => val.toLowerCase()).join(',') : '';
```
yashda: Comments can be removed.

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorkerHealthCheck.sh
```diff
@@ -0,0 +1,64 @@
...
+    exit 1
+fi
+
+# enabling the health check port
+"$DEADLINE_COMMAND" -SetIniFileSetting ResourceTrackerVersion V2
+# health check port
+"$DEADLINE_COMMAND" -SetIniFileSetting LauncherHealthCheckPort $HEALTH_CHECK_PORT
+
+# Restart service, if it exists, else restart application
+if service --status-all | grep -q 'Deadline 10 Launcher'; then
```
yashda: Pulling this out to a separate script will restart deadline launcher and worker twice. Should we also consider moving the restart command to a separate script and run it once after all the configuration scripts?
If we plan for it, we can also separate out the version comparison script and runt it once.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * Security Considerations
+ * ------------------------
+ * - The instances configured by this construct will download and run scripts from your CDK bootstrap bucket when that instance
+ *   is launched. You must limit write access to your CDK bootstrap bucket to prevent an attacker from modifying the actions
+ *   performed by these scripts. We strongly recommend that you either enable Amazon S3 server access logging on your CDK
+ *   bootstrap bucket, or enable AWS CloudTrail on your account to assist in post-incident analysis of compromised production
+ *   environments.
+ */
+export class WorkerConfiguration extends Construct {
```
yashda: Should we rename this construct to `ClientConfiguration`? The construct looks generic enough to configure any deadline client eg. Monitor.

jusiskin (dismissed):
Approved! Really great job here

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
ddneilson: The generic client configuration is the render queue's client connect methods. The worker configuration script has some very worker-specific stuff in it like getting the Launcher to start a Worker.

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorkerHealthCheck.sh
ddneilson: Heh. That makes three of us that noticed the restart -- you, me, and Josh.

It looks like the restart is pretty quick, so there isn't much of an issue with the additional restart. In this case, though, it's probably safe to just remove the restart... we know this happens before the generic worker configuration, and we know that the worker configuration script issues a restart.

yashda (dismissed):
Looks good! Thanks Daniel for the changes.

jusiskin (dismissed):

jusiskin (approved):

yashda (approved):

PR #210, opened 2020-11-04
feat(integ): add ability to run integration tests using ECR images

jericht:
Adds options to override Deadline Docker images used in integration tests. This allows us to run integration tests on Docker images that are already built and in an ECR repository.

### Testing
- Built a Docker image for Deadline RCS and pushed it to an ECR repository
- Defined the `RFDK_DOCKER_IMAGE_OVERRIDES` environment variable with the appropriate JSON
- Ran the RFDK integration tests and ensured:
    1. The Docker images specified in the overrides were used over the ones that were built from staged recipes
    1. The integration tests succeeded

----

### DIFF
--- a/integ/lib/ThinkboxDockerImageOverrides.ts
+++ b/integ/lib/ThinkboxDockerImageOverrides.ts
@@ -1,0 +1,134 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Repository } from '@aws-cdk/aws-ecr';
+import { ContainerImage } from '@aws-cdk/aws-ecs';
+import { Construct } from '@aws-cdk/core';
+import {
+  RenderQueueImages,
+  ThinkboxManagedDeadlineDockerRecipes as Recipes,
+  UsageBasedLicensingImages,
+} from 'aws-rfdk/deadline';
+
+type DeadlineDockerImageOverrides = {[key in Recipes]?: string};
+
+/**
+ * Interface for Docker image overrides stored in an ECR repository.
+ */
+interface IECRImageOverrides {
+  /**
+   * The ARN of the ECR repository.
+   */
+  readonly repositoryArn: string;
+  /**
+   * A mapping of Deadline Docker recipe names (see ThinkboxManagedDeadlineDockerRecipes in aws-rfdk) to tags of the override Docker images
+   */
+  readonly imageOverrides: DeadlineDockerImageOverrides;
+}
+
+/**
+ * Implementation of {@link IECRImageOverrides}.
+ */
+class ECRImageOverrides implements IECRImageOverrides {
+  /**
+   * Creates an {@link ECRImageOverrides} from a JSON string.
+   * @param json The JSON string.
+   */
+  public static fromJSON(json: string): ECRImageOverrides {
+    const obj = JSON.parse(json);
+
+    // Validate the input JSON
+    const errors = [];
+    [
+      'repositoryArn',
+      'imageOverrides',
+    ].forEach(prop => {
+      if (!(prop in obj)) {
+        errors.push(`Property ${prop} was expected but not found in ${obj}`);
+      }
+    });
+    if (!(obj.imageOverrides instanceof Object)) {
+      errors.push(`Expected Object for imageOverrides but found ${typeof(obj.imageOverrides)}`);
+    }
+    Object.keys(obj.imageOverrides).forEach(key => {
+      if (!Object.values(Recipes).includes(key as Recipes)) {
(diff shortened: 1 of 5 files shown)


integ/lib/render-struct.ts
```diff
@@ -30,6 +33,12 @@ export class RenderStruct extends Construct {
...
     const vpc = Vpc.fromLookup(this, 'Vpc', { tags: { StackName: infrastructureStackName }}) as Vpc;
 
+    // Retrieve Docker image overrides, if available
+    let dockerImageOverrides: (ThinkboxDockerImageOverrides | undefined) = undefined;
+    if (process.env[DOCKER_IMAGE_OVERRIDES_ENV_VAR] !== undefined) {
+      dockerImageOverrides = ThinkboxDockerImageOverrides.fromJSON(this, 'ThinkboxDockerImageOverrides', process.env[DOCKER_IMAGE_OVERRIDES_ENV_VAR]!.toString());
+    }
+
     // Stage docker recipes, which include image used for the render queue instance
     const recipes = new ThinkboxDockerRecipes(this, 'DockerRecipes', {
```
yashda: Should we initialize this if overrides are provided.
Can we completely remove the dependency of calling the staging script here if we are providing the ECR images explicitly?

integ/test-config.sh
```diff
@@ -16,6 +16,20 @@ export DEADLINE_VERSION
 #   - If set here, the version found in `manifest.json` at this path will override any value supplied for DEADLINE_VERSION
 export DEADLINE_STAGING_PATH
 
+# Override for Deadline Docker images that will be used over the images that can be built via the staged Deadline Docker recipes
+#   - If set, it must be a JSON string like:
+#     {
```
yashda, line 21: If this schema is changed to following, imageOverrides can be easily typescasted to the typescript interface and the component tags can be queried easily. Eg.:
```
const inputJson = {
  "repositoryArn": 'sampleARN',
  "imageOverrides": {
      'rcs': '10.1.9.2',
      'license-forwarder': '10.1.10.6',
    }
};
enum Images {
  RCS = 'rcs',

  UBL = 'license-forwarder',
}
interface ECRImageOverrides {
    repositoryArn: string,
    imageOverrides: {
        [key in Images]: string
    }
}

let images: ECRImageOverrides = inputJson;

console.log('rcs tag: ' + images.imageOverrides[Images.RCS]);
console.log('license-forwarder tag: ' + images.imageOverrides[Images.UBL]);
```

yashda (changes requested):
Nice work Jericho. However, I think we can simplify the class `ECRImageOverrides` based on the comments provided.

integ/lib/render-struct.ts
jericht: Good call out. I looked into if we could completely remove calling the staging script, but our integration tests currently rely on the Deadline installers to be staged. (see [code](https://github.com/aws/aws-rfdk/blob/06e7863994f7e05d89a897e5804cc543e5fd29db/integ/lib/testing-tier.ts#L146)). It looks like we're installing Deadline Client with the staged installers on the bastion instances to connect to the components we're testing. We could change this to pull a Docker image from an ECR repository instead, but that feels a bit out of scope for this PR IMO. What do you think?

For now, I've left this as is since the staging directory is being used in a few places here.

integ/test-config.sh
jericht: Good idea, this makes it much cleaner! I've changed the interface to this.

integ/lib/render-struct.ts
yashda: If the staging folder has other dependencies then I think its fine to leave it like this for now. We can handle it later. 
IMO, instead of using the docker image, we can use aws cli to copy the installer from S3 bucket directly.

integ/lib/ThinkboxDockerImageOverrides.ts
```diff
@@ -0,0 +1,137 @@
...
+      if (!Object.values(Recipes).includes(key as Recipes)) {
+        errors.push(`Key ${key} in imageOverrides is invalid. Must be in: [${Object.values(Recipes)}].`);
+      }
+    });
+
+    if (errors.length > 0) {
+      throw new Error(`Invalid JSON for ECRImageOverrides: ${errors.join('\n')}`);
+    }
+
+    return new ECRImageOverrides({
```
yashda: nit: Can we simplify it to:
`return obj as ECRImageOverrides;`

yashda (dismissed):
The code changes looks good!

integ/lib/ThinkboxDockerImageOverrides.ts
jericht: Sure, updated

yashda (approved):
All the changes looks good. Good work Jericho!

jusiskin (approved):
In offline discussions, confirmed that testing without the environment variable works as it did before.

Great job! approved.

PR #214, opened 2020-11-05
chore: updates to all-in-aws ts example

horsmand:
* Made UBL optional
* Updated stage script usage
* Updated Deadline version and added default worker AMI to match

Tested by running `cdk deploy` on the app, it can deploy without needing to modify anything at all now.

----

### DIFF
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/README.md
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/README.md
@@ -27,6 +27,12 @@ These instructions assume that your working directory is `examples/deadline/All-
       // ...
       };
     ```
+
+    ---
+
+    **Note:** The next three steps are for setting up usage based licensing and are optional. You may skip these if you do not need to use licenses for rendering.
+
+    ---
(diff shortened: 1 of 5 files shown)


jericht (dismissed):
LGTM

ddneilson (dismissed):

examples/deadline/All-In-AWS-Infrastructure-Basic/ts/README.md
```diff
@@ -32,7 +32,13 @@ These instructions assume that your working directory is `examples/deadline/All-
     ```
     aws secretsmanager create-secret --name <name> --secret-binary fileb://<path-to-zip-file>
     ```
-5.  The output from the previous step will contain the secret's ARN. Change the value of the `ublCertificatesSecretArn` variable in `bin/config.ts` to the secret's ARN:
+
+    ---
+
+    **Note:** The next two steps are for setting up usage based licensing and are optional. You may skip these if you do not need to use licenses for rendering.
```
jericht: I think I missed this last time, but we could also skip the previous step as well right? We don't need to create a secret with our UBL certs since we aren't using them.

jericht (changes requested):
Just the one small comment, otherwise it LGTM.

examples/deadline/All-In-AWS-Infrastructure-Basic/ts/README.md
horsmand: Oh, whoops, I didn't notice that. Yes I'll update this, nice catch.

jericht (approved):
LGTM

ddneilson (approved):

PR #216, opened 2020-11-05
chore: Adds supported Deadline versions to CHANGELOG

ddneilson:
This is modifying the bump script to add the officially supported Deadline versions to the release changelog.

New section to the CHANGELOG looks like:
```
### Officially Supported Deadline Versions

* [10.1.9.2 to 10.1.10.6](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/release-notes.html)
```

----

### DIFF
--- a/bump.sh
+++ b/bump.sh
@@ -55,8 +55,17 @@ if [[ $version == "patch" ]]; then
   sed -i "s|$version_header|\1|" ./CHANGELOG.md
 fi
 
-# Add a section to the changelog that state the version of CDK being used
 version_header="# \[$new_version](.*) (.*)"
+
+# Add a section to the changelog that states the supported Deadline versions
+DEADLINE_RELEASE_NOTE_URL="https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/release-notes.html"
+DEADLINE_SUPPORTED_VERSIONS=$(node ./scripts/getSupportedDeadlineVersions.ts)
+MIN_DEADLINE_VERSION=$(echo "$DEADLINE_SUPPORTED_VERSIONS" | grep 'Min' | cut -f 2 -d ' ')
+MAX_DEADLINE_VERSION=$(echo "$DEADLINE_SUPPORTED_VERSIONS" | grep 'Max' | cut -f 2 -d ' ')
+deadline_version_section="\n\n\n### Officially Supported Deadline Versions\n\n* [${MIN_DEADLINE_VERSION} to ${MAX_DEADLINE_VERSION}](${DEADLINE_RELEASE_NOTE_URL})"
+sed -i "s|\($version_header\)|\1$deadline_version_section|" ./CHANGELOG.md
+
+# Add a section to the changelog that state the version of CDK being used
 cdk_version=$(node -p "require('./package.json').devDependencies['aws-cdk']")
 cdk_version_section="\n\n\n### Supported CDK Version\n\n* [$cdk_version](https://github.com/aws/aws-cdk/releases/tag/v$cdk_version)"
 sed -i "s|\($version_header\)|\1$cdk_version_section|" ./CHANGELOG.md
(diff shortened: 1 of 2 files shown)


scripts/getLatestDeadlineVersion.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const provider=new mod.VersionProvider();
+
+provider.getVersionUris({ product: mod.Product.deadline, platform: mod.Platform.linux })
+  .then(result => {
+    const version = result.get(mod.Platform.linux);
+    console.log(`${version.MajorVersion}.${version.MinorVersion}.${version.ReleaseVersion}.${version.PatchVersion}`);
+  })
+  .catch(error => {
+    console.error(`ERROR - ${error.message}`);
```
jusiskin: this will not set a non-zero exit code for the process since we are catching this error. Should we call [`process.exit(code)`](https://nodejs.org/api/process.html#process_process_exit_code) here so the caller has an easy way of identifying an error?

scripts/getLatestDeadlineVersion.ts
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+
+const mod = require('../packages/aws-rfdk/lib/core/lambdas/nodejs/lib/version-provider');
```
jusiskin: was there something that prevented us from being able to use an import instead?

```ts
import { VersionProvider } from '../packages/aws-rfdk/lib/core/lambdas/nodejs/lib/version-provider';

File: bump.sh [hunk 3]
```diff
@@ -55,8 +55,21 @@ if [[ $version == "patch" ]]; then
   sed -i "s|$version_header|\1|" ./CHANGELOG.md
 fi
 
-# Add a section to the changelog that state the version of CDK being used
 version_header="# \[$new_version](.*) (.*)"
+
+# Add a section to the changelog that states the supported Deadline versions
+DEADLINE_RELEASE_NOTE_URL="https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/release-notes.html"
+MIN_DEADLINE_VERSION=$(grep 'MINIMUM_SUPPORTED_DEADLINE_VERSION_STRING =' $(find ./packages/aws-rfdk/lib/ -name '*.ts') | tr ' ' '\n' | grep "'" | tr -d "';")
+MAX_DEADLINE_VERSION=$(node ./scripts/getLatestDeadlineVersion.ts)
```
jusiskin: Could we do something similar where we run a node script to obtain the minimum supported version instead of grepping?

jusiskin (changes requested):
great addition for RFDK users! just a few minor suggestions here

scripts/getLatestDeadlineVersion.ts
ddneilson: Yeah. I didn't think it's worth digging into, but I was getting node errors when trying that syntax.

bump.sh
ddneilson: Sure, I'll switch things up.

jusiskin (dismissed):
Looks great. Approved.

bump.sh
```diff
@@ -55,8 +55,17 @@ if [[ $version == "patch" ]]; then
   sed -i "s|$version_header|\1|" ./CHANGELOG.md
 fi
 
-# Add a section to the changelog that state the version of CDK being used
 version_header="# \[$new_version](.*) (.*)"
+
+# Add a section to the changelog that states the supported Deadline versions
+DEADLINE_RELEASE_NOTE_URL="https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/release-notes.html"
+DEADLINE_SUPPORTED_VERSIONS=$(node ./scripts/getSupportedDeadlineVersions.ts)
```
horsmand, line 62: I'm surprised this works without using `ts-node` since it's pointing to a `.ts` file.

scripts/getSupportedDeadlineVersions.ts
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env node
```
horsmand, line 1: Should this file have our license header?

jusiskin (approved):

horsmand (approved):

PR #224, opened 2020-11-09
chore: pull amazonlinux Docker image from ECR

jusiskin:
## Problem

Beginning November 2 2020, DockerHub began [rate-limiting unauthenticated image pulls](https://docs.docker.com/docker-hub/download-rate-limit/).

Part of the RFDK release automation involves building and publishing lambda layers for use in RFDK constructs. This step uses Docker and currently pulls `amazonlinux` Docker image from DockerHub. This does not authenticate with DockerHub before pulling images. As a result, when we run the release automation, we see the following error:

```
Step 1/4 : FROM amazonlinux
time="2020-11-09T19:54:55.633791062Z" level=error msg="Not continuing with pull after error: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

## Solution

Created a new script in `scripts/pull_amazonlinux_from_ecr.sh` that pulls Amazon Linux Docker images directly from the official AWS ECR repositories [using commands derived from these instructions](https://docs.aws.amazon.com/AmazonECR/latest/userguide/amazon_linux_container_image.html) and then tags them as-if they were pulled from DockerHub.

Modified `scripts/prep_release.sh`, which is the entry-point of our release bump automation to call `scripts/pull_amazonlinux_from_ecr.sh` before building and publishing the lambda layers.

## Testing

The `scripts/pull_amazonlinux_from_ecr.sh` was written to detect the EC2 region for the ECR pull from the instance meta-data. This will work from our pipeline. It has fallback logic to obtain the region from the AWS CLI for developer use. Tested using this code path and ran:

```sh
bash scripts/prep_release.sh
```

from my branch. Confirmed that it successfully pulled and tagged images from ECR succeeded and the lambda layers were published to my AWS account.

Launched an on-demand EC2 instance and manually ran the commands in the code-path that queries the EC2 instance meta-data. Confirmed the meta-data commands and error-handling works as expected.

----

### DIFF
--- a/scripts/prep_release.sh
+++ b/scripts/prep_release.sh
@@ -16,11 +16,36 @@ SCRIPT_DIR=$(dirname $0)
 SCRIPT_DIR=$(readlink -f $SCRIPT_DIR)
 ROOT_DIR=$(readlink -f "${SCRIPT_DIR}/..")
 
+# Determine the ECR region we want to use
+set +e
+# First attempt to use the EC2 instance meta-data to determine the region
+ECR_REGION="$(curl -fs http://169.254.169.254/latest/meta-data/placement/region)"
+if [ "$?" -ne 0 ]; then
+    # Not running on EC2, so get region from AWS CLI configuration
+    ECR_REGION="$(aws configure get region)"
+    if [ "$?" -ne 0 || -z "${ECR_REGION}" ]; then
+        echo "ERROR: Could not determine region"
+        exit 1
+    fi
+fi
+set -e
+
 # Ensure that the docker daemon is running in the background
 # These commands are currently being done here instead of in buildspec due to the following issue: https://github.com/awslabs/aws-delivlib/issues/448
 # These commands come from the Codebuild sample for running docker on a custom image: https://docs.aws.amazon.com/codebuild/latest/userguide/sample-docker-custom-image.html#sample-docker-custom-image-files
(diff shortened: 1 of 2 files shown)


kozlove-aws (approved):

scripts/prep_release.sh
```diff
@@ -16,12 +16,37 @@ SCRIPT_DIR=$(dirname $0)
...
 timeout 15 sh -c "until docker info; do echo .; sleep 1; done"
 
+# Pull amazonlinux image locally from ECR. This is done to avoid pulling from DockerHub both for egress and rate
+# limiting.
+PULL_AL_FROM_ECR_ARGS=(
+    # Region of ECR repo
+    "${ECR_REGION}"
+    # Image versions to pull
+    "latest" # required for building and publishing lambda layers
+    "2"      # required for building Deadline docker images for running integration tests
```
ddneilson, line 46: The integration tests don't use the latest version of the AL2 image? Any particular reason?

jusiskin: The Deadline Docker recipes pin to use the major version 2 of the `amazonlinux` image:

```docker
FROM amazonlinux:2 AS builder
```

Technically, the integration tests don't yet run in the release automation, but #205 is looking to change that. This is a proactive step to have that image pulled locally from ECR so that building staged recipes will not be throttled when that is integrated.

ddneilson (approved):

PR #230, opened 2020-11-12
fix(deadline): lock down DocDC engine to version 3.6.0

### LINKED ISSUE #229: DocDB version is not locked to v3.6 for Deadline
Officially, Deadline only supports version 3.6 of the MongoDB API. Amazon DocumentDB has just GA'd their 4.0 engine ( https://aws.amazon.com/blogs/database/introducing-amazon-documentdb-with-mongodb-compatibility-4-0/ ).

According to the [CloudFormation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-docdb-dbcluster.html#cfn-docdb-dbcluster-engineversion), the DocumentDB cluster defaults to the latest engine version when being created -- this is now 4.0.0. In a quick test, it looks like the default is still to create a 3.6.0 engine, but I think we can safely assume that will change.

### Other

Fix will be to simply specify the engine version for Document DB as 3.6.0.

--- 

This is :bug: Bug Report

ddneilson:
Fixes: https://github.com/aws/aws-rfdk/issues/229

## Testing

Did a quick deploy of the Deadline Repository & RCS. Verified it deployed successfully, and the DocDB engine version was correct. Also changed the engine version to 4.0.0 and did the same to verify that it changed there too.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
@@ -82,6 +82,7 @@ export interface MongoDbInstanceConnectionOptions {
 export abstract class DatabaseConnection {
   /**
    * Creates a DatabaseConnection which allows Deadline to connect to Amazon DocumentDB.
+   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6.
    *
    * Resources Deployed
    * ------------------------
@@ -93,6 +94,7 @@ export abstract class DatabaseConnection {
 
   /**
    * Creates a DatabaseConnection which allows Deadline to connect to MongoDB.
+   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6.
    *
    * Resources Deployed
    * ------------------------
@@ -273,10 +275,7 @@ class DocDBDatabaseConnection extends DatabaseConnection {
     // checking the value of the engineVersion property of that object.
     if (this.props.database.node.defaultChild) {
       const cluster = this.props.database.node.defaultChild! as CfnDBCluster;
-      if (cluster.engineVersion === '3.6.0') {
-        return true;
-      }
-      return false;
+      return cluster.engineVersion?.startsWith('3.6') ?? false;
     }
 
     return true; // No information, assume it's compatible.
--- a/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
@@ -44,7 +44,8 @@ import {
 export interface DocDBConnectionOptions {
 
   /**
-   * The Document DB Cluster this connection is for
+   * The Document DB Cluster this connection is for.
+   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6.
    */
   readonly database: IDatabaseCluster;
 
@@ -61,6 +62,7 @@ export interface DocDBConnectionOptions {
 export interface MongoDbInstanceConnectionOptions {
   /**
    * The MongoDB database to connect to.
+   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6.
    */
   readonly database: IMongoDb;
 
--- a/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
@@ -5,6 +5,7 @@
 
 import * as path from 'path';
 import {
+  CfnDBCluster,
   CfnDBInstance,
   IDatabaseCluster,
 } from '@aws-cdk/aws-docdb';
@@ -22,6 +23,7 @@ import {
   ISecret,
 } from '@aws-cdk/aws-secretsmanager';
 import {
+  Annotations,
   IConstruct,
   Stack,
 } from '@aws-cdk/core';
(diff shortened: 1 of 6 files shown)


packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -490,6 +490,7 @@ export class Repository extends Construct implements IRepository {
       const instances = props.documentDbInstanceCount ?? Repository.DEFAULT_NUM_DOCDB_INSTANCES;
       const dbCluster = new DatabaseCluster(this, 'DocumentDatabase', {
         masterUser: {username: 'DocDBUser'},
+        engineVersion: '3.6.0',
```
jusiskin, line 494: Two things:

1.  This is good for the case where `props.database` is not supplied, but is it possible to add validation that the engine is correct for the code path where a caller specifies their own cluster and fail-fast?
1.  Can we add a test for this new behavior?

jusiskin (changes requested):
Can we add some guidance to the relevant TSDoc entries? I'm thinking good places might be:
*   the `Repository` construct
*   the `DatabaseConnection.forDocDB` method.

I was also hopeful we could perform a pre-flight run-time validation, but unfortunately it looks like the engine version is not available as a [`DatabaseCluster` attribute](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-docdb.DatabaseCluster.html#properties).

packages/aws-rfdk/lib/deadline/lib/repository.ts
ddneilson: It's a good idea. I'll see what I can cobble together.

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -256,6 +262,25 @@ class DocDBDatabaseConnection extends DatabaseConnection {
...
+  protected isCompatibleDocDBVersion(): boolean {
+    // The defaultChild of a DocDB DatabaseCluster is a CfnDBCluster, but we only have this
+    // child if the customer didn't import from attributes. We can check the DB version by
+    // checking the value of the engineVersion property of that object.
+    if (this.props.database.node.defaultChild) {
+      const cluster = this.props.database.node.defaultChild! as CfnDBCluster;
+      if (cluster.engineVersion === '3.6.0') {
+        return true;
+      }
+      return false;
```
jusiskin: Looks good. I'm just thinking in the event that DocDB issues a patch version for 3.6. Just in case, should we relax the final version component here? So:

```ts
return cluster.engineVersion.startsWith('3.6.');
```

It looks like RDS has patch engine versions for other DB types:

```sh
$ aws docdb describe-db-engine-versions --engine mariadb --engine-version 10.0.17
{
    "DBEngineVersions": [
        {
            "Engine": "mariadb",
            "EngineVersion": "10.0.17",
...
```

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -288,6 +288,7 @@ export interface RepositoryProps {
 
   /**
    * Specify the database where the deadline schema needs to be initialized.
+   * Note that Deadline supports only databases that are compatible with MongoDB 3.6.
```
jusiskin, line 291: This looks great. I had also suggested adding something to the TSDoc string of `DatabaseConnection.forDocDB` too.  Perhaps even `DocDBConnectionOptions.database`. What do you think?

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -163,6 +165,10 @@ class DocDBDatabaseConnection extends DatabaseConnection {
   constructor(private readonly props: DocDBConnectionOptions) {
     super();
 
+    if (!this.isCompatibleDocDBVersion()) {
+      Annotations.of(props.database).addError('engineVersion must be 3.6.0 to be compatible with Deadline');
```
jusiskin, line 173: same comment as below. Should we relax this message to `3.6.x`

jusiskin (changes requested):

packages/aws-rfdk/lib/deadline/lib/repository.ts
ddneilson: Ah, right. Sure. I can add something there...

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
ddneilson: No harm in doing that, though the DocDB service team will never have any 3.6 value other than 3.6.0 and similarly, no 4.0 other than 4.0.0. They specifically don't do patch versions.

jusiskin (dismissed):
Looks good. Approved!

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
horsmand: This still says 3.6.0 but our other messaging and the version check all use 3.6

packages/aws-rfdk/lib/deadline/lib/repository.ts
horsmand: The comment for `DocDBConnectionOptions.database` wasn't updated, do we need that too?

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
ddneilson: Ah, I discussed this with Josh offline. DocumentDB doesn't do patch-level versions. It will always be 3.6.0

packages/aws-rfdk/lib/deadline/lib/repository.ts
ddneilson: Added those too in latest rev.

horsmand (approved):

jusiskin (approved):
Great job! Thanks @ddneilson

PR #235, opened 2020-11-14
chore(lambdas): move lambdas

### LINKED ISSUE #146: Move Lambdas Directory
I propose that we move the `lambdas` folder up a directory level, out of `core`, to be at the same level as `core` and `deadline`.

### Use Case

Right now we have a nice `lib` directory inside `core/lambdas/nodejs` that has some very useful classes, but we wouldn't be able to use them in a Lambda if the Lambda itself was put under `deadline/lambdas/nodejs`.

### Proposed Solution

Having `lambdas` at the same directory level as `core` and `deadline`, instead of individual `lambdas` directories in those other directories would allow us to more share our Lambdas library with Lambdas that are meant to be used by classes in both `core` and `deadline`.

---

horsmand:
Fixes #146

Deployed the "all in aws basic" example app to test the lambdas all still work.

Also includes a correction in both the ts and python readme's for `examples/deadline/All-In-AWS-Infrastructure-Basic/` as well as fixing the spot where `packages/aws-rfdk/lib/lambdas/nodejs/version-provider/test/handler.test.ts` was writing to the console during a run of the unit tests.

----


ddneilson (dismissed):
Thanks for the cleanup, David.

packages/aws-rfdk/lib/lambdas/nodejs/version-provider/test/handler.test.ts
```diff
@@ -3,14 +3,21 @@
...
+  const consoleLog = console.log;
   beforeEach(() => {
     versionProviderResource = new VersionProviderResource();
+    // tslint:disable-next-line: no-empty
+    console.log = () => {};
+  });
+
+  afterEach(() => {
+    console.log = consoleLog;
   });
```
jusiskin: nit: jest includes a nice utility for monkey-patching existing modules/APIs. Not critical, but it would avoid the extra `consoleLog` variable and potentially the `no-console` eslint disabling since this would get delegated to jest.

```ts
beforeEach(() => {
  // ...
  jest.spyOn(console, 'log').mockReturnValue(undefined);
  // ...
});

afterEach(() => {
  // ...
  jest.clearAllMocks();
  // ...
});
```

What do you think?

jusiskin (changes requested):
Looks good. One optional suggestion, but I'll defer to you on whether it is worth it.

packages/aws-rfdk/lib/lambdas/nodejs/version-provider/test/handler.test.ts
ddneilson: iirc, I tried this in the tests for the mongodb handler (ex: here -- https://github.com/aws/aws-rfdk/blob/db0e5b93df76bc1257124e9033e01c1de6964299/packages/aws-rfdk/lib/core/lambdas/nodejs/mongodb/test/handler.test.ts#L363 ), but there were some issues with it; the details of which I do not recall. The only way that I could get it to work was the console log trick that David is doing here.

horsmand: I had copied this from the file Daniel mentioned, but I'm always down for learning new things, so I tried out the spies like Josh suggested and they worked for me. I tried them out in the Mongo tests too, and I was able to get them to work there, so I updated them as well.

packages/aws-rfdk/lib/deadline/test/version-query.test.ts
```diff
@@ -57,44 +55,6 @@ test('VersionQuery constructor full version', () => {
     ],
   }));
   expectCDK(stack).to(haveResourceLike('AWS::Lambda::Function', {
-    Code: {
-      S3Bucket: {
-        Ref: VERSION_QUERY_ASSET.Bucket,
-      },
-      S3Key: {
```
horsmand, line 64: This block has been causing problems due to the S3 bucket name and object key hashes always changing. I know Eugene has a PR open already where he was going to remove this section, so when my tests failed due to the hash changing I decided to just delete the `Code` section rather than fix the hashes.

ddneilson (approved):

packages/aws-rfdk/lib/deadline/test/version-query.test.ts
jusiskin: Can we obtain these values using `stack.resolve(asset.s3ObjectKey)` and `stack.resolve(asset.bucket.buckwetName)`?

horsmand: Unless I'm missing something, if we are pulling values from the stack and then using those values in the expectation, we're essentially saying, "Make sure that the code in the stack is the code that we found in the stack."

I don't think this would be very useful in ensuring that we are using the correct code here, it would be more of a test that CDK is working properly. Our other tests for Lambda code don't have this kind of test in them either.

jusiskin: This is fair - I don't see us losing much by removing this.

jusiskin (approved):

Issue #191, opened 2020-10-27
Worker nodes and UserData


gdday:
My team would like to be able to mount storage (such as FSx for Lustre) on the Worker nodes before the worker nodes connect to the Deadline server. This storage contains the data needed to be able to run jobs.

That isn't currently possible to do using the CDK & RFDK. I've added code to mount the storage to the UserData, but I only have the ability to add the code *after* the UserData that the RFDK adds to the Auto-Scaling Group. I can do it manually in the EC2 console, but not in the CDK application itself.

I would like the ability to either add my UserData before the RFDK's UserData, or I would like to be able to get the RFDK's UserData so that I can do this myself, and then I'm responsible for setting the UserData on the ASG.

I wasn't sure whether this was a bug, or a feature request. It's a little of both - ideally workers shouldn't connect to the server until they're ready to run jobs.

### Environment

  - **CDK CLI Version      :**
  - **CDK Framework Version:**
  - **RFDK Version:**
  - **Deadline Version:**
  - **Node.js Version:** 
  - **OS               :**
  - **Language (Version):** 

### Other

--- 

This is :bug: Bug Report

PR #239, opened 2020-11-17
feat(deadline): add custom user data commands to Worker instance startup

kozlove-aws:
## Problem
Users do not have possibility to run user data scripts before worker configured and run by RFDK configuration script.
Was created issue with request to add this functional: [Worker nodes and UserData](https://github.com/aws/aws-rfdk/issues/191)

## Solution
Was chosen solution that allows user to create class with methods that add user data scripts in all stages of worker configuration.
Was added interface `IInstanceUserDataProvider` and class `InstanceUserDataProvider` with three methods.
Every method can be overridden by user and has possibility to modify worker user data.
Were added examples for this functionality for Python and Typescript.

## Testing
Was added integration tests that cover 100% of new functionality.
Were deployed Python and Typescript examples and verified that all expected code was added to userdata and example script was copied to worker instance and run.

Was removed validation S3Bucket in version query test because these values are changed periodically.

----

### DIFF
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
@@ -17,15 +17,20 @@ from aws_cdk.aws_ec2 import (
     IVpc,
     Port
 )
+from aws_cdk.aws_s3_assets import (
+  Asset
+)
 
 from aws_rfdk import (
     HealthMonitor,
 )
 from aws_rfdk.deadline import (
+    InstanceUserDataProvider,
     IRenderQueue,
     WorkerInstanceFleet,
 )
 
+import os
 
 @dataclass
 class ComputeTierProps(StackProps):
@@ -43,6 +48,30 @@ class ComputeTierProps(StackProps):
     # The bastion host to  allow connection to Worker nodes.
     bastion: Optional[BastionHostLinux] = None
 
+class UserDataProvider(InstanceUserDataProvider):
+    def __init__(self, scope: Construct, stack_id: str):
+        super().__init__(scope, stack_id)
+        self.test_script=Asset(scope, "SampleAsset",
+            path=os.path.join(os.getcwd(), "..", "scripts", "configure_worker.sh")
+        )
+
+    def pre_cloud_watch_agent(self, host) -> None:
+        host.user_data.add_commands("echo preCloudWatchAgent")
+
+    def pre_render_queue_configuration(self, host) -> None:
+        host.user_data.add_commands("echo preRenderQueueConfiguration")
+
+    def pre_worker_configuration(self, host) -> None:
+        host.user_data.add_commands("echo preWorkerConfiguration")
+
+    def post_worker_launch(self, host) -> None:
+        host.user_data.add_commands("echo postWorkerLaunch")
+        self.test_script.grant_read(host)
+        local_path = host.user_data.add_s3_download_command(
+            bucket=self.test_script.bucket,
+            bucket_key=self.test_script.s3_object_key
+        )
+        host.user_data.add_execute_file_command(file_path=local_path)
 
 class ComputeTier(Stack):
     """
@@ -78,6 +107,7 @@ class ComputeTier(Stack):
             worker_machine_image=props.worker_machine_image,
             health_monitor=self.health_monitor,
             key_name=props.key_pair_name,
+            user_data_provider=UserDataProvider(self, 'UserDataProvider')
         )
 
         if props.bastion:
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/scripts/configure_worker.sh
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/scripts/configure_worker.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+# This script is only here to illustrate how to add custom user data when using our WorkerInstanceFleet and WorkerInstanceConfiguration constructs
+mkdir test_dir 
--- a/packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
@@ -30,6 +30,67 @@ import {
 } from './version';
 
 /**
+ * Provider for adding user data scripts
+ * Methods of this interface will be invoked in WorkerInstanceConfiguration
+ * on different stages of worker configuration
+ */
+export interface IInstanceUserDataProvider {
+  /**
+   * Method that is invoked before configuring the Cloud Watch Agent.
+   */
+  preCloudWatchAgent(host: IHost): void;
+
+  /**
+   * Method that is invoked before the render queue configuration.
+   */
+  preRenderQueueConfiguration(host: IHost): void;
+
+  /**
+   * Method that is invoked after configuring the connection to the render queue and before configuring the Deadline Worker.
+   */
+  preWorkerConfiguration(host: IHost): void;
+
+  /**
+   * Method that is invoked after all configuration is done and worker started.
+   */
+  postWorkerLaunch(host: IHost): void;
+}
+
+/**
+ * Implementation of {@link IInstanceUserDataProvider}.
+ * Can be used as sub-class with override the desired methods
+ * to add custom user data commands for WorkerInstanceFleet or WorkerInstanceConfiguration.
+ */
+export class InstanceUserDataProvider extends Construct  implements IInstanceUserDataProvider{
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+  }
+
+  /**
+   * @inheritdoc
+   */
+  preCloudWatchAgent(_host: IHost): void {
+  }
+
+  /**
+   * @inheritdoc
+   */
+  preRenderQueueConfiguration(_host: IHost): void {
+  }
+
+  /**
+   * @inheritdoc
+   */
+  preWorkerConfiguration(_host: IHost): void {
+  }
+
+  /**
+   * @inheritdoc
+   */
+  postWorkerLaunch(_host: IHost): void {
+  }
+}
+/**
  * Configuration settings for Deadline Workers
  */
 export interface WorkerSettings {
@@ -86,6 +147,12 @@ export interface WorkerInstanceConfigurationProps {
    * @default The Worker is assigned the default settings as outlined in the WorkerSettings interface.
    */
   readonly workerSettings?: WorkerSettings;
+
+  /**
+   * An optional provider of user data commands to be injected at various points during the Worker configuration lifecycle.
+   * You can provide a subclass of InstanceUserDataProvider with the methods overridden as desired.
+   */
+  readonly userDataProvider?: IInstanceUserDataProvider;
 }
 
 /**
@@ -93,6 +160,14 @@ export interface WorkerInstanceConfigurationProps {
  * log files to CloudWatch, and configure various settings of the Deadline Worker.
  *
  * The configuration happens on instance start-up using user data scripting.
+ *
+ * This configuration performs the following steps in order:
+ * 1) Configure Cloud Watch Agent
+ * 2) Configure Deadline Worker RenderQueue connection
+ * 3) Configure Deadline Worker settings
+ *
+ * A `userDataProvider` can be specified that defines callback functions.
+ * These callbacks can be used to inject user data commands at different points during the Worker instance configuration.
  *
  * Security Considerations
  * ------------------------
@@ -105,12 +180,15 @@ export interface WorkerInstanceConfigurationProps {
 export class WorkerInstanceConfiguration extends Construct {
   constructor(scope: Construct, id: string, props: WorkerInstanceConfigurationProps) {
     super(scope, id);
-
+    props.userDataProvider?.preCloudWatchAgent(props.worker);
     if (props.cloudwatchLogSettings) {
       this.configureCloudWatchLogStream(props.worker, id, props.cloudwatchLogSettings);
     }
+    props.userDataProvider?.preRenderQueueConfiguration(props.worker);
     props.renderQueue?.configureClientInstance({ host: props.worker});
+    props.userDataProvider?.preWorkerConfiguration(props.worker);
     this.configureWorkerSettings(props.worker, id, props.workerSettings);
+    props.userDataProvider?.postWorkerLaunch(props.worker);
   }
 
   /**
--- a/packages/aws-rfdk/lib/deadline/test/version-query.test.ts
+++ b/packages/aws-rfdk/lib/deadline/test/version-query.test.ts
@@ -18,8 +18,6 @@ import {
   VersionQuery,
 } from '../lib';
 
-import { VERSION_QUERY_ASSET } from './asset-constants';
-
 test('VersionQuery constructor full version', () => {
   const app = new App();
   const stack = new Stack(app, 'Stack');
@@ -57,44 +55,6 @@ test('VersionQuery constructor full version', () => {
     ],
   }));
   expectCDK(stack).to(haveResourceLike('AWS::Lambda::Function', {
-    Code: {
-      S3Bucket: {
-        Ref: VERSION_QUERY_ASSET.Bucket,
-      },
-      S3Key: {
-        'Fn::Join': [
-          '',
-          [
-            {
-              'Fn::Select': [
-                0,
-                {
-                  'Fn::Split': [
-                    '||',
-                    {
-                      Ref: VERSION_QUERY_ASSET.Key,
-                    },
-                  ],
-                },
-              ],
-            },
-            {
-              'Fn::Select': [
-                1,
-                {
-                  'Fn::Split': [
-                    '||',
-                    {
-                      Ref: VERSION_QUERY_ASSET.Key,
-                    },
-                  ],
-                },
-              ],
-            },
-          ],
-        ],
-      },
-    },
     Handler: 'version-provider.handler',
     Role: {
       'Fn::GetAtt': [
--- a/packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
+++ b/packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
@@ -44,6 +44,8 @@ import {
   escapeTokenRegex,
 } from '../../core/test/token-regex-helpers';
 import {
+  IHost,
+  InstanceUserDataProvider,
   IRenderQueue,
   RenderQueue,
   Repository,
@@ -1460,6 +1462,40 @@ describe('HealthMonitor Tests', () => {
       Protocol: 'HTTP',
       TargetType: 'instance',
     }));
+  });
+
+  test('UserData is added', () => {
+    // WHEN
+    class UserDataProvider extends InstanceUserDataProvider {
+      preCloudWatchAgent(host: IHost): void {
+        host.userData.addCommands('echo preCloudWatchAgent');
+      }
+      preRenderQueueConfiguration(host: IHost): void {
+        host.userData.addCommands('echo preRenderQueueConfiguration');
+      }
+      preWorkerConfiguration(host: IHost): void {
+        host.userData.addCommands('echo preWorkerConfiguration');
+      }
+      postWorkerLaunch(host: IHost): void {
+        host.userData.addCommands('echo postWorkerLaunch');
+      }
+    }
+    const fleet = new WorkerInstanceFleet(wfstack, 'workerFleet', {
+      vpc,
+      workerMachineImage: new GenericLinuxImage({
+        'us-east-1': 'ami-any',
+      }),
+      renderQueue,
+      healthMonitor,
+      userDataProvider: new UserDataProvider(wfstack, 'UserDataProvider'),
+    });
+    const userData = fleet.fleet.userData.render();
+
+    // THEN
+    expect(userData).toContain('echo preCloudWatchAgent');
+    expect(userData).toContain('echo preRenderQueueConfiguration');
+    expect(userData).toContain('echo preWorkerConfiguration');
+    expect(userData).toContain('echo postWorkerLaunch');
   });
 });
 
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/compute-tier.ts
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/compute-tier.ts
@@ -11,6 +11,8 @@ import {
 } from '@aws-cdk/aws-ec2';
 import * as cdk from '@aws-cdk/core';
 import {
+  IHost,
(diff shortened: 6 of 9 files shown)


examples/deadline/All-In-AWS-Infrastructure-Basic/scripts/configure_worker.sh
```diff
@@ -0,0 +1 @@
+mkdir test_dir
```
ddneilson: nit: Missing newline at end of file.

yarn.lock
```diff
@@ -10460,6 +10460,17 @@ ts-node@^8.0.2:
     source-map-support "^0.5.17"
     yn "3.1.1"
 
+ts-node@^9.0.0:
```
ddneilson: I suspect this wasn't an intended change.

packages/aws-rfdk/lib/deadline/test/version-query.test.ts
```diff
@@ -57,44 +55,6 @@ test('VersionQuery constructor full version', () => {
     ],
   }));
   expectCDK(stack).to(haveResourceLike('AWS::Lambda::Function', {
-    Code: {
```
ddneilson, line 60: How come all of this needed to get deleted?

kozlove-aws: I was just in process updating description. 
`VERSION_QUERY_ASSET.Bucket` and `VERSION_QUERY_ASSET.Key` are changed periodically so we can't relay on tests with these constants.

ddneilson: They should only be changed when the file(s) in its lambda change. Is it changing, randomly, without changing those files?

kozlove-aws: Yes, I have not changed lambda in my MR and have to adjust these constants for make test pass.

ddneilson (approved):

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,26 @@ import {
...
+export interface IInstanceUserDataProvider {
+  // Method that is invoked before configuration Cloud Watch Agent.
+  preCloudWatchAgent(host: IHost): void;
+  // Method that is invoked before Deadline worker and renderqueue configuration .
+  preDeadlineConfiguration(host: IHost): void;
+  // Method that is invoked after all configuration is done and worker started.
+  postWorkerLaunch(host: IHost): void;
+}
+
+export class InstanceUserDataProvider implements IInstanceUserDataProvider{
```
ddneilson: This should have a docstring explaining what it's for.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -86,6 +106,8 @@ export interface WorkerInstanceConfigurationProps {
    * @default The Worker is assigned the default settings as outlined in the WorkerSettings interface.
    */
   readonly workerSettings?: WorkerSettings;
+
+  readonly userDataProvider?: IInstanceUserDataProvider;
```
ddneilson, line 155: I'm surprised this got through the linting. This needs a docstring.

ddneilson (changes requested):

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
kozlove-aws: I missed it but surprised as well that linting missed it too.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
```diff
@@ -43,6 +48,22 @@ class ComputeTierProps(StackProps):
...
     bastion: Optional[BastionHostLinux] = None
 
+class UserDataProvider(InstanceUserDataProvider):
+    def pre_cloud_watch_agent(self, host) -> None:
+        host.user_data.add_commands("echo preCloudWatchAgent")
+    def pre_deadline_configuration(self, host) -> None:
+        host.user_data.add_commands("echo preDeadlineConfiguration")
+    def post_worker_launch(self, host) -> None:
+        host.user_data.add_commands("echo postWorkerLaunch")
+        test_script=Asset(host.node.scope, "SampleAsset",
```
jusiskin: I'm wondering if it would be better to have the base class (`InstanceUserDataProvider`) be a sub-class of `Construct`. That way, it has its own scope in the construct tree and any child constructs (like these assets) can be added to its scope in the constructor and not have to create constructs in the `host.node.scope` like this. It would look like this:

```py
class UserDataProvider(InstanceUserDataProvider):
    def __init__(self, scope: Construct, id: str):
        super().__init__(scope, id);
        self.test_script = Asset(self, 'test_script', path=...)

    def post_worker_launch(self, host: IHost) -> None:
        self.test_script.grant_read(host)
        local_path = host.user_data.add_s3_download_command(
            bucket=test_script.bucket,
            bucket_key=test_script.s3_object_key
        )
        host.user_data.add_execute_file_command(file_path=local_path)
```

what do you think?

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
```diff
@@ -43,6 +48,22 @@ class ComputeTierProps(StackProps):
     # The bastion host to  allow connection to Worker nodes.
     bastion: Optional[BastionHostLinux] = None
 
+class UserDataProvider(InstanceUserDataProvider):
+    def pre_cloud_watch_agent(self, host) -> None:
+        host.user_data.add_commands("echo preCloudWatchAgent")
+    def pre_deadline_configuration(self, host) -> None:
```
jusiskin: nit: pep8 requires a blank line between methods

examples/deadline/All-In-AWS-Infrastructure-Basic/scripts/configure_worker.sh
```diff
@@ -0,0 +1 @@
...
```
jusiskin, line 4: This is missing the:

```bash
#!/bin/sh
```

at the top.

Also, a comment would help to document that this script is only here to illustrate how to add custom user data when using our `WorkerInstanceFleet` and `WorkerInstanceConfiguration` constructs

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,26 @@ import {
   Version,
 } from './version';
 
+/**
+ * Provider for adding user data scripts
+ */
+export interface IInstanceUserDataProvider {
+  // Method that is invoked before configuration Cloud Watch Agent.
+  preCloudWatchAgent(host: IHost): void;
```
jusiskin: these should be made into proper TSDoc strings so that they are visible in our API docs and discoverable from IDEs.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,26 @@ import {
...
+  // Method that is invoked before Deadline worker and renderqueue configuration .
+  preDeadlineConfiguration(host: IHost): void;
```
jusiskin: These can be made `readonly`

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,26 @@ import {
   Version,
 } from './version';
 
+/**
+ * Provider for adding user data scripts
+ */
+export interface IInstanceUserDataProvider {
```
jusiskin, line 37: Do we want another callback to run before the worker launch? I can imagine a case where someone might want to set something up after Deadline is already configured to properly connect to the Render Queue, but before the Worker is launched.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,26 @@ import {
...
+  postWorkerLaunch(host: IHost): void;
+}
+
+export class InstanceUserDataProvider implements IInstanceUserDataProvider{
+  preCloudWatchAgent(_host: IHost): void {
+  }
+  preDeadlineConfiguration(_host: IHost): void {
+  }
+  postWorkerLaunch(_host: IHost): void {
+  }
```
jusiskin: nit: we should have a blank line between these functions to be consistent with the rest of the codebase. Same applies elsewhere

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
```diff
@@ -1461,6 +1469,36 @@ describe('HealthMonitor Tests', () => {
...
+      renderQueue,
+      healthMonitor,
+      userDataProvider: new UserDataProvider(),
+    });
+    const userData = fleet.fleet.userData.render();
+
+    // THEN
+    expect(userData).toContain('echo preCloudWatchAgent');
+    expect(userData).toContain('echo preDeadlineConfiguration');
+    expect(userData).toContain('echo postWorkerLaunch');
```
jusiskin: These make no guarantees that the user data are injected in the correct places. Any ideas on how we can test that here?

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
```diff
@@ -92,12 +94,18 @@ beforeEach(() => {
...
+      host.userData.addCommands('mkdir test_dir');
+    }
+  }
   const fleet = new WorkerInstanceFleet(wfstack, 'workerFleet', {
     vpc,
     workerMachineImage: new GenericWindowsImage({
       'us-east-1': 'ami-any',
     }),
     renderQueue,
+    userDataProvider: new userDataProvider(),
```
jusiskin: I'm not sure we need this to be added here. The test appears to be specifically for testing defaults and this new user data provider is optional.

jusiskin (changes requested):
Generally, this looks good. The main request is to add an additional callback and make the base class a construct.

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
kozlove-aws: Sure. It just artifacts.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
kozlove-aws: 'readonly' modifier can only appear on a property declaration or index signature.

jusiskin: Ah my bad. Thanks for confirming.

jusiskin:
We haven't been great about doing this lately, but do you mind also adding documentation for this in `packages/aws-rfdk/lib/deadline/README.md`?

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
ddneilson: This was missed.

ddneilson (changes requested):

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,37 @@ import {
   Version,
 } from './version';
 
+/**
+ * Provider for adding user data scripts
+ */
+export interface IInstanceUserDataProvider {
```
ddneilson, line 37: The docstrings on this and the `InstanceUserDataProvider` should be better. Approach it from the perspective of someone who is not reading the source code, and only reading the documentation. Will the documentation from these docstrings help you understand what the class/interface is used for? How about how it fits in with the `WorkerInstanceConfiguration` class?

Without knowing the internal workings of the `WorkerInstanceConfiguration` class we want the reader of the documentation to understand where each of the callbacks that are provided by this interface fit into the worker configuration workflow.

kozlove-aws: Docstring for WorkerInstanceConfiguration was updated as well for describe which steps we have.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,39 @@ import {
   Version,
 } from './version';
 
+/**
+ * Provider for adding user data scripts
+ * Methods of this class will be invoked in WorkerInstanceConfiguration
```
ddneilson: nit: "Methods of this class"

This isn't a class... it's an interface

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,39 @@ import {
...
+  preCloudWatchAgent(host: IHost): void;
+  // Method that is invoked before RenderQueue configuration.
+  preRenderQueueConfiguration(host: IHost): void;
+  // Method that is invoked before Deadline worker configuration.
+  preDeadlineConfiguration(host: IHost): void;
+  // Method that is invoked after all configuration is done and worker started.
+  postWorkerLaunch(host: IHost): void;
+}
+
+export class InstanceUserDataProvider extends Construct  implements IInstanceUserDataProvider{
```
ddneilson, line 64: Missing a docstring on this class.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -86,13 +119,25 @@ export interface WorkerInstanceConfigurationProps {
...
+  readonly userDataProvider?: IInstanceUserDataProvider;
 }
 
 /**
  * This construct can be used to configure Deadline Workers on an instance to connect to a RenderQueue, stream their
  * log files to CloudWatch, and configure various settings of the Deadline Worker.
  *
  * The configuration happens on instance start-up using user data scripting.
+ * 
+ * The configuration has next steps:
```
ddneilson: Maybe "This configuration performs the following steps in order:"
?

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
ddneilson: Unless I'm wrong here, the `//` syntax is just a comment and not a docstring. i.e. it doesn't end up in the API documentation.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,39 @@ import {
...
+  // Method that is invoked after all configuration is done and worker started.
+  postWorkerLaunch(host: IHost): void;
+}
+
+export class InstanceUserDataProvider extends Construct  implements IInstanceUserDataProvider{
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+  }
+
+  preCloudWatchAgent(_host: IHost): void {
```
ddneilson, line 72: We'll also want docstrings on each of these, though it can probably just be an `@inheritdoc` docstring.

ddneilson (changes requested):

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,39 @@ import {
...
+ * Methods of this class will be invoked in WorkerInstanceConfiguration
+ * on different stages of worker configuration
+ */
+export interface IInstanceUserDataProvider {
+  // Method that is invoked before configuration Cloud Watch Agent.
+  preCloudWatchAgent(host: IHost): void;
+  // Method that is invoked before RenderQueue configuration.
+  preRenderQueueConfiguration(host: IHost): void;
+  // Method that is invoked before Deadline worker configuration.
+  preDeadlineConfiguration(host: IHost): void;
```
jusiskin: May I propose a rename of this callback from `preDeadlineConfiguration` &rarr; `preWorkerConfiguration`? Technically the Deadline client has already been configured to connect to the render queue and the only configuration that follows this hook/callback is the Worker database configuration.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -86,13 +119,25 @@ export interface WorkerInstanceConfigurationProps {
...
  * log files to CloudWatch, and configure various settings of the Deadline Worker.
  *
  * The configuration happens on instance start-up using user data scripting.
+ * 
+ * The configuration has next steps:
+ * - Configure Cloud Watch Agent
+ * - Configure Deadline Worker RenderQueue connection
+ * - Configure Deadline Worker settings
+ * 
+ * When userDataProvider defined, its methods invoked before every step and after configuration.
```
jusiskin: How about:

```
A `userDataProvider` can be specified that defines callback functions. These callbacks can be used to inject user data commands at different points during the Worker instance configuration.
```

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -86,13 +119,25 @@ export interface WorkerInstanceConfigurationProps {
    * @default The Worker is assigned the default settings as outlined in the WorkerSettings interface.
    */
   readonly workerSettings?: WorkerSettings;
+
+  /**
+   * The provider for User data scripts that will be invoke during worker configuration.
+   */
+  readonly userDataProvider?: IInstanceUserDataProvider;
```
jusiskin: Few things:
*   nit: just a bit of grammar/terminology cleanup
*   might help to guide users to the APIs they need to get this working

How about: 

>   An optional provider of user data commands to be injected at various points during the Worker configuration lifecycle. You can provide a subclass of `InstanceUserDataProvider` with the methods overridden as desired.

This would also need to be applied to the `userDataProvider` docstring on `WorkerInstanceFleetProps` as well.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
```diff
@@ -43,6 +48,29 @@ class ComputeTierProps(StackProps):
     # The bastion host to  allow connection to Worker nodes.
     bastion: Optional[BastionHostLinux] = None
 
+class UserDataProvider(InstanceUserDataProvider):
+    def __init__(self, scope: Construct, stack_id: str):
+        super().__init__(scope, stack_id)
+        self.test_script=Asset(scope, "SampleAsset",
+            path=os.path.join(os.getcwd(), "..", "scripts", "configure_worker.sh")
+        )
+    def pre_cloud_watch_agent(self, host) -> None:
```
jusiskin: nit: should be a blank line here as well

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/config.py
```diff
@@ -19,7 +19,7 @@ class AppConfig:
     """
     def __init__(self):
         # A map of regions to Deadline Client Linux AMIs.
-        self.deadline_client_linux_ami_map: Mapping[str, str] = {'region': 'ami-id'}
+        self.deadline_client_linux_ami_map: Mapping[str, str] = {'us-west-2': 'ami-0b12631d34ca9939f'}
```
jusiskin: Looks like the changes in this file were accidentally committed

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,45 @@ import {
   Version,
 } from './version';
 
+/**
+ * Provider for adding user data scripts
+ * Methods of this interface will be invoked in WorkerInstanceConfiguration
+ * on different stages of worker configuration
+ */
+export interface IInstanceUserDataProvider {
+  /** Method that is invoked before configuration Cloud Watch Agent. */
```
jusiskin: These are still not in correct TSDoc syntax. Also a minor grammatical fix. Should be:

```ts
 /**
  * Method that is invoked before configuring the Cloud Watch Agent.
  */
```

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
jusiskin: I think this needs more documentation. This class is the API that RFDK users should use in order to achieve custom user data. In here, we should explain that users can sub-class this and override the desired methods to add custom user data commands for `WorkerInstanceFleet` or `WorkerInstanceConfiguration`.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -29,6 +29,48 @@ import {
...
+}
+
+/**
+ * Implementation of {@link IInstanceUserDataProvider}.
+ */
+export class InstanceUserDataProvider extends Construct  implements IInstanceUserDataProvider{
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+  }
+  /** @inheritdoc */
```
jusiskin: nit: missing blank line

jusiskin (changes requested):
Looking better, but there are a few more things to address here - mostly around the docstrings.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
jusiskin: Resp;ved

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
jusiskin: Looks like this was addressed for `WorkerInstanceConfigurationProps` but not `WorkerInstanceFleetProps`.

ddneilson (dismissed):

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
```diff
@@ -43,6 +48,30 @@ class ComputeTierProps(StackProps):
...
+            path=os.path.join(os.getcwd(), "..", "scripts", "configure_worker.sh")
+        )
+
+    def pre_cloud_watch_agent(self, host) -> None:
+        host.user_data.add_commands("echo preCloudWatchAgent")
+
+    def pre_render_queue_configuration(self, host) -> None:
+        host.user_data.add_commands("echo preRenderQueueConfiguration")
+
+    def pre_deadline_configuration(self, host) -> None:
```
jusiskin: I believe this needs to be updated with the new hook name.

jusiskin (changes requested):
I notice an issue in the python example that needs to be updated after some PR changes.

ddneilson (approved):

jusiskin (approved):

Issue #241, opened 2020-11-19
Configurable expiry for generated X.509 certificates


ddneilson:
All X.509 certificates generated by the `X509CertificatePem` construct have a default expiry of 3 years, and that is not configurable. The request is to make that expiry configurable.

### Use Case

Prevent render farms from becoming inoperable after 3 years by extending the duration of X.509 certificates.

### Other

N/A

* [x] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

---

This is a :rocket: Feature Request

horsmand:
Implemented by #242

PR #242, opened 2020-11-19
feat(core): Add configurable expiry to X.509 certificates

ddneilson:
Implements: https://github.com/aws/aws-rfdk/issues/241

----

### DIFF
--- a/packages/aws-rfdk/lib/core/lib/x509-certificate.ts
+++ b/packages/aws-rfdk/lib/core/lib/x509-certificate.ts
@@ -26,6 +26,7 @@ import {
 import { RetentionDays } from '@aws-cdk/aws-logs';
 import { ISecret, Secret } from '@aws-cdk/aws-secretsmanager';
 import {
+  Annotations,
   Construct,
   CustomResource,
   Duration,
@@ -88,6 +89,13 @@ export interface X509CertificatePemProps {
    * @default: None. The generated certificate will be self-signed
    */
   readonly signingCertificate?: X509CertificatePem;
+
+  /**
+   * The number of days that the generated certificate will be valid for.
+   *
+   * @default 1095 days (3 years)
+   */
+  readonly validFor?: number;
 }
 
 /**
@@ -263,6 +271,10 @@ export class X509CertificatePem extends X509CertificateBase implements IX509Cert
       encryptionKey: props.encryptionKey,
     });
 
+    if ((props.validFor ?? 1) < 1 && !Token.isUnresolved(props.validFor)) {
+      Annotations.of(this).addError('Certificates must be valid for at least one day.');
+    }
+
     props.signingCertificate?.cert.grantRead(this.lambdaFunc);
     props.signingCertificate?.key.grantRead(this.lambdaFunc);
     props.signingCertificate?.passphrase.grantRead(this.lambdaFunc);
@@ -295,6 +307,7 @@ export class X509CertificatePem extends X509CertificateBase implements IX509Cert
         ],
       },
       SigningCertificate: signingCertificate,
+      CertificateValidFor: props.validFor?.toString(),
     };
     const resource = new CustomResource(this, 'Default', {
       serviceToken: this.lambdaFunc.functionArn,
--- a/packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/test/certificate.test.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/test/certificate.test.ts
@@ -58,6 +58,8 @@ test('generate self-signed', async () => {
   const certVerification: string = certOut.stdout;
   const keyOut = await exec(`openssl rsa -noout -text -check -passin env:PW -in ${keyFileName}`, { env: { PW: passphrase }});
   const keyVerification = keyOut.stdout;
+  const expiryDate = new Date();
+  expiryDate.setDate(expiryDate.getDate() + 3*365);
 
   // THEN
   expect(certificate.cert).toContain('-----BEGIN CERTIFICATE-----');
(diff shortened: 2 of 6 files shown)


horsmand (dismissed):

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -88,6 +88,13 @@ export interface X509CertificatePemProps {
    * @default: None. The generated certificate will be self-signed
    */
   readonly signingCertificate?: X509CertificatePem;
+
+  /**
+   * The number of days that the generated certificate will be valid for.
+   *
+   * @default 1095 days (3 years)
+   */
+  readonly validFor?: number;
```
jusiskin, line 98: How about using a `cdk.Duration` here? It allows users to be more expressive. Same feedback for `.fromGenerated`.

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/test/certificate.test.ts
```diff
@@ -76,6 +76,32 @@ test('generate self-signed', async () => {
...
+  await writeAsciiFile(crtFileName, certificate.cert);
+  await writeAsciiFile(keyFileName, certificate.key);
+  const certOut = await exec(`openssl x509 -noout -text -in ${crtFileName}`);
+  const certVerification: string = certOut.stdout;
+  const expiryDate = new Date();
+  expiryDate.setDate(expiryDate.getDate() + 5*365);
+
+  // THEN
+  // ex: Not After : May 22 22:13:24 2023 GMT
+  expect(certVerification).toMatch(new RegExp(`Not After.*${expiryDate.getFullYear()} GMT`));
```
jusiskin, line 106: This only checks the year. Do we care about having more strict validation?

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/test/certificate.test.ts
```diff
@@ -92,7 +118,7 @@ test('generate signed certificate', async () => {
   const passphrase: string = 'test_passphrase';
 
   // WHEN
-  const certificate = await Certificate.fromGenerated(certName, passphrase, ca);
+  const certificate = await Certificate.fromGenerated(certName, passphrase, undefined, ca);
```
jusiskin, line 125: It appears `undefined` means the default expiry of three years. Should we assert that?

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/types.ts
```diff
@@ -165,6 +171,10 @@ export function implementsIX509CertificateGenerate(value: any): boolean {
   if (!implementsIX509ResourceProperties(value)) { return false; }
   if (!implementsDistinguishedNameProps(value.DistinguishedName)) { return false; }
   if (value.SigningCertificate && !implementsISecretCertificate(value.SigningCertificate)) { return false; }
+  if (value.CertificateValidFor) {
+    if (typeof(value.CertificateValidFor) !== 'string') { return false; }
+    if (value.CertificateValidFor === '' || isNaN(Number(value.CertificateValidFor))) { return false; }
```
jusiskin, line 176: Should we add a validation here to ensure we have a positive number?

jusiskin (changes requested):
Looks good. Just a few suggestions

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
ddneilson: I had the same thought, but `Duration`'s units are at most days (ex: no 'years', 'months', etc). So, it doesn't really seem to add much other than extra syntax for the customer and testing on our side.

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/test/certificate.test.ts
ddneilson: It's a bit of transparent testing; intent is just to make sure that the days value is getting all the way through to openssl. The default is 3 years, so as long as the year changes by more than 3 then we know that the value was passed through to the openssl command.

I'm also trying to avoid leap-year rounding issues by going for a more coarse-grained check.

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/types.ts
ddneilson: Not here, but it makes some sense to assert that in the construct itself. Fail early...

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/test/certificate.test.ts
ddneilson: No harm in it, sure.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
jusiskin: Ah that makes sense. Resolved

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/test/certificate.test.ts
jusiskin: This makes sense. Although we may still encounter leap year rounding issues in the form of false-negatives if we execute these tests on a date near December 31/January 1st. This is a minor edge-case though - so let's just keep it as-is

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/types.ts
jusiskin: I'm okay with this approach, although it does miss the case where the customer passes in a CDK token. Sounds like that should be avoided since, as this PR points out, it could lead to certificates being regenerated and breaking any manually configured clients.

On that basis, do you think we should also validate that the expiration is not a token?

ddneilson: Token should be allowable. That'll get expanded into its value by CloudFormation.

Though, I suppose that means that my validation check isn't quite right. :-/

ddneilson: Updated in the latest revision. I'm leaving the negative-value test out of the type validator. Those validators do not give useful diagnostic output; they're mostly to make sure that we've created the CustomResource properly.

openssl will die with an error like `req: Non-positive number "-1" for -days` if a negative value is passed through.

jusiskin: Great - thanks for looking into this.

jusiskin (approved):

horsmand (approved):

PR #243, opened 2020-11-19
chore(integ): Make assume role optional in automated integration test script

jericht:
Updated the `run-e2e-automated.sh` script so that it does not try to refresh credentials and assume an IAM role if the relevant environment variables are not specified. This allows the script to be used in situations where assuming an IAM role is not required (e.g. the role running the tests has sufficient permissions).

## Testing
**New case:**
- Defined just the `USER_ACCEPTS_SSPL_FOR_RFDK_TESTS=true` environment variable
- Ran `yarn run e2e-automated` and ensured at least one integration test successfully deployed and passed (did both the Repository and RenderQueue tests)

**Old case:**
- Added an `echo` statement inside the `if` block that ensures the IAM role env vars are defined
- Defined all the IAM role env vars
- Ran the script and ensured the `echo` statement was reached

----

### DIFF
--- a/integ/scripts/bash/run-e2e-automated.sh
+++ b/integ/scripts/bash/run-e2e-automated.sh
@@ -5,6 +5,9 @@
 
 # Used as the entry point to run the tests in an automated way. This expects the following env vars to be set:
 # USER_ACCEPTS_SSPL_FOR_RFDK_TESTS - Needs to be 'true' to use these tests
+#
+# Each of these env vars must be set to assume an IAM role in a separate account.
+# If at least one is not set, an IAM role will not be assumed.
 # ROLE_ARN - The role that will be assumed to deploy the integ tests
 # ROLE_SESSION_NAME - A name for the assume role session
 # ROLE_EXTERNAL_ID - The external ID matching the assume role
@@ -30,16 +33,20 @@ refreshcreds() {
     unset CREDS
 }
 
+if [ ! -z ${ROLE_ARN+x} ] && \
+   [ ! -z ${ROLE_SESSION_NAME+x} ] && \
+   [ ! -z ${ROLE_EXTERNAL_ID+x} ]; then
+  # Setup the hook that runs before any interactions with AWS to refresh the credentials being used.
+  # There is a 1 hour timeout on these credentials that cannot be adjusted.
+  export -f refreshcreds
+  export PRE_AWS_INTERACTION_HOOK=refreshcreds
+
+  refreshcreds
+fi
+
 # Basic integ test configuration
 export SKIP_TEST_CONFIG=true
 export DEADLINE_VERSION="10.1.11.5"
-
-# Setup the hook that runs before any interactions with AWS to refresh the credentials being used.
-# There is a 1 hour timeout on these credentials that cannot be adjusted.
-export -f refreshcreds
-export PRE_AWS_INTERACTION_HOOK=refreshcreds
-
-refreshcreds
 
 ./scripts/bash/rfdk-integ-e2e.sh
 


integ/scripts/bash/run-e2e-automated.sh
```diff
@@ -30,19 +32,30 @@ refreshcreds() {
     unset CREDS
 }
 
+ASSUME_ROLE="false"
```
horsmand, line 35: A capitalized variable like this is used to denote an environment variable. If it isn't being exported, it should be lowercase.

Also, it you're using the string `"false"` instead of the boolean `false`, which I guess is fine, but not really what the intended purpose is.

integ/scripts/bash/run-e2e-automated.sh
```diff
@@ -30,19 +32,30 @@ refreshcreds() {
...
 
-unset AWS_ACCESS_KEY_ID
-unset AWS_SECRET_ACCESS_KEY
-unset AWS_SESSION_TOKEN
\ No newline at end of file
+if [ $ASSUME_ROLE = "true" ]; then
+  unset AWS_ACCESS_KEY_ID
+  unset AWS_SECRET_ACCESS_KEY
+  unset AWS_SESSION_TOKEN
+fi
```
horsmand: Missing new line, looks like this was a miss by me when I originally wrote it.

integ/scripts/bash/run-e2e-automated.sh
horsmand: The convention in bash is to only use a capitalized variable name for environment variables. If it isn't being exported, it should be lowercase.

Also, it you're using the string `"false"` instead of the boolean `false`, which I guess is fine, but not really what the intended purpose is.

integ/scripts/bash/run-e2e-automated.sh
```diff
@@ -5,6 +5,8 @@
 
 # Used as the entry point to run the tests in an automated way. This expects the following env vars to be set:
 # USER_ACCEPTS_SSPL_FOR_RFDK_TESTS - Needs to be 'true' to use these tests
+#
+# Each of these env vars must be set to assume an IAM role. If at least one is not set, an IAM role will not be assumed.
```
horsmand: Maybe clarify that this is to assume a role in a separate account. If you would like to assume a role in the same account, we'd need to do some extra work to make the external id optional. I don't see that use case being required for anything though.

horsmand (changes requested):
Just a few really minor comments. I see that you are setting up this `assume_role` variable to be used when setting the exports as well as around the `unset` calls at the end. There is no harm in calling `unset` on a variable that already isn't set, so that if block could be removed and then the separate boolean wouldn't be necessary, you could just do the check to see if those 3 env vars are set in the if block around the exports.

integ/scripts/bash/run-e2e-automated.sh
jericht: Sounds good. Added this clarification. and yeah, I don't think we'll need to make the external ID optional any time soon.

integ/scripts/bash/run-e2e-automated.sh
jericht: I've removed this variable since I realized it wasn't necessary after reading your comment. For some reason, I thought the `unset` would affect the parent environment even though it doesn't.

integ/scripts/bash/run-e2e-automated.sh
jericht: Ah good catch. Added

horsmand (approved):

jusiskin:
The code looks good, but I just want to confirm that you also tested the code path where all the assume role variables are provided - only because the PR description doesn't mention it.

jusiskin (approved):

PR #248, opened 2020-11-23
fix(integ): Support last changes in Deadline 10.1.11

kozlove-aws:
## Problem
The last Deadline 10.1.11 brings some changes that make RFDK integration tests fail.
1) New AMIs have worker autostart is off that lead to unable to configure pools and groups
2) Restarting `DeadlineLauncher` right after start leads to worker continue without last changes in `deadline.ini`
3) New region validation does not allow to set reqion that wasn't previosly added in network configuration.

## Solution
1) If worker was not found was added default worker to the list.
2) After stop deadline launcher was added command to kill deadline worker process.
3) Were changed tests to expect `unrecognized` region instead of `testregion`

## Testing
For testing were ran integration tests and make sure that all tests passed.

----

### DIFF
--- a/integ/components/deadline/deadline_03_workerFleetHttp/test/deadline_03_workerFleetHttp.test.ts
+++ b/integ/components/deadline/deadline_03_workerFleetHttp/test/deadline_03_workerFleetHttp.test.ts
@@ -123,7 +123,8 @@ describe.each(testCases)('Deadline WorkerFleet tests (%s)', (_, id) => {
       };
       return awaitSsmCommand(bastionId, params).then( response => {
         var responseOutput = response.output;
-        expect(responseOutput).toMatch(/testpool\ntestgroup\ntestregion/);
+        // Starting Deadline 10.1.11 regions that wasn't added do not apply to worker and returned as unrecognized.
+        expect(responseOutput).toMatch(/testpool\ntestgroup\n(?:unrecognized|testregion)/);
       });
     });
 
--- a/packages/aws-rfdk/lib/deadline/scripts/powershell/configureWorker.ps1
+++ b/packages/aws-rfdk/lib/deadline/scripts/powershell/configureWorker.ps1
@@ -70,6 +70,10 @@ Foreach-Object {
         $WORKER_NAMES+="$WORKER_NAME_PREFIX-$workerConfigName"
     }
 }
+if ( $WORKER_NAMES.count -eq 0 )
+{
+    $WORKER_NAMES+=$WORKER_NAME_PREFIX
+}
 
 $WORKER_NAMES_CSV=$WORKER_NAMES -join ","
 
@@ -99,10 +103,13 @@ if($WORKER_POOLS) {
 
 $serviceName="deadline10launcherservice"
 If (Get-Service $serviceName -ErrorAction SilentlyContinue) {
-    Restart-Service $serviceName
+    Stop-Service $serviceName
+    taskkill /f /fi "IMAGENAME eq deadlineworker.exe"
+    Start-Service $serviceName
 } Else {
     $DEADLINE_LAUNCHER = $DEADLINE_PATH + '/deadlinelauncher.exe'
     & $DEADLINE_LAUNCHER -shutdownall | Out-Default
+    taskkill /f /fi "IMAGENAME eq deadlineworker.exe"
     & $DEADLINE_LAUNCHER
 }
 
--- a/integ/components/deadline/deadline_04_workerFleetHttps/test/deadline_04_workerFleetHttps.test.ts
+++ b/integ/components/deadline/deadline_04_workerFleetHttps/test/deadline_04_workerFleetHttps.test.ts
@@ -175,7 +175,8 @@ describe.each(testCases)('Deadline WorkerFleetHttps tests (%s)', (_, id) => {
       };
       return awaitSsmCommand(bastionId, params).then( response => {
         var responseOutput = response.output;
-        expect(responseOutput).toMatch(/testpool\ntestgroup\ntestregion/);
+        // Starting Deadline 10.1.11 regions that wasn't added do not apply to worker and returned as unrecognized.
+        expect(responseOutput).toMatch(/testpool\ntestgroup\n(?:unrecognized|testregion)/);
       });
     });
 
--- a/packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
+++ b/packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
@@ -81,7 +81,7 @@ for file in /var/lib/Thinkbox/Deadline10/slaves/*
 do
   file="${file##*/}"
   workerSuffix="${file%%.*}"
-  if [ -z "$workerSuffix" ]; then
+  if [[ -z "$workerSuffix" || "$workerSuffix" = "*" ]]; then
     WORKER_NAMES+=( "$WORKER_NAME_PREFIX" )
   else
     WORKER_NAMES+=( "$WORKER_NAME_PREFIX"-$workerSuffix )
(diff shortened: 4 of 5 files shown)


packages/aws-rfdk/lib/deadline/lib/rq-connection.ts
```diff
@@ -111,6 +111,7 @@ export abstract class RenderQueueConnection {
         'if [ -f "/etc/profile.d/deadlineclient.sh" ]; then',
         '  source "/etc/profile.d/deadlineclient.sh"',
         'fi',
+        '"\${DEADLINE_PATH}/deadlinecommand" -SetIniFileSetting LaunchSlaveAtStartup True',
```
grbartel: We do not want to set this here. This class is a generic class used to set up any deadline isntance being able to connect to a render queue not specifically workers as such your changes make it so any Workstation, Pulse or other component automatically starts up workers.

This setting is also already being set in the ConfigureWorker.sh which is where I would expect all of this code to exist.

packages/aws-rfdk/lib/deadline/lib/rq-connection.ts
```diff
@@ -122,6 +123,7 @@ export abstract class RenderQueueConnection {
       host.userData.addCommands(
         '$ErrorActionPreference = "Stop"',
         '$DEADLINE_PATH = (get-item env:"DEADLINE_PATH").Value',
+        '& "$DEADLINE_PATH/deadlinecommand.exe" -SetIniFileSetting LaunchSlaveAtStartup True | Out-Default',
```
grbartel: Same as Above this class should have no Worker specific code in it.

grbartel (changes requested):

packages/aws-rfdk/lib/deadline/scripts/powershell/configureWorker.ps1
```diff
@@ -99,7 +99,9 @@ if($WORKER_POOLS) {
 
 $serviceName="deadline10launcherservice"
 If (Get-Service $serviceName -ErrorAction SilentlyContinue) {
-    Restart-Service $serviceName
+    Stop-Service $serviceName
+    taskkill /f /im deadlineworker.exe -ErrorAction SilentlyContinue
+    Start-Service $serviceName
 } Else {
     $DEADLINE_LAUNCHER = $DEADLINE_PATH + '/deadlinelauncher.exe'
     & $DEADLINE_LAUNCHER -shutdownall | Out-Default
```
ryyakobe, line 111: Just making sure we don't miss anything:
For Linux here we have sudo `killall -w deadlineworker || true`.

And on Windows we don't kill `deadlineworker` here. Could you please explain why?

integ/components/deadline/deadline_03_workerFleetHttp/test/deadline_03_workerFleetHttp.test.ts
```diff
@@ -123,7 +123,8 @@ describe.each(testCases)('Deadline WorkerFleet tests (%s)', (_, id) => {
       };
       return awaitSsmCommand(bastionId, params).then( response => {
         var responseOutput = response.output;
-        expect(responseOutput).toMatch(/testpool\ntestgroup\ntestregion/);
+        // We are not expecting testregion because Deadline worker return unrecognized if region does not exist
+        expect(responseOutput).toMatch(/testpool\ntestgroup\nunrecognized/);
```
ryyakobe: Where do we set these values "testgroup", "testpool" and "testregion" ?
Maybe we could add a test with region = "us-west-2" just to be sure, that we get the correct result if the region exists, because now we only test for "unrecognized".

kozlove-aws: By default we have only `unrecognized` region. So we can only test that we have `unrecognized` instead of `none` by default.

packages/aws-rfdk/lib/deadline/scripts/powershell/configureWorker.ps1
kozlove-aws: I have missed it here.

integ/components/deadline/deadline_03_workerFleetHttp/test/deadline_03_workerFleetHttp.test.ts
grbartel: Is it possible for us to add to the test to have it create the region.

grbartel: also this change makes it so these tests no longer work with older versions of Deadline. Is the plan to drop support for those versions of Deadline?

packages/aws-rfdk/lib/deadline/lib/rq-connection.ts
kozlove-aws: After offline discussion we have found a way not to start worker before configuration Pools and Groups.

integ/components/deadline/deadline_03_workerFleetHttp/test/deadline_03_workerFleetHttp.test.ts
kozlove-aws: Code was modified to expect both cases with `urecognized` region and `testregion`

ryyakobe: This looks good to me. Resolving the conversation but tagging @grbartel so he notices the change.

ryyakobe (dismissed):
Looks good! Thanks Eugene!

integ/components/deadline/deadline_03_workerFleetHttp/test/deadline_03_workerFleetHttp.test.ts
```diff
@@ -123,7 +123,8 @@ describe.each(testCases)('Deadline WorkerFleet tests (%s)', (_, id) => {
       };
       return awaitSsmCommand(bastionId, params).then( response => {
         var responseOutput = response.output;
-        expect(responseOutput).toMatch(/testpool\ntestgroup\ntestregion/);
+        // Starting Deadline 10.1.11 regions that wasn't added do not apply to worker and returned as unrecognized.
+        expect([/testpool\ntestgroup\nunrecognized/,/testpool\ntestgroup\ntestregion/]).toContain(responseOutput);
```
ddneilson: This is inverted. It *works* but a better way to do it would be:
```js
expect(responseOutput).toMatch(/testpool\ntestgroup\n(?:unrecognized|testregion)/);
```

The leading & trailing '/' denote a regex, and the `(?:...)` syntax in the regex denotes a non-capturing group that matches words on either side of the `|`

You can use the [typescript playground](https://www.typescriptlang.org/play) to experiment, and convince yourself that works:
```js
var ss1 = "testpool\ntestgroup\nunrecognized";
var ss2 = "testpool\ntestgroup\ntestregion";
var ss3 = "testpool\ntestgroup\n";
var ss4 = "testpool\ntestgroup\nfoo";
var regex = /testpool\ntestgroup\n(?:unrecognized|testregion)/;

console.log(regex.test(ss1));
console.log(regex.test(ss2));
console.log(regex.test(ss3));
console.log(regex.test(ss4));
```

grbartel (dismissed):

integ/components/deadline/deadline_03_workerFleetHttp/test/deadline_03_workerFleetHttp.test.ts
kozlove-aws: Yes, your solution looks neatly.

ddneilson (approved):

ryyakobe (approved):

PR #252, opened 2020-11-25
fix(deadline): bad file path for Repository when using VersionQuery

### LINKED ISSUE #251: Bad filepath in Repository when using VersionQuery
There is an unresolved CDK token in the file path that the Repository generates for container-based connectors. The result is a REPO_URI being passed to the container that looks like:
```
             {
                "Name": "REPO_URI",
                "Value": "file:///opt/Thinkbox/DeadlineRepository-1.888154589708898e+289"
              },
```

Oddly, it doesn't seem to cause issues in some regions (us-west-2) but did in others (ca-central-1).

--- 

This is :bug: Bug Report

ddneilson:
Fixes: https://github.com/aws/aws-rfdk/issues/251

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/repository.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/repository.ts
@@ -47,6 +47,7 @@ import {
   RemovalPolicy,
   Stack,
   Tags,
+  Token,
 } from '@aws-cdk/core';
 import {
   CloudWatchAgent,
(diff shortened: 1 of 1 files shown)


horsmand (approved):
This looks fine to me, but I think we may want to re-evaluate if it makes sense to leave these `VersionQuery` fields like `majorVersion` as numbers when we need to jump through hoops any time we need to use them. It might make more sense to just make them strings.

jusiskin (approved):

PR #257, opened 2020-12-04
feat(deadline): configure worker log listener port

### LINKED ISSUE #190: Deadline Monitor unable to connect to workers to get logs
My team is having issues with getting access to worker logs when the workers are created using the Spot Event Plugin. They don't stream their logs to CloudWatch, and it looks like the local port (that Deadline Monitor would connect to in order to access logs) on each worker is randomized. That means that the Security Group for the workers has to open up all ports - so it isn't very secure.

Is there any chance of either making the port be static instead of random? Any other solution would also be acceptable - so long as it doesn't involve opening all of the ports.

### Environment

  - **CDK CLI Version      : 1.66.0
  - **CDK Framework Version: 1.66.0
  - **RFDK Version: 0.18
  - **Deadline Version:**
  - **Node.js Version:** 
  - **OS               :**
  - **Language (Version):** 

### Other

--- 

This is :bug: Bug Report

horsmand:
Fixes #190

Either a default port or a user supplied port will be configured for all
workers to listen on, for requests for their log stream. A helper method
is also provided to open up the port in the worker fleet's security
group to any provided IConnectable, like another security group or a CIDR.

### Testing
I followed these steps for testing:
1. Set up a render farm based on the All-in-AWS-Infrastructure-Basic TS example with these modifications:
  a. Added a Windows worker fleet added alongside the Linux fleet.
  b. Opened up the worker log listener ports in both fleets for the CIDR `10.0.0.0/24`.
1. Manually deployed another Windows host in my VPC.
1. Installed Deadline onto it and connected it to my RCS.
1. Opened Deadline Monitor and then connected to the logs for both of my workers.

----


packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
```diff
@@ -113,6 +117,13 @@ if [ ${#WORKER_POOLS[@]} -gt 0 ]; then
   "$DEADLINE_COMMAND" -SetPoolsForSlave $(IFS=, ; echo "${WORKER_NAMES[*]}") $(IFS=, ; echo "${WORKER_POOLS[*]}")
 fi
 
+# Setting listening port for the workers on this node (we cannot have multiple workers listening on the same port)
+port_offset=0
+for worker_name in "${WORKER_NAMES[@]}"; do
+  "$DEADLINE_COMMAND" -ExecuteScriptNoGui "$WORKER_LISTENING_PORT_CONFIG_SCRIPT" -n $worker_name -p $(($WORKER_LISTENING_PORT+$port_offset))
+  port_offset=$((port_offset+1))
```
horsmand, line 124: Looking back at this, I never added documentation to explain the port number would be incremented by one for any additional workers found on the host.

I'm not sure if it's better to add documentation for this, or just remove the feature. I think technically it's possible to support multiple workers on a host right now, but I believe in the future this won't work with auto-scaling.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -178,16 +186,23 @@ export interface WorkerInstanceConfigurationProps {
  *   environments.
  */
 export class WorkerInstanceConfiguration extends Construct {
+  private static readonly DEFAULT_WORKER_PORT = 56032;
```
ddneilson: Curious... randomly selected? Or is this a standard port for Deadline?

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -178,16 +186,23 @@ export interface WorkerInstanceConfigurationProps {
...
+
+  public readonly logListenerPort: number;
```
ddneilson: This should have a docstring.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
```diff
@@ -116,6 +117,13 @@ export interface WorkerSettings {
    * @default - Worker is not assigned to any region
    */
   readonly region?: string;
+
+  /**
+   * The port to configure the worker to listen on for sharing its logs.
+   *
+   * @default 56032
+   */
+  readonly logListenerPort?: number;
```
ddneilson: `logListenerPort` might not be the right name here. AFAIK, this is actually the general Deadline Worker Remote Command port.  It can be used for log streaming, but there are other uses for it as well.

packages/aws-rfdk/lib/deadline/lib/worker-fleet.ts
```diff
@@ -475,6 +498,13 @@ export class WorkerInstanceFleet extends WorkerInstanceFleetBase {
     this.fleet.addSecurityGroup(securityGroup);
   }
 
+  /**
+   * @inheritdoc
+   */
+  public allowLogListenerPort(other: IConnectable): void {
+    this.connections.allowFrom(other, this.logListeningPort, 'Worker log listening port');
```
ddneilson: We need to be careful with `allowFrom` vs `allowTo`.
Using `allowFrom` here means that any resource created in a stack that depends on the Worker's stack will not be able to deploy due to a cyclic dependency. Using `allowTo` (ex: `other.connections.allowTo(this, ...)`) allows `other` to exist in a stack that depends on the Worker's stack, but not to an `other` that is created before the Worker's stack.

I *suspect* that the `other` being in a stack that depends on the Worker stack is more likely than the other way, but I'm not sure. What do you think?

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
ddneilson: Hrm. This is tricky/complicated. The WorkerFleet doesn't currently know how many workers are running on the node, so it's not opening up ports with `port_offset > 0` in the security group.

We are also planning to be very opinionated about one worker per node for autoscaling, but maybe not everyone wants to use the autoscaling...

Some options, off the top of my head, are:
1. We assume no more than, say, 8 workers on a node and always open up a range of ports in the security group.
2. Have this only support single worker.
3. Some sort of way for the customer to tell us how many worker applications are running on the instance; though, that feels like it could balloon into needing to also configure that many workers to start.
4. other?

Thoughts?

ddneilson (changes requested):

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
horsmand: I know Josh has said there's a use case for multiple workers on a single host to share licenses, so if we can support it without too much work, I think we should. I like the idea of limiting it to 8 and opening up that range in the security group, that keeps it simple and we can always make improvements in the future if we find customers need more.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
horsmand: This was a random selection, from what I've seen Deadline always randomizes this port. I couldn't find any other common software using this port and it's in the dynamic port range defined by IANA.

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
horsmand: Updated

packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts
horsmand: Updated to just "listener port" and clarified in the comments.

packages/aws-rfdk/lib/deadline/lib/worker-fleet.ts
horsmand: I've implemented both so the customer can choose.

To test, I created an app with a worker stack and a stack with a windows instance I could grant access to the listener port. I was able to force dependencies using the ConstructNode's `addDependency()` and then use the wrong method to get the failures due to cyclic dependencies that we expected, and then use the correct method and have the instance get deployed with the proper port access. For instance, this failed:
```
windowsHost.node.addDependency(computeTier.workerFleet);
computeTier.workerFleet.allowListenerPortFrom(windowsHost.connections);
```
but this worked:
```
windowsHost.node.addDependency(computeTier.workerFleet);
computeTier.workerFleet.allowListenerPortTo(windowsHost.connections);
```

ddneilson (dismissed):

packages/aws-rfdk/lib/deadline/test/worker-configuration.test.ts
```diff
@@ -71,14 +70,35 @@ describe('Test WorkerInstanceConfiguration for Linux', () => {
...
     const userData = stack.resolve(instance.userData.render());
 
-    // THEN
+    // // THEN
+    expect(userData).toStrictEqual({
+      'Fn::Join': [
+        '',
+        [
+          '#!/bin/bash\nmkdir -p $(dirname \'/tmp/',
+          ...linuxConfigureWorkerScriptBoilerplate(`\' \'\' \'\' \'\' \'${Version.MINIMUM_SUPPORTED_DEADLINE_VERSION.toString()}\' 56032 /tmp/`),
```
kozlove-aws: Would it be better to use `DEFAULT_LISTENER_PORT ` instead of 56032?
It could make code more understandable.

packages/aws-rfdk/lib/deadline/scripts/bash/configureWorker.sh
```diff
@@ -113,6 +117,13 @@ if [ ${#WORKER_POOLS[@]} -gt 0 ]; then
   "$DEADLINE_COMMAND" -SetPoolsForSlave $(IFS=, ; echo "${WORKER_NAMES[*]}") $(IFS=, ; echo "${WORKER_POOLS[*]}")
 fi
 
+# Setting listening port for the workers on this node (we cannot have multiple workers listening on the same port)
+port_offset=0
+for worker_name in "${WORKER_NAMES[@]}"; do
+  "$DEADLINE_COMMAND" -ExecuteScriptNoGui "$WORKER_LISTENING_PORT_CONFIG_SCRIPT" -n $worker_name -p $(($WORKER_LISTENING_PORT+$port_offset))
```
kozlove-aws, line 123: What if one or several ports from the range is occupied by something? 
 Do we lost these workers?

horsmand: If a port that a worker uses to listen on is already occupied, this will set the port but the worker won't be able to access it. The customer will need to use the property on the WorkerConfiguration to set a new range that is free. I believe modifying the user data will cause the old worker to be terminated and a new host to be spun up, but this isn't easy to avoid. If the customer just uses Deadline command to change the port, they would also need to modify security groups and possibly the firewall if the worker is on Windows. All of this would cause drift between their CloudFormation templates and their deployment.

packages/aws-rfdk/lib/deadline/test/worker-configuration.test.ts
horsmand: Updated.

kozlove-aws (approved):

ddneilson (approved):

PR #260, opened 2020-12-08
chore(integ): add ability to run integration tests in parallel

kozlove-aws:
## Problem
Currently we only have possibility to run integration tests consistently that takes a lot of time

## Solution
Integration test will be running in separate processes.

## Testing
For testing were run integration tests.
Verified that all tests passed and result displayed properly.

----

### DIFF
--- a/integ/components/deadline/common/scripts/bash/deploy-utils.sh
+++ b/integ/components/deadline/common/scripts/bash/deploy-utils.sh
@@ -20,9 +20,13 @@ function deploy_component_stacks () {
   echo "Running $COMPONENT_NAME end-to-end test..."
 
   echo "Deploying test app for $COMPONENT_NAME test suite"
-  npx cdk deploy "*" --require-approval=never
+  if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
+    npx cdk deploy "*" --require-approval=never > "$INTEG_TEMP_DIR/${COMPONENT_NAME}_deploy.txt" 2>&1
+  else
+    npx cdk deploy "*" --require-approval=never
+  fi
   echo "Test app $COMPONENT_NAME deployed."
-
+  
   return 0
 }
 
@@ -32,7 +36,11 @@ function execute_component_test () {
   run_aws_interaction_hook
 
   echo "Running test suite $COMPONENT_NAME..."
-  yarn run test "$COMPONENT_NAME.test" --json --outputFile="./.e2etemp/$COMPONENT_NAME.json"
+  if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
+    yarn run test "$COMPONENT_NAME.test" --json --outputFile="$INTEG_TEMP_DIR/$COMPONENT_NAME.json" > "$INTEG_TEMP_DIR/${COMPONENT_NAME}.txt" 2>&1
+  else
+    yarn run test "$COMPONENT_NAME.test" --json --outputFile="$INTEG_TEMP_DIR/$COMPONENT_NAME.json"
+  fi
   echo "Test suite $COMPONENT_NAME complete."
 
   return 0
@@ -44,7 +52,11 @@ function destroy_component_stacks () {
   run_aws_interaction_hook
 
   echo "Destroying test app $COMPONENT_NAME..."
-  npx cdk destroy "*" -f
+  if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
+    npx cdk destroy "*" -f > "$INTEG_TEMP_DIR/${COMPONENT_NAME}_destroy.txt" 2>&1
+  else
+    npx cdk destroy "*" -f
+  fi
   rm -f "./cdk.context.json"
   rm -rf "./cdk.out"
   echo "Test app $COMPONENT_NAME destroyed."
--- a/integ/scripts/bash/rfdk-integ-e2e.sh
+++ b/integ/scripts/bash/rfdk-integ-e2e.sh
@@ -94,7 +94,7 @@ for COMPONENT in **/cdk.json; do
         # Excecute the e2e test in the component's scripts directory
         cd "$INTEG_ROOT/$COMPONENT_ROOT"
         if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
-            ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" --destroy-only &
+            (../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" --destroy-only) &
             export ${COMPONENT_NAME}_PID=$!
             COMPONENTS+=(${COMPONENT_NAME})
         else
@@ -118,6 +118,9 @@ if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
               fi
               if [ -f "$INTEG_TEMP_DIR/${COMPONENT_NAME}.txt" ]; then
                 cat "$INTEG_TEMP_DIR/${COMPONENT_NAME}.txt"
+              fi
+              if [ -f "$INTEG_TEMP_DIR/${COMPONENT_NAME}_destroy.txt" ]; then
+                cat "$INTEG_TEMP_DIR/${COMPONENT_NAME}_destroy.txt"
               fi
             fi
             export ${COMPONENT_NAME}_FINISH_TIME=$SECONDS
--- a/integ/scripts/bash/rfdk-integ-e2e.sh
+++ b/integ/scripts/bash/rfdk-integ-e2e.sh
@@ -37,6 +37,7 @@ fi
 
 # Create temp directory
 export INTEG_TEMP_DIR="$INTEG_ROOT/.e2etemp"
+rm -rf $INTEG_TEMP_DIR
 mkdir -p $INTEG_TEMP_DIR
 
 # Stage deadline from script
@@ -76,6 +77,7 @@ $BASH_SCRIPTS/deploy-infrastructure.sh || cleanup_on_failure
 export INFRASTRUCTURE_DEPLOY_FINISH_TIME=$SECONDS
 
 # Pull the top level directory for each cdk app in the components directory
+COMPONENTS=()
 for COMPONENT in **/cdk.json; do
     # In case the yarn install was done inside this integ package, there are some example cdk.json files in the aws-cdk
     # package we want to avoid.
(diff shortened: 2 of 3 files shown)


integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -90,11 +91,29 @@ for COMPONENT in **/cdk.json; do
     export ${COMPONENT_NAME}_START_TIME=$SECONDS
     if [[ "$COMPONENT_NAME" != _* ]]; then
         # Excecute the e2e test in the component's scripts directory
-        cd "$INTEG_ROOT/$COMPONENT_ROOT" && ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || cleanup_on_failure
+        cd "$INTEG_ROOT/$COMPONENT_ROOT"
+        ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || cleanup_on_failure &
```
ddneilson: Have you tested this change with a failing integration test? ex: Make a modification to one of the jest tests that would cause a failure, and make sure that the behavior is as expected.

horsmand (commented):
Has this been tested by running on CodeBuild? How will we determine what happened if our integration tests fail in the canaries or the bump script?

integ/scripts/bash/rfdk-integ-e2e.sh
kozlove-aws: I have tested it and I still can see errors in console like it was before:
```
 FAIL  components/deadline/deadline_02_renderQueue/test/deadline_02_renderQueue.test.ts (53.564 s)
  ● Deadline RenderQueue tests (HTTP mode) › deadlinecommand tests › RQ-1-4: Farm can accept sample Deadline job
    expect(received).toEqual(expected) // deep equality

    Expected: 10
    Received: 0

      228 |       return awaitSsmCommand(bastionId, params).then( response => {
      229 |         var responseCode = response.responseCode;
    > 230 |         expect(responseCode).toEqual(10);
          |                              ^
      231 |       });
      232 |     });
      233 |   });

      at components/deadline/deadline_02_renderQueue/test/deadline_02_renderQueue.test.ts:230:30
```
And result as expected:
```
Results for test component deadline_02_renderQueue:
  -Tests ran: 8
  -Tests passed: 6
  -Tests failed: 2
  -Deploy runtime:     24m 16s
  -Test suite runtime: 0m 55s
  -Cleanup runtime:    1m 43s
```

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -90,11 +91,35 @@ for COMPONENT in **/cdk.json; do
     export ${COMPONENT_NAME}_START_TIME=$SECONDS
     if [[ "$COMPONENT_NAME" != _* ]]; then
         # Excecute the e2e test in the component's scripts directory
-        cd "$INTEG_ROOT/$COMPONENT_ROOT" && ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || cleanup_on_failure
+        cd "$INTEG_ROOT/$COMPONENT_ROOT"
+        if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
+            ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || cleanup_on_failure &
```
jusiskin: By using the `&` operator, bash will run the entire pipeline of commands:

```sh
../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || cleanup_on_failure
```

in a sub-shell ([source](https://www.gnu.org/software/bash/manual/html_node/Lists.html)). This means that if any of the parallel pipelines errors, the cleanup will be triggered. Digging into the `cleanup_on_failure` code path, it calls:

```
yarn run tear-down
```

which runs [`integ/scripts/bash/tear-down.sh`](https://github.com/aws/aws-rfdk/blob/b9ad4180d0704b7c0bc9c4982ee9ff90572704d5/integ/scripts/bash/tear-down.sh). This will [loop over all of the component stacks and destroy them](https://github.com/aws/aws-rfdk/blob/b9ad4180d0704b7c0bc9c4982ee9ff90572704d5/integ/scripts/bash/tear-down.sh#L33-L46). It will also [destroy the test infrastructure stack](https://github.com/aws/aws-rfdk/blob/b9ad4180d0704b7c0bc9c4982ee9ff90572704d5/integ/scripts/bash/tear-down.sh#L50).

This means if there there are regressions in one component, it will abort any other running tests and clean everything up. As you're aware working with these tests, they take a long time to complete. It would be better if we could rearrange things to allow all tests to complete before tearing everything down all regressions can be identified in a single run of the integration tests. What do you think?

jusiskin: I've checked out this branch and ran the tests. It appears that all parallel output is interleaved to the standard output and standard error streams. This can make it difficult to review the results and figure out what is going on.

Do you have any ideas on how we can make the output more readable?

jusiskin (changes requested):
Great work parallelizing this! It appears to function nicely in the working case, but I believe it need some enhancements to make it easier to troubleshoot errors.

integ/scripts/bash/rfdk-integ-e2e.sh
horsmand: Yeah this interleaved output was a problem I was afraid of and had suggested maybe each test suite could write to its own file, but that gets complicated when we are running in CodeBuild. I don't know how easy it would be to write those to CloudWatch.

jusiskin: Yeah, at the very least, we could write the output to files and the `cat` them after all tests have complete if there were any deployment failures. Ideally though, we'd also have some form of status being output as the tests are running to be able to observe the progress of each component test running (deploying/testing/destroying/complete/failed).


```sh
$ yarn run e2e
2021-02-03 16:39:00 [DL] deploying...
2021-02-03 16:39:00 [RQ] deploying...
2021-02-03 16:39:00 [WF] deploying...
2021-02-03 16:39:00 [WFS] deploying...
2021-02-03 17:04:00 [DL] testing...
2021-02-03 17:21:00 [RQ] testing...
2021-02-03 20:00:00 [WF] testing...
2021-02-03 20:03:00 [DL] destroying...
...
```

If a deployment failed for one of those components, we could output the deployment log file after all tests are completed (so it isn't interleaved with the above output).

What do you think?

horsmand: I like it. Much simpler than having to work with CloudWatch.

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -90,11 +92,47 @@ for COMPONENT in **/cdk.json; do
...
+            if ps -p "$PID" > /dev/null; then
+              ACTIVE_COMPONENTS+=(${COMPONENT_NAME})
+            else
+              echo "Test app $COMPONENT_NAME finished."
+              if [ -f "$INTEG_TEMP_DIR/${COMPONENT_NAME}_deploy.txt" ]; then
+                cat "$INTEG_TEMP_DIR/${COMPONENT_NAME}_deploy.txt"
+              fi
+              if [ -f "$INTEG_TEMP_DIR/${COMPONENT_NAME}.txt" ]; then
+                cat "$INTEG_TEMP_DIR/${COMPONENT_NAME}.txt"
+              fi
```
jusiskin, line 121: This outputs stdout from this command in `integ/components/deadline/common/scripts/bash/deploy-utils.sh`:

```sh
yarn run test "$COMPONENT_NAME.test" --json --outputFile="$INTEG_TEMP_DIR/$COMPONENT_NAME.json" > "$INTEG_TEMP_DIR/${COMPONENT_NAME}.txt" 2>&1
```

When I tried running this code, it looks like there is no output that shows the test results. I believe jest is writing the JSON test report to `$INTEG_TEMP_DIR/$COMPONENT_NAME.json` and the original code would output this from `integ/scripts/bash/report-test-results.sh`. Strangely, I did not see the test results after running this myself. Maybe I'm missing something.

When you ran this, were you able to see the test result output?

integ/components/deadline/common/scripts/bash/deploy-utils.sh
```diff
@@ -44,7 +52,11 @@ function destroy_component_stacks () {
   run_aws_interaction_hook
 
   echo "Destroying test app $COMPONENT_NAME..."
-  npx cdk destroy "*" -f
+  if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
+    npx cdk destroy "*" -f > "$INTEG_TEMP_DIR/${COMPONENT_NAME}_destroy.txt" 2>&1
```
jusiskin, line 56: I don't see these `*_destroy.txt` log file being used anywhere. Can we output them if something goes wrong during the destroy?

jusiskin (changes requested):
This is looking good.

The main thing remaining is that I am unable to see the test results at the end.

integ/scripts/bash/rfdk-integ-e2e.sh
kozlove-aws: This is strange. I am looking on folder when I ran these tests locally and I can see 4 files for every test.
(`deadline_01_repository_deploy.txt, deadline_01_repository_destroy.txt, deadline_01_repository.json, deadline_01_repository.txt`) 
Have you run your test with `export RUN_TESTS_IN_PARALLEL=true`?

integ/components/deadline/common/scripts/bash/deploy-utils.sh
kozlove-aws: I thought destroying log will not be very useful, but on other hand there is no useless information. Added.

integ/components/deadline/common/scripts/bash/deploy-utils.sh
```diff
@@ -20,9 +20,13 @@ function deploy_component_stacks () {
   echo "Running $COMPONENT_NAME end-to-end test..."
 
   echo "Deploying test app for $COMPONENT_NAME test suite"
-  npx cdk deploy "*" --require-approval=never
+  if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
+    npx cdk deploy "*" --require-approval=never > "$INTEG_TEMP_DIR/${COMPONENT_NAME}_deploy.txt" 2>&1
```
jericht, line 24: I think it would be valuable to have an option to specify another directory to store these log files instead of the integ temp directory. Maybe having an optional environment variable like `PARALLEL_RESULTS_DIRECTORY`, with the default being the integ temp directory.

This makes it so these files don't get cleaned up by the `cleanup.sh` script that gets run after the tests, making it easier to look at the logs for each test rather than having one large file (or lots of output on the command line). What do you think?

jericht (changes requested):
Just one suggestion for the output log files, otherwise this looks good to me

integ/components/deadline/common/scripts/bash/deploy-utils.sh
kozlove-aws: As I know the only file that appear in this folder is output from test in json format.
I can't see any reason why others files should be put in different folder. As for me they should be beside this output file.

jericht: My idea was to have an option to keep these new `.txt` files as artifacts after a run of the integration tests to make it easier to inspect the results after the test. I did this myself when running the tests in parallel and thought it would be nice to have as an option for our tests.

This is more of a QOL option and isn't strictly necessary, so I won't hold up the PR for this.

jericht (dismissed):

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -90,11 +92,50 @@ for COMPONENT in **/cdk.json; do
     export ${COMPONENT_NAME}_START_TIME=$SECONDS
     if [[ "$COMPONENT_NAME" != _* ]]; then
         # Excecute the e2e test in the component's scripts directory
-        cd "$INTEG_ROOT/$COMPONENT_ROOT" && ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || cleanup_on_failure
+        cd "$INTEG_ROOT/$COMPONENT_ROOT"
+        if [ "${RUN_TESTS_IN_PARALLEL}" = true ]; then
+            ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" --destroy-only &
```
yashda: looks like we need braces here:

```suggestion
            (../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" || ../common/scripts/bash/component_e2e.sh "$COMPONENT_NAME" --destroy-only) &
```

If not, the console will wait for the former command to finish and if it fails, it will run the later one in background.

integ/test-config.sh
```diff
@@ -5,7 +5,7 @@
 
 # Accept SSPL license to install MongoDB
 #   - If not set to `true` here, the Repository test component will fail immediately on execution
-export USER_ACCEPTS_SSPL_FOR_RFDK_TESTS=false
+export USER_ACCEPTS_SSPL_FOR_RFDK_TESTS=true
```
yashda: Is this intentional? AFAIK, we leave the default value to false so that a user running it has to explicitly set the value to true.

yashda (changes requested):

integ/scripts/bash/rfdk-integ-e2e.sh
kozlove-aws: It works like I need (execute test on background and destroy when it fails). I believe `&` applied to all command.
But I get your point and added braces for visibility.

integ/test-config.sh
kozlove-aws: It was changed by mistake. Fixed.

integ/scripts/bash/rfdk-integ-e2e.sh
yashda: Putting the braces around actually changes the behavior. 
Without braces, the first command runs in foreground and the second one runs in background (if the first fails). 
With braces, both, the first and the second command will run in background.

I tried by running following command and the sleep command ran in foreground blocking my console for 5s:
`(sleep 5 && echo 'task finished' && exit 1) || (echo 'task failed') &`

Whereas, running following did not block my console at all:
`((sleep 5 && echo 'task finished' && exit 1) || (echo 'task failed')) &`

yashda (approved):
Changes looks good to me. Thanks Eugene!

integ/scripts/bash/rfdk-integ-e2e.sh
kozlove-aws: Hm. nice observing. 
But if you try to run it from shell script behavior is the same.

jericht (approved):

integ/scripts/bash/rfdk-integ-e2e.sh
yashda: Looks like `z-shell` & 'shell/bash' have different behaviors, hence it was blocking for me!

integ/scripts/bash/rfdk-integ-e2e.sh
jusiskin: I've re-ran this using your latest commit and no longer see the issue. Resolving

jusiskin (approved):

PR #264, opened 2020-12-08
feat(deadline): allow providing subnets for RenderQueue's ALB

### LINKED ISSUE #263: Allow subnet selection for RenderQueue's ALB
The ask is for the ability to control which subnets the RenderQueue will deploy its ALB(s) into.

### Use Case

When deploying into a Local Zone (ex: the LA Local Zone), the Local Zone is actually a part of a greater region (us-west-2 in the case of the LA zone). An ALB cannot be deployed into subnets that are both in the region and in the local zone. So, to support local zones we need the ability for the user to select which subnets the ALB will be created within (either entirely in the LZ or entirely in the regular Region AZs).

### Proposed Solution

Adding a `vpcAlbSubnets` property to the RenderQueue props.

### Other

N/A

* [X] :wave: I may be able to implement this feature request

---

ddneilson:
Implements: https://github.com/aws/aws-rfdk/issues/263

----


packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -334,6 +334,7 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
 
     const loadBalancer = new ApplicationLoadBalancer(this, 'LB', {
       vpc: this.cluster.vpc,
+      vpcSubnets: props.vpcAlbSubnets ?? { subnetType: SubnetType.PRIVATE },
```
horsmand: Do we need to add the `?? { subnetType: SubnetType.PRIVATE }` part here? This matches the default behaviour, but why not let the constructor handle this defaulting itself?

ddneilson: Don't *need* it. I just like being explicit.

horsmand (dismissed):

horsmand (dismissed):

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -243,6 +243,13 @@ export interface RenderQueueProps {
    */
   readonly vpcSubnets?: SubnetSelection;
 
+  /**
+   * The subnets into which to place the Application Load Balancer that is deployed.
+   *
+   * @default - One Private subnet from each AZ.
+   */
+  readonly vpcAlbSubnets?: SubnetSelection;
```
jusiskin: nit: does it make more sense to name this `vpcSubnetsAlb`? Unfortunately I had created the original `vpcSubnets` which dictates the placement of the ASG instances without thinking about the ALB. The original `vpcSubnets` came from the convention used in the CDK's standard AWS construct library:

* [`Instance` construct's `vpcSubnets` prop](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.Instance.html#vpcsubnets)
* [`AutoScalingGroup` construct's `vpcSubnets` prop](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-autoscaling.AutoScalingGroup.html#vpcsubnets)

To continue with that convention as best as possible, I suggest we stick with `vpcSubnets` prefix and append `Alb` as a suffix. That way those using code-completion will see both show up and can pick/explore accordingly. What do you think though?

jusiskin (changes requested):
Just a minor prop rename suggestion - otherwise looks good

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
ddneilson: That seems reasonable to me.

horsmand (approved):

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
jusiskin: Looks good - thanks

jusiskin (approved):

PR #266, opened 2020-12-09
feat(core): ssm policy helper

### LINKED ISSUE #265: Add Helper Method for Configuring SSM
It's very common to require access to a host in a render farm for troubleshooting. A simple way of connecting to a host is the [AWS Systems Manager Session Manager](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager.html). It would be nice to have a helper method that applies the required permissions for using Session Manager.

### Use Case

An example of where this can be used is on a worker fleet that is having issues. I would like to be able to pass the worker fleet into the helper method and then be able to go into the AWS console and connect to the host using the Session Manager to be able to access the host from the browser without having to configure ssh keys and permissions.

### Proposed Solution

A helper method that applies this policy:
```
new PolicyStatement(
    actions: [
        'ssmmessages:CreateControlChannel',
        'ssmmessages:CreateDataChannel',
        'ssmmessages:OpenControlChannel',
        'ssmmessages:OpenDataChannel',
        'ssm:UpdateInstanceInformation',
    ],
    resources: ['*']
)
```

---

horsmand:
Fixes #265

This was tested by deploying the basic TS and Python examples and then using the AWS console to connect to the worker fleet instances they created with the Session Manager.

----

### DIFF
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
@@ -32,8 +32,16 @@ These instructions assume that your working directory is `examples/deadline/All-
     popd
     pip install ../../../../dist/python/aws-rfdk-<version>.tar.gz
     ```
-4.  Change the value in the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker.
+4.  Change the value in the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker. You can use the following AWS CLI query to find AMI ID's:
+    ```bash
+    aws --region <region> ec2 describe-images \
+    --owners 357466774442 \
+    --filters "Name=name,Values=*Worker*" "Name=name,Values=*<version>*" \
+    --query 'Images[*].[ImageId, Name]' \
+    --output text
+    ```
 
+    And enter it into this section of `package/config.py`:
     ```python
     # For example, in the us-west-2 region
     self.deadline_client_linux_ami_map: Mapping[str, str] = {
@@ -104,10 +112,7 @@ These instructions assume that your working directory is `examples/deadline/All-
     # Set this value to the version of AWS Thinkbox Deadline you'd like to deploy to your farm. Deadline 10.1.9 and up are supported.
     RFDK_DEADLINE_VERSION=<version_of_deadline>
 
-    npx --package=aws-rfdk@${RFDK_VERSION} stage-deadline \
-        --deadlineInstallerURI s3://thinkbox-installers/Deadline/${RFDK_DEADLINE_VERSION}/Linux/DeadlineClient-${RFDK_DEADLINE_VERSION}-linux-x64-installer.run \
-        --dockerRecipesURI s3://thinkbox-installers/DeadlineDocker/${RFDK_DEADLINE_VERSION}/DeadlineDocker-${RFDK_DEADLINE_VERSION}.tar.gz \
-        --output stage
+    npx --package=aws-rfdk@${RFDK_VERSION} stage-deadline ${RFDK_DEADLINE_VERSION} --output stage
     ```
 12. Deploy all the stacks in the sample app:
 
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
@@ -24,6 +24,7 @@ from aws_cdk.aws_s3_assets import (
 
 from aws_rfdk import (
     HealthMonitor,
+    SessionManagerHelper,
 )
 from aws_rfdk.deadline import (
     InstanceUserDataProvider,
@@ -116,6 +117,11 @@ class ComputeTier(Stack):
             key_name=props.key_pair_name,
             user_data_provider=UserDataProvider(self, 'UserDataProvider')
         )
+
+        # This is an optional feature that will set up your EC2 instances to be enabled for use with
+        # the Session Manager. These worker fleet instances aren't available through a public subnet,
+        # so connecting to them directly through SSH isn't easy.
+        SessionManagerHelper.grant_permissions_to(self.worker_fleet)
 
         if props.usage_based_licensing and props.licenses:
             props.usage_based_licensing.grant_port_access(self.worker_fleet, props.licenses)
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/service-tier.ts
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/service-tier.ts
@@ -22,7 +22,6 @@ import {
 } from 'aws-rfdk';
 import {
   DatabaseConnection,
-  IRenderQueue,
   RenderQueue,
   Repository,
   Stage,
@@ -34,6 +33,7 @@ import {
   Secret,
 } from '@aws-cdk/aws-secretsmanager';
 import { Duration } from '@aws-cdk/core';
+import { SessionManagerHelper } from 'aws-rfdk/lib/core';
 
 /**
  * Properties for {@link ServiceTier}.
@@ -90,7 +90,7 @@ export class ServiceTier extends cdk.Stack {
   /**
    * The render queue.
    */
-  public readonly renderQueue: IRenderQueue;
+  public readonly renderQueue: RenderQueue;
 
   /**
    * The UBL licensing construct. (License Forwarder)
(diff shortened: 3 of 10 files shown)


examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
```diff
@@ -117,6 +118,12 @@ def __init__(self, scope: Construct, stack_id: str, *, props: ComputeTierProps,
             user_data_provider=UserDataProvider(self, 'UserDataProvider')
         )
 
+        # This is an example of how you can set up a fleet to be enabled for use with the Session
```
ddneilson: Maybe add a note that for this to work they either have to be using an AMI that is based on Amazon Linux 2 or an Amazon-provided Windows Server, or have installed the SSM-Agent and set it to run on startup?

examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
```diff
@@ -32,8 +32,16 @@ These instructions assume that your working directory is `examples/deadline/All-
     popd
     pip install ../../../../dist/python/aws-rfdk-<version>.tar.gz
     ```
-4.  Change the value in the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker.
+4.  Change the value in the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker. You can use the follow AWS CLI query to find AMI ID's:
+    ```bash
```
ddneilson, line 36: Nice addition!

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
```diff
@@ -117,6 +118,12 @@ def __init__(self, scope: Construct, stack_id: str, *, props: ComputeTierProps,
...
+        # Manager from the AWS Console. After the fleet deploys, you should be able to go into the
+        # EC2 Instances Console, right click on your worker instance, select "Connect", and then
+        # select the Session Manager and open a shell session in the browser.
+        SessionManagerHelper.grant_permissions_to(self.worker_fleet);
```
ddneilson: Might be worth adding this to the RenderQueue & UBL as well.

ddneilson (changes requested):

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
horsmand: I actually tried it on the render queue and it didn't work. I checked the role's policy and it had the additional actions for SSM, so I figured it was something to do with the way we set these instances up. I can look into it a little more.

ddneilson: I think I know why this might not be working. The `grantPrincipal` on the RenderQueue & UsageBasedLicensing constructs are the Roles for the task; whereas the SSM permissions need to be added to the Role for the instance on which the ECS tasks are running.

ddneilson: So, that might be a point in favour of having a method on the `RenderQueue` & `UsageBasedLicensing` constructs themselves that adds the session manager permissions to the ECS instance; the method can use this `SessionManagerHelper` under the hood.

ddneilson: Alternatively, `SessionManagerHelper.grant_permissions_to(self.render_queue.asg)` ( ref: https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.RenderQueue.html#asg )

But, we'd want to be sure to draw attention to that in the example as the correct way to do it.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
horsmand: This seemed like a better fit for the class's docstring, so I added it there.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
horsmand: I swear I tried adding it to the render queue's ASG when I tried before, but I clearly did something wrong as your suggestions have worked for me. I've added configuring the render queue and UBL constructs to the TS and Python basic examples.

I started adding a specific method to do this configuration to the render queue, but it felt cleaner to just show in the example that it should be applied to the ASG property.

packages/aws-rfdk/lib/core/lib/session-manager-helper.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * This is a helper class meant to make it easier to use the AWS Systems Manager Session Manager
+ * with any EC2 Instances or AutoScalingGroups. Once enabled, the Session Manager can be used to
+ * connect to an EC2 Instance through the AWS Console and open a shell session in the browser.
+ *
+ * Note that in order for the Session Manager to work, you will need an AMI that has the SSM-Agent
+ * installed and set to run at startup. The Amazon Linux 2 and Amazon provided Windows Server AMI's
+ * have this configured by default.
+ *
+ * More details about the AWS Systems Manager Session Manager ca be found here:
```
ddneilson:
```suggestion
 * More details about the AWS Systems Manager Session Manager can be found here:
```

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
ddneilson: Perfect. Thanks David!

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/compute_tier.py
ddneilson: I agree.

packages/aws-rfdk/lib/core/lib/session-manager-helper.ts
horsmand: Fixed.

ddneilson (dismissed):

examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
```diff
@@ -32,8 +32,16 @@ These instructions assume that your working directory is `examples/deadline/All-
     popd
     pip install ../../../../dist/python/aws-rfdk-<version>.tar.gz
     ```
-4.  Change the value in the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker.
+4.  Change the value in the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker. You can use the follow AWS CLI query to find AMI ID's:
```
ryyakobe: Nit: should it be 
```suggestion
You can use the following AWS CLI query to find AMI ID's:
```
I might be wrong :) No need to fix it.

examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/service-tier.ts
```diff
@@ -111,8 +111,10 @@ export class ServiceTier extends cdk.Stack {
   constructor(scope: cdk.Construct, id: string, props: ServiceTierProps) {
     super(scope, id, props);
 
-    // Bastion instance for convenience (e.g. SSH into RenderQueue and WorkerFleet instances)
-    // Not a critical component of the farm, so this can be safely removed
+    // Bastion instance for convenience (e.g. SSH into RenderQueue and WorkerFleet instances).
+    // Not a critical component of the farm, so this can be safely removed. An alternative way
+    // to access your hosts is also provided by the Session Manager, which is also configured
+    // later in this example.
```
ryyakobe, line 117: Can we add the same comment to the python example?

examples/deadline/All-In-AWS-Infrastructure-Basic/python/README.md
horsmand: You are correct. I've fixed it.

examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/service-tier.ts
horsmand: Added.

ryyakobe (approved):
Great work!

ddneilson (approved):

PR #289, opened 2021-01-07
chore(deps): bumping cdk from 1.78.0 to 1.83.0

ryyakobe:
I deployed the all in AWS infrastructure basic app to confirm that everything is working.

Besides the CDK version bump, this includes:

- A change in `usage-based-licensing.ts` caused by the change in CDK https://github.com/aws/aws-cdk/pull/12223 . Our `UsageBasedLicensing` construct uses `desiredCount` not only for `Ec2Service`, but also for `ASG`. As we want these values to be the same, we won't be removing the `DesiredCount` similar to CDK, but will keep the same behavior as before, defaulting to 1. We do this, because if we don't set `DesiredCount` explicitly for the `Service`, it will use the current task number, but we won't be able to get this number for ASG from the service.

----


horsmand (approved):

packages/aws-rfdk/lib/deadline/lib/usage-based-licensing.ts
```diff
@@ -564,7 +564,7 @@ export class UsageBasedLicensing extends Construct implements IGrantable {
     this.service = new Ec2Service(this, 'Service', {
       cluster: this.cluster,
       taskDefinition,
-      desiredCount: props.desiredCount,
+      desiredCount: props.desiredCount ?? 1,
```
ddneilson, line 567: Remind me... what was the CDK change that required this change?

edit: n/m, I found it -- https://github.com/aws/aws-cdk/pull/12223

ddneilson (approved):

packages/aws-rfdk/lib/deadline/lib/usage-based-licensing.ts
ryyakobe: The link and short description is in the description of this PR. There were 2 options available:
1. In RFDK we just default to size 1 if `desiredCount` is not set explicitly. (what I did in this PR)
2. Do not set any default in RFDK for `Ec2Service`. In this case CDK will not add `DesiredCount` property to the CloudFormation template. But then it's not clear what to use for `minCapacity` and `maxCapacity` in ASG here https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/deadline/lib/usage-based-licensing.ts#L508, that's why I decided to go with a simpler option 1.

PR #296, opened 2021-01-22
fix(core): windows cloudwatch agent install script

### LINKED ISSUE #295: Windows worker fleets failing to start
Trying to deploy a worker fleet (or new instance into an existing worker fleet) that uses any Windows AMI is failing due to a bug in a script that we execute on the host as part of its initialization process (in the UserData). The nature of the bug means that it affects all versions of RFDK in production, and all worker fleets that try to deploy using an AMI with a Windows operating system.

The script that is failing attempts to install and configure the CloudWatch agent onto the instance. Because this script fails, the script that is supposed to configure the Deadline worker to connect to the render queue and start it never gets executed and the health check fails, causing the CDK deployment to fail and roll back.

Since the failure prevents CloudWatch from setting up properly, no logs are uploaded to CloudWatch and viewable from the AWS Console, and the host gets terminated.

Log statement seen that signals we're falling back to the latest version of the CloudWatch agent rather than the version we try to pin to: https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1#L26
Error message that was observed: https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1#L52

### Environment

  - **CDK CLI Version: all**
  - **CDK Framework Version: all**
  - **RFDK Version: all**
  - **Deadline Version: all**

--- 

This is :bug: Bug Report

horsmand:
Fixes #295

This fixes a bug that is preventing the deployment of any worker
instance that is using an AMI with a Windows OS.

To test this fix, I modified the All-In-AWS-Infrastructure-Basic example app to deploy a Windows worker fleet and made sure it could properly deploy.

----

### DIFF
--- a/packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
+++ b/packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
@@ -16,62 +16,105 @@ param (
 
 $ErrorActionPreference = "Stop"
 
-$file = "$env:temp\amazon-cloudwatch-agent.msi"
+Write-Output "Starting CloudWatch installation and configuration script."
 
-# We hardcode the version number here since the verification below is tightly coupled with this specific version
+$is_cloudwatch_installed = $False
 try {
-  Read-S3Object -BucketName amazoncloudwatch-agent -Key windows/amd64/1.242486.0/amazon-cloudwatch-agent.msi -File $file -Region us-east-1
+  # If this command doesn't throw an error, we have the CloudWatch agent installed already
+  $status = & $Env:ProgramFiles\Amazon\AmazonCloudWatchAgent\amazon-cloudwatch-agent-ctl.ps1 -m ec2 -a status
+  $is_cloudwatch_installed = $True
+  Write-Output "Found CloudWatch agent already installed, skipping installation."
 } catch {
-  # Fallback to the latest version (this is the case when the above version is currently "latest")
-  Write-Output "Initial attempt to download Amazon CloudWatch agent failed. Falling back to the latest version."
+  Write-Output "CloudWatch agent is not already installed, proceeding with installation."
+}
+
+if (-Not $is_cloudwatch_installed) {
+  $cwa_installer = "$env:temp\amazon-cloudwatch-agent.msi"
   try {
-    Read-S3Object -BucketName amazoncloudwatch-agent -Key windows/amd64/latest/amazon-cloudwatch-agent.msi -File $file -Region us-east-1
+    Read-S3Object -BucketName amazoncloudwatch-agent -Key windows/amd64/latest/amazon-cloudwatch-agent.msi -File $cwa_installer -Region us-east-1
   } catch {
-    Write-Output "Failed to download CloudWatchAgent installer."
+    Write-Output "Failed to download CloudWatch agent installer."
     Exit 1
   }
+
+  $cwa_installer_sig = "$env:temp\amazon-cloudwatch-agent.msi.sig"
+  try {
+    Read-S3Object -BucketName amazoncloudwatch-agent -Key windows/amd64/latest/amazon-cloudwatch-agent.msi.sig -File $cwa_installer_sig -Region us-east-1
+  } catch {
+    Write-Output "Failed to download CloudWatch agent installer signature file."
+    Exit 1
+  }
+
+  if (-Not $s) {
+    $gpg_keyring = "$env:temp\keyring.gpg"
+
+    # Download GPG
+    $gpg_installer = "$env:temp\gnupg-w32-2.2.27_20210111.exe"
+    wget https://gnupg.org/ftp/gcrypt/binary/gnupg-w32-2.2.27_20210111.exe -OutFile $gpg_installer
+
+    # Verify GPG
+    $gpg_sig = Get-AuthenticodeSignature $gpg_installer
+    $status = $gpg_sig | Select-Object -ExpandProperty 'Status'
+    if ( $status -ne 'Valid' ) {
+      Write-Output "GPG installer does not have a valid signature."
+      Exit 1
+    }
+    $sha256_expected = '5D89E239790822711EAE2899467A764879D21440AB68E9413452FA96CEDEBA50'
+    $sha256 = Get-FileHash $gpg_installer -Algorithm SHA256
+    if ( $sha256 -inotmatch $sha256_expected) {
+      Write-Output "GPG failed checksum verification. Expected sha256 to equal $sha256_expected but got:"
+      Write-Output $sha256
+      Exit 1
+    }
+
+    # Install GPG
+    Start-Process -Wait -FilePath $gpg_installer -ArgumentList "/S /v/qn" -PassThru
+
+    # Refresh the PATH so gpg is available
+    $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine")
+
+    # Download Amazon's public key and import it to GPG's keyring
+    $cloudwatch_pub_key = "$env:temp\amazon-cloudwatch-agent.gpg"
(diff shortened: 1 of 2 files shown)


packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
```diff
@@ -16,62 +16,90 @@ param (
...
+  wget https://gnupg.org/ftp/gcrypt/binary/gnupg-w32-2.2.27_20210111.exe -OutFile $gpg_installer
+
+  # Verify GPG
+  $gpg_sig = Get-AuthenticodeSignature $gpg_installer
+  $status = $gpg_sig | Select-Object -ExpandProperty 'Status'
   if ( $status -ne 'Valid' ) {
-    Write-Output "CloudWatchAgent installer does not have a valid signature."
+    Write-Output "GPG installer does not have a valid signature."
     Exit 1
   }
```
horsmand: Not sure if we need to check this since we don't check any other parts of the Authenticode signature. We could check the certificates like we did in the [old version](https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1#L50), but I think just checking the sha1 checksum should be enough.

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
```diff
@@ -16,62 +16,90 @@ param (
...
-  #  - Thumbprint: 614D271D9102E30169822487FDE5DE00A352B01D
-  $ca = $sig | Select-Object -ExpandProperty 'TimeStamperCertificate'
-  $ca_thumbprint = $ca | Select-Object -ExpandProperty 'Thumbprint'
-  if ( $ca_thumbprint -ne '614D271D9102E30169822487FDE5DE00A352B01D' ) {
-    Write-Output "CA Thumbprint failed verification. Has the CA certificate been changed? If so, then please submit a PR to fix this."
-    Write-Output $( $ca | Format-List )
+  # Download Amazon's public key and import it to GPG's keyring
+  $cloudwatch_pub_key = "$env:temp\amazon-cloudwatch-agent.gpg"
+  Read-S3Object -BucketName amazoncloudwatch-agent -Key assets/amazon-cloudwatch-agent.gpg -File $cloudwatch_pub_key -Region us-east-1
+  gpg --import $cloudwatch_pub_key
```
ddneilson: Mind modifying this to use a temporary keyring instead of mutating the default keyring?
See the `--no-default-keyring` option usage in: https://github.com/aws/aws-rfdk/blob/b4f1d15ebe494f70b881e572a5d03e3fc46a3d95/packages/aws-rfdk/lib/core/scripts/bash/configureCloudWatchAgent.sh#L52

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
```diff
@@ -16,62 +16,90 @@ param (
...
+  # Verify GPG
+  $gpg_sig = Get-AuthenticodeSignature $gpg_installer
+  $status = $gpg_sig | Select-Object -ExpandProperty 'Status'
   if ( $status -ne 'Valid' ) {
-    Write-Output "CloudWatchAgent installer does not have a valid signature."
+    Write-Output "GPG installer does not have a valid signature."
     Exit 1
   }
+  # Checksum information found at https://gnupg.org/download/integrity_check.html
+  $sha1 = Get-FileHash $gpg_installer -Algorithm SHA1
```
ddneilson: sha1 is considered to be insecure due to vulnerability to collision attacks; sha256 is more secure.

Mind manually calculating the sha256 of an installer that you've verified, and putting a sha256 check in here instead?

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
horsmand: Updated.

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
horsmand: Updated.

ddneilson (dismissed):
Great work. Thanks David!

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
```diff
@@ -16,62 +16,104 @@ param (
...
+    Read-S3Object -BucketName amazoncloudwatch-agent -Key assets/amazon-cloudwatch-agent.gpg -File $cloudwatch_pub_key -Region us-east-1
+    gpg --no-default-keyring --keyring $gpg_keyring --import $cloudwatch_pub_key
+
+    # Verify that the imported key has the correct fingerprint
+    $fingerprint = '937616F3450B7D806CBD9725D58167303B789C72'
+    $keys = gpg --no-default-keyring --keyring $gpg_keyring -k
+    $keys = ($keys | Out-String).Trim() -replace "\r", "" -replace "\n", "" -replace "\[", "" -replace "\]", ""
+
+    if ($keys -inotlike "*$fingerprint*") {
+      Write-Output "Expect CloudWatch agent's public key to equal 937616F3450B7D806CBD9725D58167303B789C72 but got:"
```
jericht: nit: Expect --> Expected.

Also, should we reference the the `$fingerprint` variable in the string instead of duplicating it? I believe Powershell supports variable expansion in strings like this:
```ps
Write-Output "Expect CloudWatch agent's public key to equal ${fingerprint} but got:"
```

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
```diff
@@ -16,62 +16,104 @@ param (
...
+
+    # Verify GPG
+    $gpg_sig = Get-AuthenticodeSignature $gpg_installer
+    $status = $gpg_sig | Select-Object -ExpandProperty 'Status'
+    if ( $status -ne 'Valid' ) {
+      Write-Output "GPG installer does not have a valid signature."
+      Exit 1
+    }
+    $sha256 = Get-FileHash $gpg_installer -Algorithm SHA256
+    if ( $sha256 -inotmatch '5D89E239790822711EAE2899467A764879D21440AB68E9413452FA96CEDEBA50') {
```
jericht: nit: Should we store this hash in a variable like we do for `$fingerprint`?

jericht (changes requested):
Just a couple of nitpicks, otherwise looks good. Thanks for fixing this!

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
horsmand: Fixed. You actually don't need the `{}` for the variable expansion, you can see it being done on the line above.

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
horsmand: Fixed

jericht (dismissed):
Looks good, thanks!

ddneilson (dismissed):

jericht (approved):

ddneilson (approved):
Thanks David. Let's do a fast-follow to update the Linux script to add the "already installed" check to it.

PR #297, opened 2021-01-25
chore(deps): bumping cdk from 1.83.0 to 1.86.0

horsmand:
Deployed the all-in-aws-infrastructure-basic app to confirm everything still works.

----


tools/pkglint/package.json
```diff
@@ -18,8 +18,8 @@
   "scripts": {
     "build": "tsc -b && eslint . --ext=.ts && chmod +x bin/pkglint",
     "test": "echo success",
-    "build+test": "npm run build && npm test",
-    "build+test+package": "npm run build+test",
+    "build+test": "yarn run build && yarn test",
+    "build+test+package": "yarn run build+test",
```
kozlove-aws, line 22: Is any reason for these changes?

ddneilson (approved):

lambda-layers/package.json
```diff
@@ -29,10 +29,10 @@
   "stability": "stable",
   "maturity": "stable",
   "devDependencies": {
-    "@types/node": "^14.14.20",
+    "@types/node": "^14.14.22",
     "typescript": "~4.1.3"
   },
   "dependencies": {
-    "aws-sdk": "^2.823.0"
```
ryyakobe, line 36: I noticed that in `lambda-layers/yarn.lock` we still have older version of `types/node` and `aws-sdk`.
Running
`cd cd lambda-layers`
`yarn install`
updates this `yarn.lock` file but I'm not sure if we usually commit these changes.
What do you think?

tools/pkglint/package.json
horsmand: This was done by CDK as well: https://github.com/aws/aws-rfdk/pull/297

lambda-layers/package.json
jusiskin: AFAIK we shoudn't need `lambda-layers/yarn.lock` file. RFDK uses yarn workspaces (aka a mono repo) and no other packages in the repo have a `yarn.lock` except in the root. The lock file does not appear to be built into the lambda layers, so I'd say we can disregard for now and remove it later after testing.

horsmand: Hmm, there's no reason for that directory to have its own `yarn.lock`. It's defined as a yarn workspace in our base `package.json` and gets built when running the `build.sh` script, so it should be using the `yarn.lock` located there. 

We need to decide if it makes sense to have it as a yarn workspace and built with our main build script, or if we should remove from there and treat it separately. There's pros and cons to each and I'd rather not lump this in with the CDK bump.

There's no harm in an out-of-date lockfile in this directory because all of our instructions and automated scripting runs `yarn` from the `lambda-layers` directory, which will make updates to the lockfile. If we used the `--frozen-lockfile` flag, that would be different.

ryyakobe (approved):
Thanks, David!

PR #301, opened 2021-01-29
feat(deadline): add ability to horizontally scale the RenderQueue

kozlove-aws:
## Problem
Since Deadline 10.1.10 we have possibility to create multiple RCS machines.

## Solution
Removing limitation of maximum instances for RCS if version greater than 10.1.9.

## Testing
 - Was creating farm with two RCS instances (set parameter `desired` in `RenderQueue` construct to 2
 - Verified that two RCS instances appeared in `Connection servers` panel
 - Increasing number of workers to 200 and run 2000 task, after that verified that both instances were used.
 - Increasing number of desired instances to 7 and verified that after deploying changed code were created 7 RCS instances and all appeared on `Connection servers` panel
 - Decreasing number of desired instances to 3 and verified that only 3 RCS instances were running after deployment.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
@@ -77,8 +77,8 @@ export interface RenderQueueSizeConstraints {
    * If this is set to a number, every deployment will reset the number of RCS processes
    * to this number. It is recommended to leave this value undefined.
    *
-   * Currently, the Deadline RCS does not properly support being horizontally scaled behind a load-balancer. For this
-   * reason, the desired number of processes can only be set to 1 currently.
+   * Deadline versions earlier than 10.1.10 do not support being horizontally scaled behind a load-balancer.
+   * For these versions of Deadline, the desired number of processes can only be set to 1.
    *
    * @default The min size.
    */
@@ -87,14 +87,27 @@ export interface RenderQueueSizeConstraints {
   /**
    * Minimum number of Deadline RCS processes that will serve RenderQueue requests.
    *
-   * Currently, the Deadline RCS does not properly support being horizontally scaled behind a load-balancer. For this
-   * reason, the minimum can be at most one, otherwise an error is thrown.
+   * Deadline versions earlier than 10.1.10 do not support being horizontally scaled behind a load-balancer.
+   * For these versions of Deadline, the minimum number of processes can only be set to 1, otherwise an error is thrown.
    *
-   * The minimum that this can value be set to is 1.
+   * The minimum that this value can be set to is 1.
    *
    * @default 1
    */
   readonly min?: number;
+
+  /**
+   * Maximum number of Deadline RCS processes that will serve RenderQueue requests.
+   *
+   * Deadline versions earlier than 10.1.10 do not support being horizontally scaled behind a load-balancer.
+   * For these versions of Deadline, the maximum number of processes can only be set to 1, otherwise an error is thrown.
+   *
+   * The minimum that this value can be set to is 1.
+   * This value cannot be less than `min` or `desired`.
+   *
+   * @default 1
+   */
+  readonly max?: number;
 }
 
 /**
@@ -283,7 +296,7 @@ export interface RenderQueueProps {
    * Constraints on the number of Deadline RCS processes that can be run as part of this
    * RenderQueue.
    *
-   * @default Allow no more and no less than one Deadline RCS to be running.
+   * @default Allow no less than one Deadline RCS to be running.
    */
   readonly renderQueueSize?: RenderQueueSizeConstraints;
 
--- a/packages/aws-rfdk/lib/deadline/lib/render-queue.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/render-queue.ts
@@ -65,6 +65,7 @@ import {
   IRepository,
   IVersion,
   RenderQueueProps,
+  RenderQueueSizeConstraints,
   VersionQuery,
 } from '.';
 
@@ -82,6 +83,7 @@ import {
 import {
   RenderQueueConnection,
 } from './rq-connection';
+import { Version } from './version';
 import {
   WaitForStableService,
 } from './wait-for-stable-service';
@@ -172,6 +174,11 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
   };
 
   /**
+  * The minimum Deadline version required for the Remote Connection Server to support load-balancing
+  */
+  private static readonly MINIMUM_LOAD_BALANCING_VERSION = new Version([10, 1, 10, 0]);
+
+  /**
    * Regular expression that validates a hostname (portion in front of the subdomain).
    */
   private static readonly RE_VALID_HOSTNAME = /^[a-z](?:[a-z0-9-]{0,61}[a-z0-9])?$/i;
@@ -248,6 +255,12 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
   private readonly rqConnection: RenderQueueConnection;
 
   /**
+   * Constraints on the number of Deadline RCS processes that can be run as part of this
+   * RenderQueue.
+   */
+  private readonly renderQueueSize: RenderQueueSizeConstraints;
+
+  /**
    * The listener on the ALB that is redirecting traffic to the RCS.
    */
   private readonly listener: ApplicationListener;
@@ -265,12 +278,19 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
   constructor(scope: Construct, id: string, props: RenderQueueProps) {
     super(scope, id);
 
-    // The RCS does not currently support horizontal scaling behind a load-balancer, so we limit to at most one instance
-    if (props.renderQueueSize && props.renderQueueSize.min !== undefined && props.renderQueueSize.min > 1) {
-      throw new Error(`renderQueueSize.min cannot be greater than 1 - got ${props.renderQueueSize.min}`);
-    }
-    if (props.renderQueueSize && props.renderQueueSize.desired !== undefined && props.renderQueueSize.desired > 1) {
-      throw new Error(`renderQueueSize.desired cannot be greater than 1 - got ${props.renderQueueSize.desired}`);
+    this.renderQueueSize = props?.renderQueueSize ?? {min: 1, max: 1};
+
+    if (props.version.isLessThan(RenderQueue.MINIMUM_LOAD_BALANCING_VERSION)) {
+      // Deadline versions earlier than 10.1.10 do not support horizontal scaling behind a load-balancer, so we limit to at most one instance
+      if ((this.renderQueueSize.min ?? 0) > 1) {
+        throw new Error(`renderQueueSize.min for Deadline version less than ${RenderQueue.MINIMUM_LOAD_BALANCING_VERSION.toString()} cannot be greater than 1 - got ${this.renderQueueSize.min}`);
+      }
+      if ((this.renderQueueSize.desired ?? 0) > 1) {
+        throw new Error(`renderQueueSize.desired for Deadline version less than ${RenderQueue.MINIMUM_LOAD_BALANCING_VERSION.toString()} cannot be greater than 1 - got ${this.renderQueueSize.desired}`);
+      }
+      if ((this.renderQueueSize.max ?? 0) > 1) {
+        throw new Error(`renderQueueSize.max for Deadline version less than ${RenderQueue.MINIMUM_LOAD_BALANCING_VERSION.toString()} cannot be greater than 1 - got ${this.renderQueueSize.max}`);
+      }
     }
 
     this.version = props?.version;
@@ -315,12 +335,16 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
     if (minCapacity < 1) {
       throw new Error(`renderQueueSize.min capacity must be at least 1: got ${minCapacity}`);
     }
+    const maxCapacity = this.renderQueueSize.max ?? this.renderQueueSize?.desired;
+    if (this.renderQueueSize?.desired && maxCapacity && this.renderQueueSize?.desired > maxCapacity) {
+      throw new Error(`renderQueueSize.desired capacity cannot be more than ${maxCapacity}: got ${this.renderQueueSize.desired}`);
+    }
     this.asg = this.cluster.addCapacity('RCS Capacity', {
       vpcSubnets: props.vpcSubnets ?? { subnetType: SubnetType.PRIVATE },
       instanceType: props.instanceType ?? new InstanceType('c5.large'),
       minCapacity,
-      desiredCapacity: props.renderQueueSize?.desired,
-      maxCapacity: 1,
+      desiredCapacity: this.renderQueueSize?.desired,
+      maxCapacity,
       blockDevices: [{
         deviceName: '/dev/xvda',
         // See: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-ami-storage-config.html
@@ -386,7 +410,7 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
     this.pattern = new ApplicationLoadBalancedEc2Service(this, 'AlbEc2ServicePattern', {
       certificate: this.clientCert,
       cluster: this.cluster,
-      desiredCount: props.renderQueueSize?.desired,
+      desiredCount: this.renderQueueSize?.desired,
       domainZone: props.hostname?.zone,
       domainName: loadBalancerFQDN,
       listenerPort: externalPortNumber,
--- a/packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
+++ b/packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
@@ -241,12 +241,8 @@ describe('RenderQueue', () => {
       });
     });
 
-    // Asserts that only one RCS container and ASG instance can be created.
-    // Deadline currently requires that successive API requests are serviced by a single RCS.
-    test.each([
-      [0],
-      [2],
-    ])('clamps to 1 - using %d', (min: number) => {
+    // Asserts that at least one RCS container and ASG instance must be created.
+    test('throws error when minimum size is 0', () => {
       // GIVEN
       const props: RenderQueueProps = {
         images,
@@ -254,7 +250,7 @@ describe('RenderQueue', () => {
         version: renderQueueVersion,
         vpc,
         renderQueueSize: {
-          min,
+          min: 0,
         },
       };
 
@@ -263,13 +259,40 @@ describe('RenderQueue', () => {
         new RenderQueue(stack, 'RenderQueue', props);
       })
         // THEN
-        .toThrow('renderQueueSize.min');
+        .toThrow('renderQueueSize.min capacity must be at least 1: got 0');
+    });
+
+    // Deadline before 10.1.10 requires that successive API requests are serviced by a single RCS.
+    test('validates Deadline pre 10.1.10 has min value of at most 1', () => {
+      // GIVEN
+      const min = 2;
+      const newStack = new Stack(app, 'NewStack');
+      const versionOld = new VersionQuery(newStack, 'VersionOld', {version: '10.1.9'});
+      const props: RenderQueueProps = {
+        images,
+        repository,
+        version: versionOld,
+        vpc,
+        renderQueueSize: {
+          min,
+        },
+      };
+
+      // WHEN
+      expect(() => {
+        new RenderQueue(newStack, 'RenderQueue', props);
+      })
+      // THEN
+        .toThrow(`renderQueueSize.min for Deadline version less than 10.1.10.0 cannot be greater than 1 - got ${min}`);
     });
 
     // Asserts that when the renderQueueSize.min prop is specified, the underlying ASG's min property is set accordingly.
-    test('configures minimum number of ASG instances', () => {
+    test.each([
+      [1],
+      [2],
+      [10],
+    ])('configures minimum number of ASG instances to %d', (min: number) => {
       // GIVEN
-      const min = 1;
       const isolatedStack = new Stack(app, 'IsolatedStack');
       const props: RenderQueueProps = {
         images,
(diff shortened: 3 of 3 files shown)


packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -444,6 +460,25 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
     tagConstruct(this);
   }
 
+  protected onValidate(): string[] {
```
ddneilson: 10.1.9 is over 6 months old at this point. It's probably just as good to bump the minimum supported version for the RFDK from Deadline 10.1.9 to 10.1.10 instead.

IIRC, Roman's upcoming work will require bumping it to 10.1.12.

Eliminates complexity here and in the additions to the VersionQuery.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -232,16 +235,25 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
-    // The RCS does not currently support horizontal scaling behind a load-balancer, so we limit to at most one instance
-    if (props.renderQueueSize && props.renderQueueSize.min !== undefined && props.renderQueueSize.min > 1) {
-      throw new Error(`renderQueueSize.min cannot be greater than 1 - got ${props.renderQueueSize.min}`);
-    }
-    if (props.renderQueueSize && props.renderQueueSize.desired !== undefined && props.renderQueueSize.desired > 1) {
-      throw new Error(`renderQueueSize.desired cannot be greater than 1 - got ${props.renderQueueSize.desired}`);
-    }
+    this.version = props?.version;
+
+    this.renderQueueSize = props?.renderQueueSize ? props?.renderQueueSize : {min: 1, max: 1} ;
```
jusiskin: nit: this can be simplified to:

```ts
this.renderQueueSize = props?.renderQueueSize ?? {min: 1, max: 1} ;

File: packages/aws-rfdk/lib/deadline/lib/version-query.ts [hunk 3]
```diff
@@ -139,9 +139,10 @@ export class VersionQuery extends VersionQueryBase {
       resourceType: 'Custom::RFDK_DEADLINE_INSTALLERS',
     });
 
-    this.majorVersion = Token.asNumber(deadlineResource.getAtt('MajorVersion'));
-    this.minorVersion = Token.asNumber(deadlineResource.getAtt('MinorVersion'));
-    this.releaseVersion = Token.asNumber(deadlineResource.getAtt('ReleaseVersion'));
+    const components = props?.version?.split('.').map(x => parseInt(x));
+    this.majorVersion = components ? components[0] : Token.asNumber(deadlineResource.getAtt('MajorVersion'));
+    this.minorVersion = components && components?.length > 0 ? components[1] : Token.asNumber(deadlineResource.getAtt('MinorVersion'));
+    this.releaseVersion = components && components?.length > 1 ? components[2] : Token.asNumber(deadlineResource.getAtt('ReleaseVersion'));
```
jusiskin: I believe both length checks need to be incremented here:

```ts
    this.minorVersion = components && components?.length > 1 ? components[1] : Token.asNumber(deadlineResource.getAtt('MinorVersion'));
    this.releaseVersion = components && components?.length > 2 ? components[2] : Token.asNumber(deadlineResource.getAtt('ReleaseVersion'));
```

packages/aws-rfdk/lib/deadline/lib/version-query.ts
```diff
@@ -139,9 +139,10 @@ export class VersionQuery extends VersionQueryBase {
       resourceType: 'Custom::RFDK_DEADLINE_INSTALLERS',
     });
 
-    this.majorVersion = Token.asNumber(deadlineResource.getAtt('MajorVersion'));
-    this.minorVersion = Token.asNumber(deadlineResource.getAtt('MinorVersion'));
-    this.releaseVersion = Token.asNumber(deadlineResource.getAtt('ReleaseVersion'));
+    const components = props?.version?.split('.').map(x => parseInt(x));
+    this.majorVersion = components ? components[0] : Token.asNumber(deadlineResource.getAtt('MajorVersion'));
```
jusiskin: The truth value of an empty array is `true`. This means that the current logic here will lead to an index out of bounds error if `props.version` is an empty string:

```js
> ''.split('.') ? true : false
true
```

packages/aws-rfdk/lib/deadline/lib/version-query.ts
```diff
@@ -139,9 +139,10 @@ export class VersionQuery extends VersionQueryBase {
       resourceType: 'Custom::RFDK_DEADLINE_INSTALLERS',
     });
 
-    this.majorVersion = Token.asNumber(deadlineResource.getAtt('MajorVersion'));
-    this.minorVersion = Token.asNumber(deadlineResource.getAtt('MinorVersion'));
-    this.releaseVersion = Token.asNumber(deadlineResource.getAtt('ReleaseVersion'));
+    const components = props?.version?.split('.').map(x => parseInt(x));
```
jusiskin: We should add  test coverage for this new behavior. I'm actually surprised that we don't have coverage of this and we haven't broken an existing test already. We should add tests to see what gets synthesized when the values of:

*    `VersionQuery.majorVersion`
*    `VersionQuery.minorVersion`
*    `VersionQuery.releaseVersion`

are referenced. We can write tests that assert the value of [`Stack.resolve(version.majorVersion)`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html#resolveobj) for that.

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -76,24 +76,28 @@ export interface RenderQueueSizeConstraints {
    * If this is set to a number, every deployment will reset the number of RCS processes
    * to this number. It is recommended to leave this value undefined.
    *
-   * Currently, the Deadline RCS does not properly support being horizontally scaled behind a load-balancer. For this
-   * reason, the desired number of processes can only be set to 1 currently.
-   *
```
jusiskin: This comment still applies to versions earlier than 10.1.10. Perhaps we should keep this in, but just qualify the statement:

>    Deadline versions earlier than 10.1.10 do not support being horizontally scaled behind a load-balancer. For these versions of Deadline, the desired number of processes can only be set to 1.

Same feedback applies to the minimum property below.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -232,16 +235,25 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+   */
+  private readonly version?: IVersion;
+
+  /**
+   * Constraints on the number of Deadline RCS processes that can be run as part of this
+   * RenderQueue.
+   *
+   * @default Allow no less than one Deadline RCS to be running.
+   */
+  readonly renderQueueSize: RenderQueueSizeConstraints;
```
jusiskin: By default, TypeScript will mark this as `public`. I'm not sure this needs to be. What do you think?

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -444,6 +460,25 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
     tagConstruct(this);
   }
 
+  protected onValidate(): string[] {
+    const validationErrors = [];
+
+    if (this.version?.majorVersion == 10 && this.version?.minorVersion == 1 && this.version?.releaseVersion == 9 ) {
+      if (this.renderQueueSize) {
+        if (this.renderQueueSize.min !== undefined && this.renderQueueSize.min > 1) {
```
jusiskin: nit: we can use [nullish coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing) to simplify this to:

```ts
if (this.renderQueueSize.min ?? 0 > 1) {
```

Run [this example](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYjgGwFlRkBvAKGSuSgjgBNd8BPZAW1AH4AuZEAVzYAjaAG5yAX3LkEuAM5hkANwL8IvDFlwFiIZAF4yU8sBjIAFCvxqAdBz2dOyAAzIAfMgCMASjKVqsiBy2PgQNvjYAObmAORWtvbIkbRwkFDIYAAWcHqeMd7iUhD4cigU1MiBwaHhUbHxYYkg2IrJdGkZ2bn5hUA) for proof.

Same applies to the checks below.

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -255,13 +252,41 @@ describe('RenderQueue', () => {
         new RenderQueue(stack, 'RenderQueue', props);
       })
         // THEN
-        .toThrow('renderQueueSize.min');
+        .toThrow('renderQueueSize.min capacity must be at least 1: got 0');
+    });
+
+    // Deadline before 10.1.10 requires that successive API requests are serviced by a single RCS.
+    test('old version use more than 1', () => {
```
jusiskin: If this breaks in the future, the developer will only see "old version use more than 1" failed.  How would you feel about moving the content of this comment into the test name that gets output by jest. If you agree, how about:

```ts
test('validates Deadline pre 10.1.10 has min value of at most 1', () => {
```

Same feedback applies to the equivalent `renderQueueSize.max` and `renderQueueSize.desired` tests.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -444,6 +460,25 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+          validationErrors.push(`renderQueueSize.min cannot be greater than 1 for this Deadline version - got ${this.renderQueueSize.min}`);
```
jusiskin: To help guide users that get this validation failure, should we let them know what is the minimum version of Deadline required?

packages/aws-rfdk/lib/deadline/lib/version-query.ts
jusiskin: It appears that we currently do not have any synthesis-time validation on the version string passed in. Do you think it would be better to validate the version query expression (e.g. using a regex) before we perform these string operations on it? That way we have an opportunity to fail with a more user-friendly error message than an unhandled exception here.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -277,16 +289,20 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+      throw new Error(`renderQueueSize.desired capacity can not be more than ${maxCapacity}: got ${this.renderQueueSize.desired}`);
+    }
     this.asg = this.cluster.addCapacity('RCS Capacity', {
       vpcSubnets: props.vpcSubnets ?? { subnetType: SubnetType.PRIVATE },
       instanceType: props.instanceType ?? new InstanceType('c5.large'),
       minCapacity,
-      desiredCapacity: props.renderQueueSize?.desired,
-      maxCapacity: 1,
+      desiredCapacity: this.renderQueueSize?.desired,
+      maxCapacity,
```
jusiskin, line 347: Looks like we now can configure the auto-scaling group (ASG), but we have not done anything to scale the ECS service below as well.

This requires careful consideration on how we go about this.

jusiskin (changes requested):
This looks mostly there for configuring the auto-scaling group, but we might need to take a step back to figure out how we want to scale the ECS service in sync with the ASG's capacity.

packages/aws-rfdk/lib/deadline/lib/version-query.ts
kozlove-aws: I am verifying not empty value but undefined. 
Maybe such line will be more clear?
`this.majorVersion = props?.version ? components[0] : Token.asNumber(deadlineResource.getAtt('MajorVersion'));`

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
kozlove-aws: I can imagine situation when customer need to get this value.
But as soon as we do not use it I can make it private.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -444,6 +460,25 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
     tagConstruct(this);
   }
 
+  protected onValidate(): string[] {
+    const validationErrors = [];
+
+    if (this.version?.majorVersion == 10 && this.version?.minorVersion == 1 && this.version?.releaseVersion == 9 ) {
```
ryyakobe: Is there a reason we use here `==` and not a more strict `===`?

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
ryyakobe: This method is actually deprecated according to https://docs.aws.amazon.com/cdk/api/latest/docs/constructs.Construct.html

So probably it's not a good idea to use it.

jusiskin: That doc reference points to `@aws-cdk/constructs.Construct` whereas the here we're using `@aws-cdk/core.Construct`. The [`onValidate` is not deprecated](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Construct.html#protected-onwbrvalidate) in this case. That being said, I believe we should probably use the `validate` method instead. The `onValidate` is a wrapper of `validate` that subclasses can use to add pre/post logic to the validation lifecycle.

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -87,14 +87,24 @@ export interface RenderQueueSizeConstraints {
...
    *
    * @default 1
    */
   readonly min?: number;
+
+  /**
+   * Maximum number of Deadline RCS processes that will serve RenderQueue requests.
+   *
+   * The minimum that this can value be set to is 1.
+   * This value can not be less than min or desired
```
jericht:
```suggestion
   * This value cannot be less than `min` or `desired`.
```

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -307,16 +324,20 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
     });
 
-    const minCapacity = props.renderQueueSize?.min ?? 1;
+    const minCapacity = this.renderQueueSize?.min ?? 1;
     if (minCapacity < 1) {
       throw new Error(`renderQueueSize.min capacity must be at least 1: got ${minCapacity}`);
     }
+    const maxCapacity = this.renderQueueSize.max ?? this.renderQueueSize?.desired;
+    if (this.renderQueueSize?.desired && maxCapacity && this.renderQueueSize?.desired > maxCapacity) {
+      throw new Error(`renderQueueSize.desired capacity can not be more than ${maxCapacity}: got ${this.renderQueueSize.desired}`);
```
jericht:
```suggestion
      throw new Error(`renderQueueSize.desired capacity cannot be more than ${maxCapacity}: got ${this.renderQueueSize.desired}`);
```

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -242,19 +242,15 @@ describe('RenderQueue', () => {
     });
 
     // Asserts that only one RCS container and ASG instance can be created.
```
jericht:
```suggestion
    // Asserts that at least one RCS container and ASG instance must be created.
```

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -242,19 +242,15 @@ describe('RenderQueue', () => {
...
-    // Deadline currently requires that successive API requests are serviced by a single RCS.
-    test.each([
-      [0],
-      [2],
-    ])('clamps to 1 - using %d', (min: number) => {
+    test('clamps to 1 - using 0', () => {
```
jericht:
```suggestion
    test('throws error when minimum size is 0', () => {
```

jericht (changes requested):
A few very minor comments, otherwise this LGTM!

packages/aws-rfdk/lib/deadline/lib/version-query.ts
jusiskin: ~~Hmm, it might be more readable to assign `components` in the undefined case:~~

```ts
const components = props?.version?.split('.').map(x => parseInt(x)) ?? [];
```

~~This way `components` is guaranteed to be an array and we can simplify the following statements. For example:~~

```ts
this.majorVersion = components.length > 0 ? components[0] : Token.asNumber(deadlineResource.getAtt('MajorVersion'));
```

EDIT: Was looking at the old revision of this PR. These changes were removed from the PR and no longer apply

packages/aws-rfdk/lib/deadline/lib/version-query.ts
jusiskin: Closing as the PR no longer modifies this file

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Looks good! thanks

packages/aws-rfdk/lib/deadline/lib/version-query.ts
jusiskin: No longer applies. Resolving

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
jusiskin: Looks good. Thanks!

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: I've always taken the opposite approach: default to `private` access and only make public APIs when there is a use-case.

These are passed in as props, so the CDK app developer should already have the values. If we make these public, they become a public API that we need to maintain and then we'd want to have an automated test for asserting their behavior. What do you think?

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
jusiskin: Much better, thanks. resolving!

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Thanks - looks great. Resolved

packages/aws-rfdk/lib/deadline/lib/version-query.ts
jusiskin: Stale. Resolving

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Resolved

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -87,14 +87,24 @@ export interface RenderQueueSizeConstraints {
   /**
    * Minimum number of Deadline RCS processes that will serve RenderQueue requests.
    *
-   * Currently, the Deadline RCS does not properly support being horizontally scaled behind a load-balancer. For this
-   * reason, the minimum can be at most one, otherwise an error is thrown.
+   * Deadline versions earlier than 10.1.10 do not support being horizontally scaled behind a load-balancer.
+   * For these versions of Deadline, the desired number of processes can only be set to 1, otherwise an error is thrown.
```
jusiskin: Looks like a copy/paste artifact:

`desired` &rarr; `minimum`

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -87,14 +87,24 @@ export interface RenderQueueSizeConstraints {
...
+
+  /**
+   * Maximum number of Deadline RCS processes that will serve RenderQueue requests.
+   *
+   * The minimum that this can value be set to is 1.
+   * This value cannot be less than `min` or `desired`.
+   *
+   * @default 1
+   */
+  readonly max?: number;
```
jusiskin: Looks like the:

>   Deadline versions earlier than 10.1.10 do not support being horizontally scaled behind a load-balancer ...

note for this TSDoc string is missing. The `RenderQueue` constructor validates this `max` property with the `version` check, so we probably want to document it here.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -247,6 +249,14 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
    */
   private readonly rqConnection: RenderQueueConnection;
 
+  /**
+   * Constraints on the number of Deadline RCS processes that can be run as part of this
+   * RenderQueue.
+   *
+   * @default Allow no less than one Deadline RCS to be running.
+   */
+  readonly renderQueueSize: RenderQueueSizeConstraints;
```
jusiskin: nit: we don't require a `@default` here since this is an attribute and it is not null-able.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -265,16 +275,24 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
   constructor(scope: Construct, id: string, props: RenderQueueProps) {
     super(scope, id);
 
-    // The RCS does not currently support horizontal scaling behind a load-balancer, so we limit to at most one instance
-    if (props.renderQueueSize && props.renderQueueSize.min !== undefined && props.renderQueueSize.min > 1) {
-      throw new Error(`renderQueueSize.min cannot be greater than 1 - got ${props.renderQueueSize.min}`);
-    }
-    if (props.renderQueueSize && props.renderQueueSize.desired !== undefined && props.renderQueueSize.desired > 1) {
-      throw new Error(`renderQueueSize.desired cannot be greater than 1 - got ${props.renderQueueSize.desired}`);
+    const minimumVersion: Version = new Version([10, 1, 10, 0]);
```
jusiskin: nit: can we make this a class constant and rename it to be slightly more specific? how about:

```ts
/**
 * The minimum Deadline version required for the Remote Connection Server to support load-balancing
 */
private static readonly MINIMUM_LOAD_BALANCING_VERSION = new Version([10, 1, 10, 0]);
```

this would make the code below read:

```ts
if (props.version.isLessThan(RenderQueue.MINIMUM_LOAD_BALANCING_VERSION)) {
```

which seems more readable to me. What do you think?

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -291,6 +314,89 @@ describe('RenderQueue', () => {
...
+      new RenderQueue(isolatedStack, 'RenderQueue', props);
+
+      // THEN
+      expectCDK(isolatedStack).to(haveResourceLike('AWS::AutoScaling::AutoScalingGroup', {
+        MaxSize: max.toString(),
+      }));
+    });
+  });
+
   describe('renderQueueSize.desired', () => {
```
jusiskin, line 400: It looks like we're missing a test for the values of the `DesiredCapacity` of the `AWS::ECS::Service`  resource when `renderQueueSize.desired` is not passed in.

jusiskin (changes requested):
Looking great, @kozlove-aws. Just a few minor copy/paste issues, one missing test, and a code readability/organization suggestion. Otherwise, great work and thanks for all of the end-to-end testing.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Looks good. As discussed offline, my sample code was missing parentheses. It became:

```ts
if ((this.renderQueueSize.min ?? 0) > 1) {
```

It looks good now. Resolved

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
jusiskin: Looks good. thanks!

jusiskin (dismissed):
All looks good! Thanks for this important feature enhancement

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -87,14 +87,27 @@ export interface RenderQueueSizeConstraints {
...
    */
   readonly min?: number;
+
+  /**
+   * Maximum number of Deadline RCS processes that will serve RenderQueue requests.
+   *
+   * Deadline versions earlier than 10.1.10 do not support being horizontally scaled behind a load-balancer.
+   * For these versions of Deadline, the maximum number of processes can only be set to 1, otherwise an error is thrown.
+   *
+   * The minimum that this can value be set to is 1.
```
jericht:
```suggestion
   * The minimum that this value can be set to is 1.
```

jericht (changes requested):
Just one tiny typo in a docstring otherwise this looks good to me!

jericht (approved):
Looks good!

jusiskin (approved):

PR #306, opened 2021-02-03
fix(deadline): VersionQuery cross-stack issue

kozlove-aws:
### Problem
The VersionQuery RFDK construct abuses the CloudFormation resource model.
When we are referencing the VersionQuery from another stack, the values in stack outputs are not allowed to change.
This becomes a problem when new version released and was not used specific version.

### Solution
To solve this problem in short-term perspective we add a synthesis-time validation check to ensure consumers of a VersionQuery are in the same stack. This would mitigate the risk of customers using them cross-stack and guide them towards one VersionQuery instance per stack.

### Testing
Were created unit tests and validated that error throwing when used VersionQuery from different stack.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/render-queue.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/render-queue.ts
@@ -52,13 +52,16 @@ import {
 import {
   Construct,
   IConstruct,
+  Stack,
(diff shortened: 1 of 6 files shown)


ddneilson (approved):

jericht (approved):
LGTM

PR #314, opened 2021-02-10
feat(examples): local zones examples

### LINKED ISSUE #313: Local Zones Example
### Use Case
If you're in southern California and are trying to transfer assets directly to workers, the latency to transfer to us-west-2 in Oregon could be reduced by utilizing the local zones in LA. There are more local zones being started in other parts of the US that will create a similar benefit in other regions too.

### Proposed Solution
Write up an example app that positions a worker fleet in a local zone.

horsmand:
Fixes https://github.com/aws/aws-rfdk/issues/313

----


examples/deadline/Local-Zone/README.md
```diff
@@ -0,0 +1,41 @@
+# RFDK Sample Application - Local Zones
+
+If you have large asset files that your worker instances need to process, any bit of latency can have a big impact on the time your renders take. This example will walk you through setting up your workers in a local zone while leaving the rest of the render farm in standard availability zones. Currently Amazon has launched a local zone in Los Angeles that is a part of the us-west-2 region, but they have more on the way. For more information on where local zones are avialable, how to get access, and what services they provide, refer to the [AWS Local Zones about page](https://aws.amazon.com/about-aws/global-infrastructure/localzones/).
```
ddneilson: I think we'd want to add that the use-case for the local zone would be where you are connecting your Workers to a datacenter, or on-premises network, via Direct Connect or VPN and want to have your workers directly mount your filesystem over that connection. If the filesystem is closer to the local zone than other AWS regions, then the improved latency will improve render times.

This just says that latency can have a big impact, but doesn't draw the line from the on-prem/datacenter filer to wanting to use a local zone.

Also, are we just looking at Workers in the local zone? Is it possible to host more of the farm infrastructure in the local zone? e.g. The EFS & DocDB for the repo, and/or the RenderQueue?

examples/deadline/Local-Zone/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Architecture
+
+This example app assumes you're familiar with the general architecture of an RFDK render farm. If not, please refer to the [All-In-AWS-Infrastructure-Basic](../All-In-AWS-Infrastructure-Basic/README.md) example for the basics.
+
+### Components
+
+#### Network Tier
+
+The network tier sets up a [VPC](https://aws.amazon.com/vpc/), which is within a [Private Hosted Zone](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zones-private.html). These both span across all of the standard availability zones and local zones that are used, but the NAT Gateway for the VPC is only added to the stardard zones, as it is not avaiable in any local zones at this time.
```
ddneilson: Not quite correct. The VPC isn't within a Private Hosted Zone. The private hosted zone is setup within the VPC. The VPC has its own internal DNS via Route53; the Private Hosted Zone adds, essentially, a DNS namespace within the VPC's DNS into which we can & do create records for the RenderQueue.

examples/deadline/Local-Zone/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+#### Security Tier
+
+This holds the root CA certificate used for signing any certificates required by the farm, such as the one used by the render queue.
+
+#### Service Tier
+
+The service tier includes a bastion instance, positioned in a public subnet in one of the standard availability zones. It can be used to connect to instances in the worker fleet by first connecting to it through the AWS console using the session manager. Note that you'd have to provide your worker fleet with an ssh key and then copy it over onto the bastion host for this to work. For more information visit the [Bastion Hosts CDK documentation](https://docs.aws.amazon.com/cdk/api/latest/docs/aws-ec2-readme.html#bastion-hosts).
```
ddneilson: It's probably worth eliminating the Bastion from our examples in favour of using the SSM Connect method to connect to the server components.

examples/deadline/Local-Zone/python/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+.env
+*.egg-info
+venv
+build
+
+# CDK asset staging directory
+.cdk.staging
+cdk.out
+cdk.context.json
+stage
```
ddneilson: nit: Missing newline at end of file.

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+6. Change the value of the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker.
+
+7. Also in `package/lib/config.py`, you can set the `availability_zones_standard` and `availability_zones_local` values to the availability zones you want to use. These values must all be from the same region. It's recommended you use at least two standard zones, but you can use more if you'd like. For the local zones you can use as many as you'd like.
+
+8. Due to limitations of the Python implementation of CDK, you'll need to manually modify the availability zones in the CDK context file so your VPC deploys where you want it. In the TypeScript version we accomplish this by overriding `availabilityZones()` on the network tier stack, but this doesn't work in Python. Instead, either create or modify `cdk.context.json` to have an entry like the following, note you need to replace `<account #>` with your AWS account number. The example uses us-west-2, so the key includes this, but if you were to use a different region you'd also have to modify it in the key.
```
ddneilson: afaik, the `cdk.context.json` file only gets created after doing a synth, no? If so, then there's a step missing where they'll have to do a synth to generate the file so that they can modify it.

examples/deadline/Local-Zone/ts/.gitignore
```diff
@@ -0,0 +1,20 @@
...
+# Exclude package artifacts
+dist
+*.tgz
+.LAST_PACKAGE
+
+#Defines license that must be present
+!license-header.js
+
+# The staged files for Deadline
+stage
```
ddneilson: nit: missing newline

examples/deadline/Local-Zone/ts/lib/network-tier.ts
```diff
@@ -0,0 +1,121 @@
...
+    // We're creating a SubnetSelection with only the standard availability zones to be
+    // used to put the NAT gateway in and the VPC endpoints, because the local zones do no
+    // have these available.
+    const subnets: SubnetSelection = {
+      availabilityZones: config.availabilityZonesStandard,
+      subnetType: SubnetType.PUBLIC,
+    };
+
+    this.vpc = new Vpc(this, 'Vpc', {
+      maxAzs: this.availabilityZones.length,
```
ddneilson, line 89: How is this creating subnets in the Local Zone? CDK essentially just randomly picks the first N AZs in the region -- https://github.com/aws/aws-cdk/blob/09ba573a816cc4fa9898c1700136bb332801721c/packages/%40aws-cdk/aws-ec2/lib/vpc.ts#L1217

Unlike the Python example, this isn't modifying `cdk.context.json`

ddneilson (dismissed):

examples/deadline/Local-Zone/README.md
horsmand: EFS and DocDB don't have support in the local zone. I could look at moving the Render Queue, but I'm not sure what benefit there would be since the repository would still be in a standard availability zone.

examples/deadline/Local-Zone/ts/lib/network-tier.ts
horsmand: This stack overrides `public get availabilityZones(): string[]` which will define what gets returned by `this.availabilityZones`

examples/deadline/Local-Zone/python/package/lib/config.py
```diff
@@ -0,0 +1,51 @@
...
+
+        # The local availability zones that will hold the worker fleet. They must belong to the same region as the standard
+        # zones. The default value being provided here is one of the two local zones in us-west-2, located in Los Angeles.
+        self.availability_zones_local: List[str] = ['us-west-2-lax-1a']
+
+        # The version of Deadline to use on the render farm. Leave as None for the latest release or specify a version
+        # to pin to. Some examples of pinned version values are "10", "10.1", or "10.1.13"
+        self.deadline_version: Optional[str] = '10.1.13.1'
+
+        # A map of regions to Deadline Client Linux AMIs.As an example, the Linux Deadline 10.1.13.1 AMI ID
```
gdday: Nit: Missing space between period and 'As'.

examples/deadline/Local-Zone/python/package/lib/network_tier.py
```diff
@@ -0,0 +1,114 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from aws_cdk.core import (
+    Stack,
+    Construct
```
gdday: Nit: The imports aren't alphabetical.

examples/deadline/Local-Zone/README.md
gdday: This is the same problem I'm facing too. In my case it does make sense to support the Workers in the Local Zone because I can create the FSx volumes (that store the data for the render jobs) in the Local Zones too.

But there are quite a few resources that will have to remain in us-west-2 (or the parent region, however you want to call it).

examples/deadline/Local-Zone/README.md
gdday: Yes! I think this is a much better workflow than SSH keys.

examples/deadline/Local-Zone/python/README.md
gdday: I have a sinking feeling about this. That's quite a defect in the JSII.

I wonder whether this can be worked around using the `construct.node.setContext` method? It seems like `availabilityZones` is a part of the context.

examples/deadline/Local-Zone/python/package/lib/service_tier.py
```diff
@@ -0,0 +1,173 @@
...
+        :param props: The properties for this construct.
+        :param kwargs: Any kwargs that need to be passed on to the parent class.
+        """
+        super().__init__(scope, stack_id, **kwargs)
+
+        # Bastion instance for convenience (e.g. SSH into RenderQueue and WorkerFleet instances).
+        # It is being deployed into the standard availability zones, but has access to the worker
+        # instances that get deployed into a local zone. Not a critical component of the farm, so
+        # this can be safely removed.
+        self.bastion = BastionHostLinux(
```
gdday: Mentioned elsewhere, but mentioning here: we should always recommend using SSM over Bastions (where possible!)

examples/deadline/Local-Zone/python/README.md
ddneilson: @horsmand I just noticed this section of the CDK's VPC docs: https://docs.aws.amazon.com/cdk/api/latest/docs/aws-ec2-readme.html#control-over-availability-zones

Have you tried the AZ override that they suggest (overriding the Stack's `availabilityZones` property)?

edit: Wouldn't surprise me if that workaround was needed on all stacks in the app.

horsmand: I've tried overriding the `availabilityZones()` method as they mention there (this worked in the TypeScript app) as well as the property itself, but it's immutable.

I didn't try `construct.node.setContext()`, I'll give that a go.

ddneilson: Should be possible in Python as well, though the syntax will be a hair more complex. AFAIK, it's something like:
```python
@property
def availability_zones():
  return [ ... <whatever> ... ]
```

examples/deadline/Local-Zone/python/package/lib/service_tier.py
horsmand: Updated

examples/deadline/Local-Zone/python/package/lib/network_tier.py
horsmand: updated

examples/deadline/Local-Zone/python/package/lib/config.py
horsmand: updated

examples/deadline/Local-Zone/ts/lib/network-tier.ts
horsmand: Added this comment to the readme

examples/deadline/Local-Zone/ts/.gitignore
horsmand: updated

examples/deadline/Local-Zone/python/README.md
horsmand: Well I thought I tried overriding the `availability_zones` method every way possible on my first attempt, but it seems I was able to get it to work by copying the signature exactly:
```
@builtins.property # type: ignore
    @jsii.member(jsii_name="availabilityZones")
    def availability_zones(self) -> typing.List[builtins.str]:
```

examples/deadline/Local-Zone/python/.gitignore
horsmand: updated

examples/deadline/Local-Zone/README.md
horsmand: updated

examples/deadline/Local-Zone/README.md
horsmand: Updated

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,60 @@
...
+    ./build.sh
+    ./pack.sh
+    # Exit the Docker container
+    exit
+    # Navigate back to the example directory
+    popd
+    pip install ../../../../dist/python/aws-rfdk-<version>.tar.gz
+    ```
+
+4.  You must read and accept the [AWS Thinkbox End-User License Agreement (EULA)](https://www.awsthinkbox.com/end-user-license-agreement) to deploy and run Deadline. To do so, change the value of the `accept_aws_thinkbox_eula` in `package/lib/config.py` to `ACCEPTS` like this:
```
jericht: > to `ACCEPTS` like this:

Using `USER_ACCEPTS_AWS_THINKBOX_EULA` instead of `ACCEPTS` seems more accurate here. What do you think?

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,60 @@
...
+    ```
+
+5. Change the value of the `deadline_version` variable in `package/lib/config.py` to specify the desired version of Deadline to be deployed to your render farm. RFDK is compatible with Deadline versions 10.1.9.x and later. To see the available versions of Deadline, consult the [Deadline release notes](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/release-notes.html). It is recommended to use the latest version of Deadline available when building your farm, but to pin this version when the farm is ready for production use. For example, to pin to the latest `10.1.13` release of Deadline, use `10.1.13.1`.
+
+6. Change the value of the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker.
+
+7. Also in `package/lib/config.py`, you can set the `availability_zones_standard` and `availability_zones_local` values to the availability zones you want to use. These values must all be from the same region. It's recommended you use at least two standard zones, but you can use more if you'd like. For the local zones you can use as many as you'd like.
```
jericht:
```suggestion
7. Also in `package/lib/config.py`, you can set the `availability_zones_standard` and `availability_zones_local` values to the availability zones you want to use. These values must all be from the same region. It's recommended you use at least two standard zones, but you can use more if you'd like. For the local zones, you can use as many as you'd like.
```

examples/deadline/Local-Zone/python/package/lib/compute_tier.py
```diff
@@ -0,0 +1,99 @@
...
+        :param props: The properties of this construct.
+        :param kwargs: Any kwargs that need to be passed on to the parent class.
+        """
+        super().__init__(scope, stack_id, **kwargs)
+
+        # We can put the health monitor and worker fleet in all of the local zones we're using
+        subnets = SubnetSelection(
+            availability_zones=props.availability_zones,
+            subnet_type=SubnetType.PRIVATE,
+            # one_per_az=True,
```
jericht: Do we want this left out? We have it as `true` in the Typescript version

examples/deadline/Local-Zone/python/package/lib/network_tier.py
```diff
@@ -0,0 +1,124 @@
...
+        # We're creating a SubnetSelection with only the standard availability zones
+        # to be used to put the NAT gateway in and the VPC endpoints, because the local zones do no have
+        # these available.
+        subnets = SubnetSelection(
+            availability_zones=config.availability_zones_standard,
+            subnet_type=SubnetType.PUBLIC
+        )
+
+        # The VPC that all components of the render farm will be created in.
+        # This overrides the availability zones the Stack will use. The zones that we set here are what
```
jericht: Looks like this is a copy of the comment in `def availability_zones` above, but doesn't make sense in this context. Can we update this to say that we're using the override above for the `Vpc`?

examples/deadline/Local-Zone/python/package/lib/service_tier.py
```diff
@@ -0,0 +1,151 @@
...
+            traffic_encryption=RenderQueueTrafficEncryptionProps(
+                external_tls=RenderQueueExternalTLSProps(
+                    rfdk_certificate=server_cert
+                ),
+                internal_protocol=ApplicationProtocol.HTTPS
+            ),
+            version=self.version,
+            vpc_subnets=render_queue_subnets,
+            vpc_subnets_alb=render_queue_alb_subnets,
+            deletion_protection=False
```
jericht, line 155: We had to add a comment above this in our first example app so that users were aware of the consequences of disabling deletion protection on the RQ, so I wonder if we should add it in here as well.
```
            # TODO - Evaluate deletion protection for your own needs. This is set to false to
            # cleanly remove everything when this stack is destroyed. If you would like to ensure
            # that this resource is not accidentally deleted, you should set this to true.
```
I'm not sure if we need to have it for every example we do though. What do you think?

examples/deadline/Local-Zone/ts/README.md
```diff
@@ -0,0 +1,52 @@
...
+    ```
+
+3.  Change the value of the `deadlineVersion` variable in `bin/config.ts` to specify the desired version of Deadline to be deployed to your render farm. RFDK is compatible with Deadline versions 10.1.9.x and later. To see the available versions of Deadline, consult the [Deadline release notes](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/release-notes.html). It is recommended to use the latest version of Deadline available when building your farm, but to pin this version when the farm is ready for production use. For example, to pin to the latest `10.1.13` release of Deadline, use `10.1.13.1`.
+
+4. Change the value of the `deadlineClientLinuxAmiMap` variable in `bin/config.ts` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker.
+
+5. Also in `bin/config.ts`, you can set the `availabilityZonesStandard` and `availabilityZonesLocal` values to the availability zones you want to use. These values must all be from the same region. It's recommended you use at least two standard zones, but you can use more if you'd like. For the local zones you can use as many as you'd like.
```
jericht:
```suggestion
5. Also in `bin/config.ts`, you can set the `availabilityZonesStandard` and `availabilityZonesLocal` values to the availability zones you want to use. These values must all be from the same region. It's recommended you use at least two standard zones, but you can use more if you'd like. For the local zones, you can use as many as you'd like.
```

examples/deadline/Local-Zone/ts/bin/app.ts
```diff
@@ -0,0 +1,62 @@
...
+import { ServiceTier } from '../lib/service-tier';
+
+  // ------------------------------ //
+  // --- Validate Config Values --- //
+  // ------------------------------ //
+  if (!config.keyPairName) {
+    console.log('EC2 key pair name not specified. You will not have SSH access to the render farm.');
+  }
+
+  if (config.deadlineClientLinuxAmiMap === {['region']: 'ami-id'}) {
```
jericht: We can remove this since it was a safeguard for when the `config.ts` file didn't provide default AMI IDs.

examples/deadline/Local-Zone/ts/lib/compute-tier.ts
```diff
@@ -0,0 +1,114 @@
...
+    const subnets: SubnetSelection = {
+      availabilityZones: props.availabilityZones,
+      subnetType: SubnetType.PRIVATE,
+      onePerAz: true,
+    };
+
+    this.healthMonitor = new HealthMonitor(this, 'HealthMonitor', {
+      vpc: props.vpc,
+      vpcSubnets: subnets,
+      deletionProtection: false,
```
jericht, line 105: Same comment about deletion protection here as in the RenderQueue. Should we add in that `TODO` comment here as well?

examples/deadline/Local-Zone/ts/lib/compute-tier.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    this.workerFleet = new WorkerInstanceFleet(this, 'WorkerFleet', {
+      vpc: props.vpc,
+      renderQueue: props.renderQueue,
+      workerMachineImage: props.workerMachineImage,
+      healthMonitor: this.healthMonitor,
+      keyName: "OpenSshDevBox",
```
jericht: Should this be `props.keyPairName`?

examples/deadline/Local-Zone/ts/lib/compute-tier.ts
```diff
@@ -0,0 +1,114 @@
...
+      // Not all instance types will be available in local zones. For a list of the instance types
+      // available in each local zone, you can refer to:
+      // https://aws.amazon.com/about-aws/global-infrastructure/localzones/features/#AWS_Services
+      instanceType: InstanceType.of(InstanceClass.T3, InstanceSize.LARGE),
```
jericht: nit: We're using the `InstanceType.of(InstanceClass.STANDARD5, InstanceSize.LARGE)` in the python sample but a different one here. Should we change one of them to be the same as the other for consistency?

examples/deadline/Local-Zone/ts/lib/service-tier.ts
```diff
@@ -0,0 +1,162 @@
...
+      trafficEncryption: {
+        externalTLS: {
+          rfdkCertificate: serverCert,
+        },
+        internalProtocol: ApplicationProtocol.HTTPS,
+      },
+      version: this.version,
+      vpcSubnets: renderQueueSubnets,
+      vpcSubnetsAlb: renderQueueAlbSubnets,
+      deletionProtection: false,
```
jericht, line 158: Same comment as the python app's Render Queue about deletion protection, should we consider adding the `TODO` comment here?

jericht (changes requested):
Just some small comments, otherwise this looks good to me.

examples/deadline/Local-Zone/python/package/lib/service_tier.py
horsmand: I tried to trim down anything that the Basic example already explains to make it more obvious as to what we're changing here. I do have a comment in the Architecture section of the readme stating that the example assumes some familiarity with the general architecture of an RFDK farm and to refer to the Basic example if they are not.

examples/deadline/Local-Zone/python/README.md
horsmand: Yeah, I agree, but it seems a bit redundant to put it on this line and then give the full example. I've just removed it from the paragraph and left the example code.

examples/deadline/Local-Zone/python/README.md
horsmand: updated

examples/deadline/Local-Zone/python/package/lib/compute_tier.py
horsmand: Added it back in.

examples/deadline/Local-Zone/python/package/lib/network_tier.py
horsmand: Oh ya that doesn't make sense. Updated.

examples/deadline/Local-Zone/ts/README.md
horsmand: updated

examples/deadline/Local-Zone/ts/bin/app.ts
horsmand: updated

examples/deadline/Local-Zone/ts/lib/compute-tier.ts
horsmand: Whoops, I did that for testing. Updated.

examples/deadline/Local-Zone/ts/lib/compute-tier.ts
horsmand: I had it like this because for some reason the TS and Python implementations of the `InstanceClass` enum have different entries. I did a little more digging and it seems that `T3` is equivalent to `BURSTABLE3` which is in both implementations, so I'll change them both to use that.

examples/deadline/Local-Zone/python/package/lib/service_tier.py
jericht: Sounds good. The reference to the Basic example should be good enough.

examples/deadline/Local-Zone/ts/lib/compute-tier.ts
jericht: Oh weird, I didn't know they had didn't enum values. `BURSTABLE3` sounds good to me.

jericht (dismissed):

jericht (dismissed):

jericht (dismissed):

examples/deadline/Local-Zone/README.md
```diff
@@ -0,0 +1,39 @@
+# RFDK Sample Application - Local Zones
+
+If you have large asset files that your worker instances need to process, any bit of latency can have a big impact on the time your renders take. This example will walk you through setting up your workers in a local zone while leaving the rest of the render farm in standard availability zones. Currently Amazon has launched a local zone in Los Angeles that is a part of the us-west-2 region, but they have more on the way. For more information on where local zones are available, how to get access, and what services they provide, refer to the [AWS Local Zones about page](https://aws.amazon.com/about-aws/global-infrastructure/localzones/).
```
jusiskin: How would you feel about changing to:

>   If you have large asset files that your Worker instances need to access from your on-prem infrastructure, deploying the Workers to a geographically close [AWS Local Zone](https://aws.amazon.com/about-aws/global-infrastructure/localzones/) can reduce the latency and increase the speed of your renders.

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,60 @@
...
+    pip install ../../../../dist/python/aws-rfdk-<version>.tar.gz
+    ```
+
+4.  You must read and accept the [AWS Thinkbox End-User License Agreement (EULA)](https://www.awsthinkbox.com/end-user-license-agreement) to deploy and run Deadline. To do so, change the value of the `accept_aws_thinkbox_eula` in `package/lib/config.py` like this:
+
+    ```py
+    self.accept_aws_thinkbox_eula: AwsThinkboxEulaAcceptance = AwsThinkboxEulaAcceptance.USER_ACCEPTS_AWS_THINKBOX_EULA
+    ```
+
+5. Change the value of the `deadline_version` variable in `package/lib/config.py` to specify the desired version of Deadline to be deployed to your render farm. RFDK is compatible with Deadline versions 10.1.9.x and later. To see the available versions of Deadline, consult the [Deadline release notes](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/release-notes.html). It is recommended to use the latest version of Deadline available when building your farm, but to pin this version when the farm is ready for production use. For example, to pin to the `10.1.13` release of Deadline, use `10.1.13.2`.
```
jusiskin: This recommendation differs from the other examples where we say to specify the version up to the release component (e.g. `10.1.13`). In the `All-In-AWS-Basic`, we say:

>   For example, to pin to the latest `10.1.12.x` release of Deadline, use:
>
>   ```ts
>   public readonly deadlineVersion: string = '10.1.12';
>   ```

Same feedback applies to the TypeScript example

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+6. Change the value of the `deadline_client_linux_ami_map` variable in `package/config.py` to include the region + AMI ID mapping of your EC2 AMI(s) with Deadline Worker.
+
+7. Also in `package/lib/config.py`, you can set the `availability_zones_standard` and `availability_zones_local` values to the availability zones you want to use. These values must all be from the same region. It's recommended you use at least two standard zones, but you can use more if you'd like. For the local zones, you can use as many as you'd like.
```
jusiskin: nit: maybe we change the word "recommended" to "required that". I believe DocDB requires at least two AZs to be specified.

Same feedback applies to the TypeScript example

examples/deadline/Local-Zone/python/package/lib/network_tier.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        super().__init__(scope, stack_id, **kwargs)
+
+        # We're creating a SubnetSelection with only the standard availability zones
+        # to be used to put the NAT gateway in and the VPC endpoints, because the local zones do no have
+        # these available.
+        subnets = SubnetSelection(
+            availability_zones=config.availability_zones_standard,
+            subnet_type=SubnetType.PUBLIC
+        )
```
jusiskin: perhaps `nat_gateway_subnets` is a slightly more descriptive variable name here. what do you think?

examples/deadline/Local-Zone/python/package/lib/network_tier.py
```diff
@@ -0,0 +1,122 @@
...
+            ],
+            nat_gateway_subnets=subnets
+        )
+
+        # Add interface endpoints
+        for idx, service_info in enumerate(_INTERFACE_ENDPOINT_SERVICES):
+            service_name = service_info['name']
+            service = service_info['service']
+            self.vpc.add_interface_endpoint(
+                f'{service_name}{idx}',
```
jusiskin: nit: I realize this is copied from our other example, but I believe we should leave out `idx` from the construct ID here and for the gateway endpoints as well. The construct IDs are used to derive the logical IDs of resources in the CloudFormation template. See the note about **Logical ID stability** in the [CDK docs for logical IDs](https://docs.aws.amazon.com/cdk/latest/guide/identifiers.html#identifiers_logical_ids).

The risk here is that someone takes this example, deploys, and then later decides they want to add/remove one of the endpoints in `_INTERFACE_ENDPOINT_SERVICES`. All of the endpoints beneath it will get recreated by Cloud Formation because of the index cascade, causing an unintended transient outage during deployment.

Same feedback applies to the TypeScript example.

examples/deadline/Local-Zone/ts/bin/app.ts
```diff
@@ -0,0 +1,58 @@
...
+  acceptAwsThinkboxEula: config.acceptAwsThinkboxEula,
+});
+
+new ComputeTier(app, 'ComputeTier', {
+  env,
+  vpc: network.vpc,
+  availabilityZones: config.availabilityZonesLocal,
+  renderQueue: service.renderQueue,
+  workerMachineImage: MachineImage.genericLinux(config.deadlineClientLinuxAmiMap),
+  keyPairName: config.keyPairName ? config.keyPairName : undefined,
```
jusiskin: nit: can this be simply:

```ts
keyPairName: config.keyPairName,
```

examples/deadline/Local-Zone/README.md
```diff
@@ -0,0 +1,39 @@
+# RFDK Sample Application - Local Zones
```
jusiskin, line 1: This example does a great job of showing how to deploy the Workers to a local zone - great work!

I think one thing that might be missing though is an explanation of why this is important/useful. RFDK users want this so they can have a low-latency connection between their AWS infrastructure and their infrastructure hosted outside AWS. It might help to spend some time at the beginning of this README setting the stage here. It may also help to add some guidance, links, and next steps to point users in the right direction for how they'd take this example and proceed to the next step of connecting their infrastructure outside of AWS.

One good resource we could link to is our developer guide documentation on [connecting to a RFDK render farm](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html)

jusiskin (changes requested):
Great job! Just a few small proposed changes but otherwise this is solid

examples/deadline/Local-Zone/python/package/lib/network_tier.py
horsmand: This was used for the VPC endpoints as well, so I changed it to standard_zone_subnets. Let me know what you think

jericht (dismissed):

examples/deadline/Local-Zone/README.md
jusiskin: Looks good. Thanks!

examples/deadline/Local-Zone/python/package/lib/network_tier.py
jusiskin: That looks good. Resolved

examples/deadline/Local-Zone/README.md
jusiskin: Thanks for adding documentation for this in the top-level README.

As discussed offline, let's add a "next steps" section to the Python/TS READMEs that list the steps required to connect an on-premise networked file-system to the Workers. Even if we don't have all of the steps completely detailed, for now having a checklist would help guide readers to research and complete them. These are the steps that I foresee here:

1.  Follow the [connecting to your render farm](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html) dev guide page to modify the `NetworkTier` with the the VPN connection.
2. Ensure there is a route from the worker subnet to the on-prem file-system
3. Ensure the VPN endpoint's security group has ingress rules to allow NFS port traffic from the Workers
4. Add user-data to mount the NFS on the compute tier
5. (optional) set up path mapping rules in Deadline

jusiskin (dismissed):
Looks good. Thanks for all your hard work on this one!

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+    ```bash
+    aws --region <region> ec2 describe-images --owners 357466774442 --filters "Name=name,Values=*Worker*" "Name=name,Values=*<version>*" --query 'Images[*].[ImageId, Name]' --output text
+    ```
+
+7. Also in `package/lib/config.py`, you can set the `availability_zones_standard` and `availability_zones_local` values to the availability zones you want to use. These values must all be from the same region. It's required that you use at least two standard zones, but you can use more if you'd like. For the local zones, you can use one or more.
+
+8. To gain the benefits of putting your workers in a local zone close to your asset server, you are going to want to set up a connection from your local network to the one you're creating in AWS.
+   1. You should start by reading through the [Connecting to the Render Farm](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html) documentation and implemening one of the methods for connecting your network to your AWS VPC described there.
```
jericht:
```suggestion
   1. You should start by reading through the [Connecting to the Render Farm](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html) documentation and implementing one of the methods for connecting your network to your AWS VPC described there.
```

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,84 @@
...
+   2. With whichever option you choose, you'll want to make sure you are propagating the worker subnets to your local network. All the options in the document show how to propagate all the private subnets, which will include the ones used by the workers.
+   3. Ensure your worker fleet's security group allows traffic from your network on the correct ports that your NFS requires to be open. The documentation shows how to [allowing connections to the Render Queue](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html#allowing-connection-to-the-render-queue), which you may also want to enable if you plan on connecting any of your local machines to your render farm, but you would also want to do something similar for the worker fleet, for example, ports `22` and `2049` are commonly required for NFS, so this code could be added to the `ComputeTier`:
```
jericht:
```suggestion
   3. Ensure your worker fleet's security group allows traffic from your network on the correct ports that your NFS requires to be open. The documentation shows how to [allow connections to the Render Queue](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html#allowing-connection-to-the-render-queue), which you may also want to enable if you plan on connecting any of your local machines to your render farm, but you would also want to do something similar for the worker fleet, for example, ports `22` and `2049` are commonly required for NFS, so this code could be added to the `ComputeTier`:
```

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+    ```python
+    // The customer-prefix-cidr-range needs to be replaced by the CIDR range for your local network that you used when configuring the VPC connection
```
jericht:
```suggestion
    # The customer-prefix-cidr-range needs to be replaced by the CIDR range for your local network that you used when configuring the VPC connection
```

examples/deadline/Local-Zone/python/README.md
```diff
@@ -0,0 +1,84 @@
...
+    self.worker_fleet.connections.allow_from(Peer.ipv4('customer-prefix-cidr-range'), Port.tcp(22));
+    self.worker_fleet.connections.allow_from(Peer.ipv4('customer-prefix-cidr-range'), Port.udp(22));
+    self.worker_fleet.connections.allow_from(Peer.ipv4('customer-prefix-cidr-range'), Port.tcp(2049));
+    self.worker_fleet.connections.allow_from(Peer.ipv4('customer-prefix-cidr-range'), Port.tcp(2049));
+    ```
+
+   4. Add user-data to mount the NFS on the compute tier. This can be provided in the `UserDataProvider` in the `ComputeTier`.
+   5. (optional) set up path mapping rules in Deadline
```
jericht:
```suggestion
   5. (optional) Set up [path mapping rules in Deadline](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/cross-platform.html)
```

examples/deadline/Local-Zone/ts/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+    ```bash
+    aws --region <region> ec2 describe-images --owners 357466774442 --filters "Name=name,Values=*Worker*" "Name=name,Values=*<version>*" --query 'Images[*].[ImageId, Name]' --output text
+    ```
+
+5. Also in `bin/config.ts`, you can set the `availabilityZonesStandard` and `availabilityZonesLocal` values to the availability zones you want to use. These values must all be from the same region. It's required that you use at least two standard zones, but you can use more if you'd like. For the local zones, you can use one or more.
+
+6. To gain the benefits of putting your workers in a local zone close to your asset server, you are going to want to set up a connection from your local network to the one you're creating in AWS.
+   1. You should start by reading through the [Connecting to the Render Farm](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html) documentation and implemening one of the methods for connecting your network to your AWS VPC described there.
```
jericht:
```suggestion
   1. You should start by reading through the [Connecting to the Render Farm](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html) documentation and implementing one of the methods for connecting your network to your AWS VPC described there.
```

examples/deadline/Local-Zone/ts/README.md
```diff
@@ -0,0 +1,76 @@
...
+   2. With whichever option you choose, you'll want to make sure you are propagating the worker subnets to your local network. All the options in the document show how to propagate all the private subnets, which will include the ones used by the workers.
+   3. Ensure your worker fleet's security group allows traffic from your network on the correct ports that your NFS requires to be open. The documentation shows how to [allowing connections to the Render Queue](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html#allowing-connection-to-the-render-queue), which you may also want to enable if you plan on connecting any of your local machines to your render farm, but you would also want to do something similar for the worker fleet, for example, ports `22` and `2049` are commonly required for NFS, so this code could be added to the `ComputeTier`:
```
jericht:
```suggestion
   3. Ensure your worker fleet's security group allows traffic from your network on the correct ports that your NFS requires to be open. The documentation shows how to [allow connections to the Render Queue](https://docs.aws.amazon.com/rfdk/latest/guide/connecting-to-render-farm.html#allowing-connection-to-the-render-queue), which you may also want to enable if you plan on connecting any of your local machines to your render farm, but you would also want to do something similar for the worker fleet, for example, ports `22` and `2049` are commonly required for NFS, so this code could be added to the `ComputeTier`:
```

examples/deadline/Local-Zone/ts/README.md
```diff
@@ -0,0 +1,76 @@
...
+    ```ts
+    // The customer-prefix-cidr-range needs to be replaced by the CIDR range for your local network that you used when configuring the VPC connection
+    this.workerFleet.connections.allowFrom(Peer.ipv4('customer-prefix-cidr-range'), Port.tcp(22));
+    this.workerFleet.connections.allowFrom(Peer.ipv4('customer-prefix-cidr-range'), Port.udp(22));
+    this.workerFleet.connections.allowFrom(Peer.ipv4('customer-prefix-cidr-range'), Port.tcp(2049));
+    this.workerFleet.connections.allowFrom(Peer.ipv4('customer-prefix-cidr-range'), Port.tcp(2049));
+    ```
+
+   4. Add user-data to mount the NFS on the compute tier. This can be provided in the `UserDataProvider` in the `ComputeTier`.
+   5. (optional) set up path mapping rules in Deadline
```
jericht:
```suggestion
   5. (optional) Set up [path mapping rules in Deadline](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/cross-platform.html)
```

jericht (changes requested):
Just a few nits/typos, otherwise this looks great!

jericht (approved):
Looks good, thanks!

jusiskin (approved):

examples/deadline/Local-Zone/README.md
jusiskin: The use-case is now documented with some guidance on how to go about connecting infrastructure outside of AWS. Resolving

examples/deadline/Local-Zone/README.md
jusiskin: mention of Route53 is removed which seems fair. There is a link to the All-in-AWS-Infrastructure-Basic example which mentioned the Private Hosted Zone. Resolving

examples/deadline/Local-Zone/README.md
jusiskin: Bastion was removed. Resolving

examples/deadline/Local-Zone/python/README.md
jusiskin: This is now removed. Resolving

PR #319, opened 2021-02-19
feat(deadline): add security group configuration for Repository and RenderQueue

### LINKED ISSUE #307: Expand configuration options for Security Group configuration
It is not currently possible to provide/configure the Security Group for:
1. Repository -- https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.Repository.html
2. The RCS host instance in RenderQueue -- https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.RenderQueue.html

It would be grand to be able to have fine-grained control over those security groups.

### Use Case

Presently, the security groups created for these resources allow full-egress by default. Customers that are aiming for enhanced layers of network-level access controls have a need to set these security groups to deny all egress by default, and to explicitly add their own egress rules.

Customers may also have created their own security group that, say, controls access to VPC Interface Endpoints, and they need a means by which those security groups can be added to the Repository & RenderQueue's hosts.

### Proposed Solution

1. Property on the construct(s) that allow providing the security group that will be used. Note: RenderQueue should provide separate properties for the ALB & ECS host; the Connections object of the RenderQueue should remain as the ALB's SG, but there should also be an easy-access way to get to the RCS-host's SG.
2. Addition of an `addSecurityGroup()` method (ex: https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.WorkerInstanceFleet.html#add-wbr-security-wbr-groupsecuritygroup ) to these constructs that allows the customer to add additional security groups to the construct after creation. In the RenderQueue, there should be separate methods for the ALB & RCS SGs. Note: These additional security groups should not be added to the Connections object of the construct -- doing that would make any use of the construct's Connections object also change the added SGs.

### Other

N/A

---

jericht:
Fixes https://github.com/aws/aws-rfdk/issues/307

### Changes
- `RenderQueue`: Add `addSecurityGroups` method that adds security groups to the `AutoScalingGroup`, `LoadBalancer`, or both.
- `Repository`: Add `addSecurityGroup` method that adds security groups to the `DatabaseConnection`.
- `DatabaseConnection`
    - DocDB: Adds security groups to the `CfnDBCluster`.
    - MongoDB: Adds security groups to the MongoDB server.
- `MongoDbInstance`: Added an `addSecurityGroup` method to add security groups to the `AutoScalingGroup` of the `StaticPrivateIpServer`.

### Testing
- Added unit tests
- Manual end-to-end tests by modifying the All-In-AWS-Infrastructure-Basic typescript sample app
    - Added a new `SecurityGroup` in `StorageTier` and added it to the `Repository` construct in the `ServiceTier`.
    - Added a new `SecurityGroup` in `ServiceTier` and added it to the `RenderQueue` construct in the `ServiceTier`.
    - Verified that the new security groups get added to the correct resources:
        - `Repository` (DocDB): Added to the DocDB Cluster
        - `Repository` (MongoDB): Added to the AutoScalingGroup
        - `RenderQueue`: Added to the AutoScalingGroup and LoadBalancer
    - Test cases
        1. DocDB repository, pass security groups into constructors
        2. DocDB repository, use `addSecurityGroup(s)` methods to add security groups
        3. MongoDB repository, pass security groups into constructors
        4. MongoDB repository, use `addSecurityGroup(s)` methods to add security groups

----

### DIFF
--- a/packages/aws-rfdk/lib/core/lib/mongodb-instance.ts
+++ b/packages/aws-rfdk/lib/core/lib/mongodb-instance.ts
@@ -286,6 +286,12 @@ export interface IMongoDb extends IConnectable, IConstruct {
    * The version of MongoDB that is running on this instance.
    */
   readonly version: MongoDbVersion;
+
+  /**
+   * Adds security groups to the database.
+   * @param securityGroups The security groups to add.
+   */
+  addSecurityGroup(...securityGroups: ISecurityGroup[]): void;
 }
 
 /**
@@ -485,6 +491,13 @@ export class MongoDbInstance extends Construct implements IMongoDb, IGrantable {
 
     // Tag deployed resources with RFDK meta-data
     tagConstruct(this);
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public addSecurityGroup(...securityGroups: ISecurityGroup[]): void {
+    securityGroups?.forEach(securityGroup => this.server.autoscalingGroup.addSecurityGroup(securityGroup));
   }
 
   /**
--- a/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
@@ -7,10 +7,12 @@ import * as path from 'path';
 import {
   CfnDBCluster,
   CfnDBInstance,
+  DatabaseCluster,
   IDatabaseCluster,
 } from '@aws-cdk/aws-docdb';
 import {
   IConnectable,
+  ISecurityGroup,
   OperatingSystemType,
   Port,
 } from '@aws-cdk/aws-ec2';
@@ -155,6 +157,13 @@ export abstract class DatabaseConnection {
    * @param child The child to make dependent upon this database.
    */
   public abstract addChildDependency(child: IConstruct): void;
+
+  /**
+   * Adds a security group to the database.
+   *
+   * @param securityGroups The security group to add.
+   */
+  public abstract addSecurityGroup(...securityGroups: ISecurityGroup[]): void;
 }
 
 /**
@@ -268,6 +277,42 @@ class DocDBDatabaseConnection extends DatabaseConnection {
   }
 
   /**
+   * @inheritdoc
+   */
+  public addSecurityGroup(...securityGroups: ISecurityGroup[]): void {
+    let added = false;
+    const errorReasons: string[] = [];
+    if (this.props.database instanceof DatabaseCluster) {
+      const resource = (this.props.database as DatabaseCluster).node.tryFindChild('Resource');
+
+      // TODO: Replace this code with the addSecurityGroup method of DatabaseCluster once this PR is merged:
+      // https://github.com/aws/aws-cdk/pull/13290
+      if (resource instanceof CfnDBCluster) {
+        const cfnCluster = resource as CfnDBCluster;
+        const securityGroupIds = securityGroups.map(sg => sg.securityGroupId);
+
+        if (cfnCluster.vpcSecurityGroupIds === undefined) {
+          cfnCluster.vpcSecurityGroupIds = securityGroupIds;
+        } else {
+          cfnCluster.vpcSecurityGroupIds.push(...securityGroupIds);
+        }
+        added = true;
+      } else {
+        errorReasons.push('The internal implementation of AWS CDK\'s DocumentDB cluster construct has changed.');
+      }
+    } else {
+      errorReasons.push('The "database" property passed to this class is not an instance of AWS CDK\'s DocumentDB cluster construct.');
+    }
+
+    if (!added) {
+      Annotations.of(this.props.database).addWarning(
+        `Failed to add the following security groups to ${this.props.database.node.id}: ${securityGroups.map(sg => sg.node.id).join(', ')}. ` +
+        errorReasons.join(' '),
+      );
+    }
+  }
+
+  /**
    * Deadline is only compatible with MongoDB 3.6. This function attempts to determine whether
    * the given DocDB version is compatible.
    */
@@ -373,6 +418,13 @@ class MongoDbInstanceDatabaseConnection extends DatabaseConnection {
       const db = this.props.database as MongoDbInstance;
       child.node.addDependency(db.server.autoscalingGroup.node.defaultChild!);
     }
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public addSecurityGroup(...securityGroups: ISecurityGroup[]): void {
+    this.props.database.addSecurityGroup(...securityGroups);
   }
 
   /**
--- a/packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
@@ -213,9 +213,9 @@ export interface RenderQueueAccessLogProps {
 }
 
 /**
- * Options for security groups of the Render Queue.
+ * Security groups of the Render Queue.
  */
-export interface RenderQueueSecurityGroupsOptions {
+export interface RenderQueueSecurityGroups {
   /**
    * The security group for the backend components of the Render Queue, which consists of the AutoScalingGroup for the Deadline RCS.
    */
@@ -329,9 +329,11 @@ export interface RenderQueueProps {
   readonly deletionProtection?: boolean;
 
   /**
-   * Options to add additional security groups to the Render Queue.
+   * Security groups to use for the Render Queue.
+   *
+   * @default - new security groups are created
    */
-  readonly securityGroupsOptions?: RenderQueueSecurityGroupsOptions;
+  readonly securityGroups?: RenderQueueSecurityGroups;
 }
 
 /**
--- a/packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
@@ -8,6 +8,7 @@ import {
 } from '@aws-cdk/aws-certificatemanager';
 import {
   InstanceType,
+  ISecurityGroup,
   IVpc,
   SubnetSelection,
 } from '@aws-cdk/aws-ec2';
@@ -212,6 +213,20 @@ export interface RenderQueueAccessLogProps {
 }
 
 /**
+ * Options for security groups of the Render Queue.
+ */
+export interface RenderQueueSecurityGroupsOptions {
+  /**
+   * The security group for the backend components of the Render Queue, which consists of the AutoScalingGroup for the Deadline RCS.
+   */
+  readonly backend?: ISecurityGroup;
+  /**
+   * The security group for the frontend of the Render Queue, which is its load balancer.
+   */
+  readonly frontend?: ISecurityGroup;
+}
+
+/**
  * Properties for the Render Queue
  */
 export interface RenderQueueProps {
@@ -312,6 +327,11 @@ export interface RenderQueueProps {
    * @see https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#deletion-protection
    */
   readonly deletionProtection?: boolean;
+
+  /**
+   * Options to add additional security groups to the Render Queue.
+   */
+  readonly securityGroupsOptions?: RenderQueueSecurityGroupsOptions;
 }
 
 /**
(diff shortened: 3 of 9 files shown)


packages/aws-rfdk/lib/core/lib/mongodb-instance.ts
```diff
@@ -487,6 +493,13 @@ export class MongoDbInstance extends Construct implements IMongoDb, IGrantable {
     tagConstruct(this);
   }
 
+  /**
+   * @inheritdoc
+   */
+  public addSecurityGroup(...securityGroups: ISecurityGroup[]): void {
```
ddneilson, line 499: This is a good addition. Thanks!

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -267,6 +276,35 @@ class DocDBDatabaseConnection extends DatabaseConnection {
...
+      }
+    }
+
+    if (!added) {
+      Annotations.of(this.props.database).addWarning(
+        `Failed to add the following security groups to ${this.props.database.node.id}: ${securityGroups.map(sg => sg.node.id).join(', ')}. ` +
+        'This is because either the "database" property passed to this class is not an instance of AWS CDK\'s DocumentDB cluster construct or the ' +
+        'internal implementation of AWS CDK\'s DocumentDB cluster construct has changed.',
+      );
+    }
```
kozlove-aws, line 312: What do you say on adding separate annotation far every  case instead of general one?
I believe it will be more informative for user.

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -267,6 +276,35 @@ class DocDBDatabaseConnection extends DatabaseConnection {
...
   }
 
+  /**
+   * @inheritdoc
+   */
+  public addSecurityGroup(...securityGroups: ISecurityGroup[]): void {
+    let added = false;
+    if (this.props.database instanceof DatabaseCluster) {
+      const resource = (this.props.database as DatabaseCluster).node.findChild('Resource');
+      if (resource instanceof CfnDBCluster) {
```
ddneilson, line 290: This is clever. If you're willing, and have the time, it might be worth whipping up a quick PR for the CDK to add an `addSecurityGroup()` to the DocumentDB Cluster construct. Then, we could switch this code to use that once it becomes available.

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -211,6 +212,20 @@ export interface RenderQueueAccessLogProps {
...
 }
 
+/**
+ * Options for security groups of the `RenderQueue`.
+ */
+export interface RenderQueueSecurityGroupsOptions {
+  /**
+   * The `AutoScalingGroup` security groups.
+   */
+  readonly autoScalingGroup?: ISecurityGroup[];
```
ddneilson: 1) Maybe name these `backend` and `frontend` instead? `frontend` -> Security group for the load balancer. `backend` -> Security group for everything behind the load balancer. It's probable that we add the Deadline WebService to this same construct, and it makes some sense for it to have the same SG as the RCS.

2) The normal pattern in CDK constructs is to accept a single `ISecurityGroup` as a construct property, rather than an array, and then expose an `addSecurityGroups()` method. What do you think about following that pattern here?

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -312,6 +327,11 @@ export interface RenderQueueProps {
    * @see https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#deletion-protection
    */
   readonly deletionProtection?: boolean;
+
+  /**
+   * Options to add additional security groups to the `RenderQueue`.
+   */
+  readonly securityGroups?: RenderQueueSecurityGroupsOptions;
```
kozlove-aws: How about more informative variable name?
Like `securityGroupsOptions`.

kozlove-aws (commented):
Change looks good. Just a couple small remarks

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -448,6 +449,10 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
       });
     }
 
+    if (props.securityGroups) {
```
ddneilson: I suggest replacing this in favour of:

1. We'll want to modify the `const loadBalancer = new ApplicationLoadBalancer(this, 'LB', {` to pass the frontend security group, from the props, as the `securityGroup` [property](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-elasticloadbalancingv2.ApplicationLoadBalancer.html#securitygroup). All listeners created for the ALB will use the LoadBalancer's security groups.
2. Similarly, we'll want to modify the  `this.asg = this.cluster.addCapacity('RCS Capacity', {` to pass the backend security group. The properties of that call are just AutoScalingGroup properties -- you can see that here: https://github.com/aws/aws-cdk/blob/4553d4b37cc7c2062eebc36bfe7a03b0226b598d/packages/%40aws-cdk/aws-ecs/lib/cluster.ts#L226 .

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -519,6 +524,15 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
     child.node.addDependency(this.taskDefinition);
   }
 
+  public addSecurityGroups(securityGroups: RenderQueueSecurityGroupsOptions) {
```
ddneilson: I suggest splitting this in two, to make the interface simpler in Python:
1. `addFrontendSecurityGroups( array of SGs )` -> Add the security groups to the load balancer.
2. `addBackendSecurityGroups( array of SGs )` -> Add the security groups to the backend components (currently just the RCS's ASG).

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -339,6 +340,11 @@ export interface RepositoryProps {
    * @default Duration.days(15) for the database
    */
   readonly backupOptions?: RepositoryBackupOptions;
+
+  /**
+   * The security groups to add to the database.
+   */
+  readonly securityGroups?: ISecurityGroup[];
```
ddneilson: Thinking a little out loud here ... There are three components in the Repository that need security groups.
1. EFS
2. DocDB
3. Repository installer instance/ASG.

It makes a certain amount of sense to expose the ability to provide a single SG for each of those; similar to the suggestion for the RenderQueue. Then, if we make sure that there are public object members for each of those things, the user can use the EFS/Cluster/ASG construct's own `addSecurityGroup()` and `IConnectable` interface to add/modify security group settings.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -535,6 +541,10 @@ export class Repository extends Construct implements IRepository {
       });
     }
 
+    if (props.securityGroups) {
```
ddneilson: Then, here, similar to the RenderQueue. Replace this with:

1. Passing the filesystem SG as a [property](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-efs.FileSystem.html#securitygroupspan-classapi-icon-api-icon-experimental-titlethis-api-element-is-experimental-it-may-change-without-noticespan) when creating the EFS.
2. Passing the database SG as a [property](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-docdb.InstanceProps.html#securitygroupspan-classapi-icon-api-icon-experimental-titlethis-api-element-is-experimental-it-may-change-without-noticespan) when creating the DocDB.
3. Passing the installer's SG as a property when creating the ASG.

ddneilson (changes requested):

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
jericht: Sure, updated

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
jericht: Sounds good, updated

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
jericht: Updated

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: Sounds good. I've moved the frontend security group into the ALB constructor. The `addCapacity` method doesn't look like it "officially" takes in a security group, but based on the code you linked it looks safe to force it in the options by using `@ts-ignore`, which is what I've put for now.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: Sure, split this up into two

packages/aws-rfdk/lib/deadline/lib/repository.ts
jericht: Good idea, updated.

Exposing the EFS was a bit tricky since we have the `IMountableLinuxFilesystem` interface that, I'm assuming, allows users to pass in their own implementation of it using any filesystem they want, so we can't be sure that the underlying filesystem is EFS. The best thing I could think of was exposing an "optional" member for the EFS that is only defined if the Repository creates its own, else `undefined` (which should be fine since that means the user already has a reference to their file system object).

packages/aws-rfdk/lib/deadline/lib/repository.ts
jericht: Updated

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
jericht: Good idea, made a PR: https://github.com/aws/aws-cdk/pull/13290

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -2271,6 +2274,120 @@ describe('RenderQueue', () => {
...
+      // WHEN
+      renderQueue.addFrontendSecurityGroups(frontendSecurityGroup);
+
+      // THEN
+      assertSecurityGroupsWereAdded({
+        frontend: frontendSecurityGroup,
+      });
+    });
+
+    test('security groups added post-construction are not attached to Connections object', () => {
```
ddneilson, line 2343: This is great. Good idea adding this test.

ddneilson (dismissed):
This is great work, Jericho. Thanks for the contribution!

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -211,6 +212,20 @@ export interface RenderQueueAccessLogProps {
...
+export interface RenderQueueSecurityGroupsOptions {
+  /**
+   * The security group for the backend components of the Render Queue, which consists of the AutoScalingGroup for the Deadline RCS.
+   */
+  readonly backend?: ISecurityGroup;
+  /**
+   * The security group for the frontend of the Render Queue, which is its load balancer.
+   */
+  readonly frontend?: ISecurityGroup;
+}
```
kozlove-aws, line 227: Now it is not options any more. I can see possible two security groups that should be added.
Is any reason we are limited with only one security group for backend and frontend?

jericht: It's mostly to follow the pattern that CDK has established to take in a [single security group in the constructor](https://github.com/aws/aws-cdk/blob/ea91aa31db9e2f31c734ad6d7e1f64d5d432dfd4/packages/%40aws-cdk/aws-ec2/lib/instance.ts#L122) and have a [method to add more](https://github.com/aws/aws-cdk/blob/ea91aa31db9e2f31c734ad6d7e1f64d5d432dfd4/packages/%40aws-cdk/aws-ec2/lib/instance.ts#L418), as Daniel pointed out earlier.

kozlove-aws (approved):
Looks good for me.

ddneilson (approved):

PR #338, opened 2021-03-09
feat(core): make cloudwatch agent installation optional

### LINKED ISSUE #308: Modify behavior of CloudWatchAgent agent installation
Many RFDK constructs use the CloudWatchAgent construct ( https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.CloudWatchAgent.html ) by default. This construct always ensures that the CloudWatchAgent is installed (either pre-installed, or by installing it itself), and will fail the UserData script if it cannot be installed.

I suggest a modification to the scripting to not always install the agent if it is not present.

Note: The instance initialization (ex: cloud-init) can be viewed in the EC2 console, or API, by querying the instance's syslog; on Linux, at least.

### Use Case

The S3 method of installing the CloudWatchAgent requires that the host be able to reach the us-east-1 endpoints of S3, as that is where the bucket containing the agent is located. If deploying to any region other than us-east-1, this necessitates public internet access (tcp port 443) for any construct that uses the RFDK's CloudWatchAgent.

Using `yum install` on an Amazon Linux 2 instance only requires port 80 access to the regional S3 endpoints, which can be obtained, securely, through the use of a VPC Gateway Endpoint for S3 in the VPC.

For Windows, the current CloudWatchAgent construct also requires public internet access to fetch the gpg program. This makes it impossible to deploy an RFDK worker to an isolated subnet, as required by some security standards in the industry.

The current construct also make it impossible for a customer to use Linux distributions that do not support RPM.

### Proposed Solution

I suggest a change of behavior to:
1. Only try to install the agent on Amazon Linux 2 ( this will preserve functionality for the Repository ), but only do the install via `yum` (check for whether we're running on AL2 before trying to install); do not install the agent using the S3
2. Do not fail the script if CloudWatchAgent is not present, or cannot be installed.
3. Document, clearly, in the CloudWatchAgent construct and all RFDK constructs that use it, that customers should pre-install the CloudWatchAgent on their AMI, or use AL2 with access to regional S3 on port 80, if they want to enable log streaming functionality on their host(s).

### Other

---

horsmand:
Fixes https://github.com/aws/aws-rfdk/issues/308

Testing was a bit tricky to go through all the different code paths. I extracted the powershell and bash scripts and modified them slightly to work on their own, then manually spun up Windows and Linux hosts through the EC2 console using the scripts as user data. Doing this I was able to make modifications like setting the inputs to different values to using bad S3 locations I knew it wouldn't be able to download from, to see how it worked through different code paths. After I could use the scripts successfully like that, I then deployed the Basic example app, modified to have both a Linux and Windows worker fleet, to make sure that the unmodified scripts worked.

----

### DIFF
--- a/THIRD-PARTY
+++ b/THIRD-PARTY
@@ -1,6 +1,7 @@
 ** aws-cdk -- https://github.com/aws/aws-cdk
 ** jsii -- https://github.com/aws/jsii
-** MongoDB NodeJs Driver -- https://github.com/mongodb/node-mongodb-native
+** MongoDB NodeJs Driver --
+https://github.com/mongodb/node-mongodb-native
 
 Apache License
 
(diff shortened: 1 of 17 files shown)


THIRD-PARTY
```diff
@@ -1,6 +1,7 @@
-** aws-cdk -- https://github.com/aws/aws-cdk
-** jsii -- https://github.com/aws/jsii
-** MongoDB NodeJs Driver -- https://github.com/mongodb/node-mongodb-native
+** aws-cdk; version 1.52.0 -- https://github.com/aws/aws-cdk
```
ddneilson: Would be nice to remove the version numbers from these. They change almost every RFDK release, and we don't keep them up-to-date.

THIRD-PARTY
```diff
@@ -213,8 +214,622 @@ limitations under the License.
 
 ------
 
-** openssl -- https://www.openssl.org/
-Copyright (c) 1998-2021 The OpenSSL Project
+** GnuPG; version 2.2.27 -- https://gnupg.org/
```
ddneilson, line 217: Is it reasonable to remove the version number here as well?

THIRD-PARTY
```diff
@@ -213,8 +214,622 @@ limitations under the License.
...
+The GNU General Public License does not permit incorporating your program into
+proprietary programs. If your program is a subroutine library, you may consider
+it more useful to permit linking proprietary applications with the library. If
+this is what you want to do, use the GNU Lesser General Public License instead
+of this License. But first, please read
+<https://www.gnu.org/licenses/why-not-lgpl.html>.
+
+------
+
+** openssl 1.0.2k-fips; version 1.0.2k-fips -- https://www.openssl.org/
```
ddneilson: Ditto here for version number

packages/aws-rfdk/lib/core/lib/cloudwatch-agent.ts
```diff
@@ -116,16 +135,29 @@ export class CloudWatchAgent extends Construct {
    * @param host The host to configure the CloudWatch agent on
    * @param skipValidation Skips the validation of the CloudWatch agent installer if set to true.
```
ddneilson, line 137: nit: docstring needs updating,

ddneilson (changes requested):
Just some small comments. 

Can you elaborate on how this was tested?

THIRD-PARTY
horsmand: Since we need to link to the source code (I had to upload the original source to https://amazon-source-code-downloads.s3.amazonaws.com/aws/thinkbox/gnupg/gnupg-2.2.27.tar.bz2 as a part of the distribution request, this was a requirement for GPL) I think we should probably leave this one as-is. I will write up a short wiki or maybe a README alongside the script I wrote to upload the GPG installer to our buckets in the RFDK repository that explains the update process.

packages/aws-rfdk/lib/core/lib/cloudwatch-agent.ts
horsmand: Updated.

THIRD-PARTY
horsmand: Updated.

THIRD-PARTY
horsmand: Updated.

THIRD-PARTY
ddneilson: Sounds good.

ddneilson (approved):

packages/aws-rfdk/lib/core/lib/cloudwatch-agent.ts
```diff
@@ -114,18 +133,32 @@ export class CloudWatchAgent extends Construct {
...
    */
-  private configure(host: IScriptHost, skipValidation: boolean) {
+  private configure(
+    host: IScriptHost,
+    shouldInstallAgent: boolean,
+    skipValidation: boolean,
+  ) {
     // Grant access to the required CloudWatch Agent installer files
     const cloudWatchAgentBucket = Bucket.fromBucketArn(this, 'CloudWatchAgentBucket', 'arn:aws:s3:::amazoncloudwatch-agent');
```
ddneilson: Just noticed that we missed updating this. The bucket ARN changes based on the region now, no?

ddneilson (changes requested):

packages/aws-rfdk/lib/core/lib/cloudwatch-agent.ts
horsmand: Yes good catch, this was definitely a miss in my testing. When running the script through the RFDK example app, I was using the Thinkbox Deadline Linux and Windows base AMI's, which both have the CloudWatch agent pre-installed.

Since I don't have an AMI with Deadline installed but the CloudWatch agent not installed, I commented out the check for an existing agent in both configuration scripts and then turned off the worker fleet's health check so I could deploy the fleets and confirm the failure by logging into the hosts through the Systems Manager and checking the logs. After that, I updated this code to grant permissions for the regional bucket and redeployed to confirm that it was working.

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
```diff
@@ -3,68 +3,83 @@
...
-    [Parameter(Mandatory=$True)] $ssmParameterName,
+  [Parameter(Mandatory=$True)] $region,
+  [Parameter(Mandatory=$True)] $ssmParameterName,
 
-    # This parameter name intentionally does not follow the standard PascalCase style used in Powershell
-    # so that in our Typescript code, we can pass the same flag regardless of the script being called
-    [switch] $s = $False
+  # These parameter names intentionally do not follow the standard PascalCase style used in Powershell
+  # so that in our Typescript code, we can pass the same flag regardless of the script being called
+  [switch] $i = $True,
```
jericht: We have the default value for this flag as false in the bash script. Should we change the default value here to false as well?

jericht (changes requested):
Just one small comment, otherwise this looks good to me!

packages/aws-rfdk/lib/core/scripts/powershell/configureCloudWatchAgent.ps1
horsmand: Yes that should be false, I've updated it.

ddneilson (dismissed):

ddneilson (approved):

jericht (approved):

PR #339, opened 2021-03-09
feat: add ability to use EFS access points

jusiskin:
## Summary

This change modifies the `MountableEfs` class to accept an optional `accessPoint` property. When specified, the construct will generate user data that mounts the EFS file-system using the access point.

This allows the process accessing the file-system to use a UID/GID different than what UID/GID of the user running the process on the file-system mount.

## Additional Changes

The AWS-all-in-basic examples were modified to use EFS access points.

## Testing

*   Both of the Python and TypeScript examples were tested with Deadline Docker recipes where the container UID/GID would not have POSIX permissions to the EFS without an access point. When using EFS access points, the container is able to read/write to the EFS as expected.
*   Ran the RFDK integration tests on this branch and they all passed

----

### DIFF
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
@@ -30,7 +30,7 @@ from aws_cdk.aws_route53 import (
 
 from aws_rfdk import (
     DistinguishedName,
-    IMountableLinuxFilesystem,
+    MountableEfs,
     SessionManagerHelper,
     X509CertificatePem
 )
@@ -58,8 +58,8 @@ class ServiceTierProps(StackProps):
     vpc: IVpc
     # The database to connect to.
     database: DatabaseConnection
-    # The file system to install Deadline Repository to.
-    file_system: IMountableLinuxFilesystem
+    # The file-system to install Deadline Repository to.
+    mountable_file_system: MountableEfs
     # The ARN of the secret containing the UBL certificates .zip file (in binary form).
     ubl_certs_secret_arn: typing.Optional[str]
     # The UBL licenses to configure
@@ -109,9 +109,9 @@ class ServiceTier(Stack):
             ]
         )
 
-        # Granting the bastion access to the file system mount for convenience.
-        # This can also safely be removed.
-        props.file_system.mount_to_linux_instance(
+        # Mounting the root of the EFS file-system to the bastion access for convenience.
+        # This can safely be removed.
+        MountableEfs(self, filesystem=props.mountable_file_system.file_system).mount_to_linux_instance(
             self.bastion.instance,
             location='/mnt/efs'
         )
@@ -127,8 +127,9 @@ class ServiceTier(Stack):
             'Repository',
             vpc=props.vpc,
             database=props.database,
-            file_system=props.file_system,
+            file_system=props.mountable_file_system,
             repository_installation_timeout=Duration.minutes(20),
+            repository_installation_prefix='/',
             version=self.version
         )
 
--- a/packages/aws-rfdk/docs/upgrade/upgrading-0.27.md
+++ b/packages/aws-rfdk/docs/upgrade/upgrading-0.27.md
@@ -1,0 +1,272 @@
+# Upgrading to RFDK v0.27.x or Newer
+
+The Deadline container images used for server components (`RenderQueue` and `UsageBasedLicensing` constructs) have
+changed and as a result, [EFS access points][efs_access_points] must now be used to properly mount Deadline's Repository
+file-system. To upgrade without disruption, you must make changes to your CDK application and/or deployed EFS
+file-system.
+
+The upgrade process differs based on your usage pattern of the `Repository` RFDK construct. These usage patterns are
+documented in the following sections. Once you've identified which pattern applies to your CDK applicaiton, review the
+steps and plan your upgrade as documented.
+
+Scenarios:
+
+1.  [Deadline Repository with Implicit EFS File-System](#Deadline-Repository-with-Implicit-EFS-File-System)
+
+    **TypeScript:**
+    ```ts
+    new Repository(scope, 'Repository', {
+        // ...
+        // no "fileSystem" property
+        // ...
+    })
+    ```
+
+    **Python:**
+    ```py
+    Repository(scope, 'Repository',
+        # ...
+        # no "file_system" property
+        # ...
+    )
+    ```
+
+2.  [Deadline Repository with Provided EFS File-System](#Deadline-Repository-with-Provided-EFS-File-System)
+
+    **TypeScript:**
+    ```ts
+    new Repository(scope, 'Repository', {
+        // ...
+        fileSystem: myFileSystem
+        // ...
+    });
+    ```
+
+    **Python:**
+    ```py
+    Repository(scope, 'Repository',
+        # ...
+        file_system=my_file_system,
+        # ...
+    )
+    ```
+
+## Deadline Repository with Implicit EFS File-System
+
+If your app creates a `Repository` instance without a supplied file-system, then the `Repository` construct creates an
+EFS file-system for the Deadline Repository files.
+
+In RFDK 0.27.x, the `Repository` construct's behavior has changed in this scenario. It now creates and uses an EFS
+Access Point. Both the `Repository` and the `RenderQueue` mount the EFS using this new access point. Each EFS access
+point corresponds to a POSIX UID/GID combination. All file operations performed via a mount of the access point assume
+those UID/GIDs. For the above CDK code snippets, when no file-system is supplied to the `Repository` construct, it now
+creates an Access Point with:
+
+*   **UID**: `0`
+*   **GID**: `0`
+
+In effect, this gives the `Repository` and `RenderQueue` unrestricted access to the EFS file-system. This usage pattern
+is intended to abstract and encapsulate the file-system.
+
+**If you intend to share an EFS file-system with other parts of your infrastructure beyond the `Repository` and
+`RenderQueue` constructs**, you should instead create an EFS `FileSystem` construct in your CDK app and supply it to the
+`Repository` construct with a secured EFS Access Point. For examples of this, have a look at the newly updated example
+code for [TypeScript][access_point_ex_ts] and [Python][access_point_ex_py].
+
+**If you only require the file-system to be accessed by the `Repository` and `RenderQueue`**, then no further action is
+required and you can safely upgrade RFDK and deploy without disruption.
+
+## Deadline Repository with Provided EFS File-System
+
+In this scenario, your app creates a `Repository` instance with an explicitly specified file-system. Here are the steps
+to upgrade from pre-0.27.x to 0.27.x and higher:
+
+### Before Upgrading RFDK
+
+1.  Decide UID and GID values that you want to to be used by the `Repository` and `RenderQueue` constructs when
+    accessing EFS file-system for the Deadline Repository.
+    
+    If you do not have any requirement to share the EFS file-system with other agents in your infrastructure, these can
+    both be set to `0`. Note, that this grants the `Repository` and `RenderQueue` unrestricted access to the EFS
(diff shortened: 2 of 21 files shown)


packages/aws-rfdk/lib/core/lib/mountable-filesystem.ts
```diff
@@ -14,7 +17,7 @@ import { IScriptHost } from './script-assets';
  * {@link https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.Instance.html|EC2 Instance}
  * or an {@link https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-autoscaling.AutoScalingGroup.html|EC2 Auto Scaling Group}
  */
-export interface IMountingInstance extends IConnectable, IScriptHost {
+export interface IMountingInstance extends IConnectable, IConstruct, IScriptHost {
```
ddneilson: How come we needed the addition of `IConstruct`? I don't see any additions to the `MountableEfs` that would, I think, require it

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -488,22 +488,35 @@ export class Repository extends Construct implements IRepository {
...
-        filesystem: this.efs,
+
+      const accessPoint = fs.addAccessPoint('AccessPoint', {
+        path: '/DeadlineRepository',
+        createAcl: {
+          ownerUid: '10000',
+          ownerGid: '10000',
+          permissions: '770',
+        },
+        posixUser: {
```
ddneilson, line 510: This will break any user that just used the Repository, without passing in an EFS, correct?

ddneilson (changes requested):

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
```diff
@@ -109,9 +109,9 @@ def __init__(self, scope: Construct, stack_id: str, *, props: ServiceTierProps,
             ]
         )
 
-        # Granting the bastion access to the file system mount for convenience.
-        # This can also safely be removed.
-        props.file_system.mount_to_linux_instance(
+        # Mounting the root of the EFS file-system to the bastion access for convenience.
+        # This can safely be removed.
+        MountableEfs(self, filesystem=props.mountable_file_system.file_system).mount_to_linux_instance(
```
jericht, line 114: Can we use the `MountableEfs` in the props as is?
```
props.mountable_file_system.mount_to_linux_instance(...)
```

packages/aws-rfdk/lib/core/lib/mountable-filesystem.ts
jericht: I believe it's for the [node dependency addition](https://github.com/aws/aws-rfdk/pull/339/files#diff-4add095eaf52008f1b5fb6476d138899d4583b2dc36b57bddd9e223a1ea22fd8R95)
```
target.node.addDependency(this.props.filesystem.mountTargetsAvailable);
```

jusiskin: I snuck in a change to ensure that EFS mount targets are available before mounting the EFS. See `packages/aws-rfdk/lib/core/lib/mountable-efs.ts`:

```ts
target.node.addDependency(this.props.filesystem.mountTargetsAvailable);
```

According to [the `FileSystem.mountTargetsAvailable` API docs](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-efs.FileSystem.html#mounttargetsavailablespan-classapi-icon-api-icon-experimental-titlethis-api-element-is-experimental-it-may-change-without-noticespan):

>   Dependable that can be depended upon to ensure the mount targets of the filesystem are ready.

Seems like we should have this for cases when the EFS is in the same stack as the host where it is mounted. Without it, the CloudFormation deployment might have a race condition between the mount targets and the host instance since the host's user data only references the file-system itself. I don't think this further constrains IMountingInstance, since we cannot add user data to a non-construct resource.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -488,22 +488,35 @@ export class Repository extends Construct implements IRepository {
...
+          uid: '10000',
+          gid: '10000',
+        },
       });
-    }
+
+      return new MountableEfs(this, {
+        filesystem: fs,
+        accessPoint,
+        extraMountOptions: [`accesspoint=${accessPoint.accessPointId}`],
```
jericht: Looks like we can remove this since we're adding it in `MountableEfs` already: https://github.com/aws/aws-rfdk/pull/339/files#diff-4add095eaf52008f1b5fb6476d138899d4583b2dc36b57bddd9e223a1ea22fd8R123-R128

jericht (changes requested):

packages/aws-rfdk/lib/deadline/lib/repository.ts
jusiskin: Good catch. I changed this to use root (UID/GID 0). This should be okay when no file-system is supplied to the construct. In such cases, RFDK users don't care about the underlying file-system and file-system POSIX access controls. I've added relevant documentation to make this discoverable.

Let me know what you think.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
jusiskin: I've made this mount the root the file-system so customers can see everything and not be restricted to the lens of the access point.

Do you think we should change that though?

packages/aws-rfdk/lib/deadline/lib/repository.ts
jusiskin: Good catch. removed

packages/aws-rfdk/lib/core/lib/mountable-filesystem.ts
ddneilson: Gotcha. Could the dependency code be conditional? If the thing isn't a construct, then there's no need for a dep.
e.g. Something like
```
target.node?.addDependency(....)
```

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
jericht: Ah, no. I missed that this was mounting to the bastion instance, so I think it's fine.

packages/aws-rfdk/lib/core/lib/mountable-efs.ts
```diff
@@ -75,6 +98,23 @@ export class MountableEfs implements IMountableLinuxFilesystem {
       throw new Error('Target instance must be Linux.');
     }
 
+    target.node.addDependency(this.props.filesystem.mountTargetsAvailable);
+
+    if (this.props.accessPoint) {
+      const grantActions = MountPermissionsHelper.toEfsIAMActions(mount?.permissions);
+      target.grantPrincipal.addToPrincipalPolicy(new PolicyStatement({
```
ddneilson, line 124: Looks good. Locks down to the specific EFS and Access Point ARN.

packages/aws-rfdk/lib/core/lib/mountable-efs.ts
```diff
@@ -75,6 +98,23 @@ export class MountableEfs implements IMountableLinuxFilesystem {
       throw new Error('Target instance must be Linux.');
     }
 
+    target.node.addDependency(this.props.filesystem.mountTargetsAvailable);
```
ddneilson: Does this work if it's `target.node?.` ?  Or, alternatively, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty ?

Then we don't have to change the definition of `IMountingInstance`

packages/aws-rfdk/docs/upgrade/upgrading-0.27.md
```diff
@@ -0,0 +1,272 @@
...
+
+# This access point demonstrates the least-effort migration. It grants the
+# Repository and RenderQueue with root access to the entire EFS file-system.
+# If you intend to use the EFS file-system elsewhere, you may desired to
+# restrict this access further.
+AccessPoint(scope, 'AccessPoint',
+    file_system=file_system,
+    path='/',
+    posix_user=PosixUser(
+        uid='10000',
```
jericht:
```suggestion
        uid='0',
```

packages/aws-rfdk/docs/upgrade/upgrading-0.27.md
```diff
@@ -0,0 +1,272 @@
...
+        gid='10000',
```
jericht:
```suggestion
        gid='0',
```

ddneilson (changes requested):

jericht (changes requested):
Just one inconsistency in an upgrade doc example, otherwise it looks good to me!

packages/aws-rfdk/lib/core/lib/mountable-efs.ts
```diff
@@ -86,8 +126,14 @@ export class MountableEfs implements IMountableLinuxFilesystem {
 
     const mountDir: string = path.posix.normalize(mount.location);
     const mountOptions: string[] = [ MountPermissionsHelper.toLinuxMountOption(mount.permissions) ];
+    if (this.props.accessPoint) {
+      mountOptions.push(
```
ddneilson, line 149: Does passing these to the nfsv4 mount helper cause a break?

Minimally, there should be a warning on the MountableEfs's access point property says that the option is only available when mounting onto instances with the EFS mount helper available (e.g. Amazon Linux 2). Is the EFS helper available in any other distro's package repo?

ddneilson: Addition -- Found: https://docs.aws.amazon.com/efs/latest/ug/installing-amazon-efs-utils.html#installing-other-distro

Access points for other distros can be a followup if it's asked for, imo. Restricting to AL2 only for now seems reasonable.

Or, calling out that for non-AL2 distros the user should ensure that the EFS mount helper is pre-installed onto the AMI.

packages/aws-rfdk/lib/core/lib/mountable-efs.ts
jusiskin: I think `MountableEfs` already requires a construct, unless I'm mistaken. It needs to be able to modify the `IMountingInstance`'s user data, right? On that basis, there *should* be no harm in this addition. Perhaps I'm missing some corner case though...

ddneilson: I remember there was a reason that I didn't put `IConstruct` in the original definition. I don't recall what that reason was... some corner case.

jusiskin: Looks like I should be able to use [`Construct.isConstruct(...)`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Construct.html#static-iswbrconstructx). Changing to that....

packages/aws-rfdk/lib/core/lib/mountable-efs.ts
jusiskin: Good point here. We are unable to detect the target Linux distro at synthesis time. One idea is we could add an optional boolean to indicate AL2. If omitted or false, we could raise the warning. Using it would look like:

```ts
new MountableEfs(..., {
  // ...
}).mountToLinuxInstance(host, {
  // ...
  isAmazonLinux2: true,
});
```

We could use this API to dismiss CDK warnings when not known and hard-fail at synthesis when known and false. What do you think?

ddneilson: Discussed offline... decided not to do something like the `isAmazonLinux2` property.

ddneilson (approved):
Looks great. Thanks Josh!

jericht (approved):

PR #365, opened 2021-03-26
feat(core): add PadEfsStorage construct

### LINKED ISSUE #364: Provide construct to increase the baseline throughput of a bursting EFS filesystem
Add a mechanism to the RFDK by which the size of an Amazon EFS, in bursting-throughput mode, can have files added to it for the purpose of increasing the size of the filesystem to increase the baseline throughput available.

### Use Case

An Amazon EFS in bursting-throughput mode has baseline throughput directly proportional to the amount of data in the filesystem (50KB/s per 1GB stored in the filesystem). A limited pool of burst credits are consumed when exceeding that baseline throughput; credits are accrued when throughput is below the baseline. When credits are exhausted the ability to burst throughput is disabled until a positive credit balance is reached.

By default, via the Deadline Repository construct, and in our examples we deploy an Amazon EFS in bursting-throughput mode for Deadline's Repository filesystem; this stores files that Deadline needs to operate like the plugin files, and also acts as the place where job reports and such are stored. The amount of data in this filesystem is naturally very small -- a fresh install is about 1GB, and it grows fairly slowly as the render farm is utilized. It is possible for a customer's render farm to become essentially inoperable when burst credits are exhausted.

---

ddneilson:
Implements: https://github.com/aws/aws-rfdk/issues/364

#### Testing

1. Deployed a simple app with just an EFS, PadEfsStorage, and a bastion instance with the EFS mounted. Deployed a couple of times with desired sizes of 5 and 55. Verified that the step function executes and adds the correct number of files to the EFS.
2. `cdk synth` on both modified examples.

----


packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
```diff
@@ -0,0 +1,316 @@
...
+   * be added/removed from the root directory of the given access point.
+   * We strongly recommend that you provide an access point that is for a dedicated padding-files
+   * directory in your EFS filesystem, rather than the root directory or some other in-use directory
+   * of the filesystem.
+   */
+  readonly accessPoint: IAccessPoint;
+  /**
+   * Security group for the AWS Lambdas created by this construct.
+   *
+   * @default Security group with full egress and no ingress will be automaticaally created for each Lambda.
```
jericht:
```suggestion
   * @default Security group no egress or ingress will be automatically created for each Lambda.
```

Looks like we are setting `allowAllOutbound: false` in the default security group below

packages/aws-rfdk/lib/core/test/pad-efs-storage.test.ts
```diff
@@ -0,0 +1,204 @@
...
+    // THEN
+    cdkExpect(stack).to(haveResourceLike('AWS::Lambda::Function', {
+      FileSystemConfigs: [
+        {
+          Arn: stack.resolve(accessPoint.accessPointArn),
+          LocalMountPath: '/mnt/efs',
+        },
+      ],
+      Handler: 'pad-efs-storage.getDiskUsage',
+      Runtime: 'nodejs14.x',
```
jericht, line 79: Should we assert the `Timeout` here as well, like we do for the padFilesystem lambda?

packages/aws-rfdk/lib/core/test/pad-efs-storage.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      },
+    }));
+  });
+
+  test('Set desiredPadding', () => {
+    // WHEN
+    new PadEfsStorage(stack, 'PadEfs', {
+      vpc,
+      accessPoint,
+      desiredPaddingGB: 200,
```
jericht: nit: Should we move the `desiredPaddingGB` to a variable and have the assertions reference the variable?
```
// GIVEN
const desiredPaddingGb = 200;

// WHEN
// Create PadEfsStorage

// THEN
// reference desiredPaddingGB in the assertions
```

packages/aws-rfdk/lib/core/test/pad-efs-storage.test.ts
```diff
@@ -0,0 +1,204 @@
...
+          '',
+          arrayWith('","input":"{\\"desiredPadding\\":200}"}}'),
+        ],
+      },
+    }));
+  });
+
+  test('Provide SecurityGroup', () => {
+    // WHEN
+    const sg = new SecurityGroup(stack, 'TestSG', {
```
jericht, line 297: nit: Seems like the security group should be under a `// GIVEN` "block"

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -506,6 +514,22 @@ export class Repository extends Construct implements IRepository {
         securityGroup: props.securityGroupsOptions?.fileSystem,
       });
 
+      const paddingAccess = fs.addAccessPoint('PaddingAccessPoint', {
+        createAcl: {
+          ownerGid: '0',
+          ownerUid: '0',
+          permissions: '744',
+        },
+        path: '/PaddingFiles',
```
jericht: nit: Should we rename this path with an `RFDK` prefix? It would make it more obvious that the files in this path are created by RFDK.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -182,21 +197,21 @@ test('repository installer honors vpcSubnet', () => {
   // private subnets.
 
   // WHEN
-  const publicSubnetIds = [ 'PublicSubnet1', 'PublicSubnet2' ];
+  const isolatedSubnetIds = [ 'PublicSubnet1', 'PublicSubnet2' ];
```
jericht: nit: Should we rename the IDs to `IsolatedSubnet#` as well?

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/handlers.ts
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Add numbered files (e.g. 00000, 00001) of a given size to a filesystem.
+ * Note: exported so that we can test it.
+ * @param numFilesToAdd How many numbered files to add.
+ * @param filesize Size, in MB, of the files to create.
+ * @param mountPoint Directory in which to create the directory.
+ */
+export async function growFilesystem(numFilesToAdd: number, filesize: number, mountPoint: string): Promise<void> {
+  // We find the highest numbered file created thus far, and start adding numbered files after it.
+  var filename: string = await determineNextSequentialFilename(mountPoint);
```
jericht:
```suggestion
  let filename: string = await determineNextSequentialFilename(mountPoint);
```

Same goes for other uses of `var` that can be block-scoped

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/filesystem-ops.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+export async function listNumberedFiles(location: string): Promise<string[]> {
+  const dirContents = await fsp.readdir(location);
+  const digitsRegex = /\d+/;
+  const numericFiles = dirContents.filter(name => digitsRegex.test(name)).sort();
+  return numericFiles;
+}
+
+/**
+ * Invoke "du -sh -BM" on the given location, to determine the total size in MB stored
```
jericht, line 57:
```suggestion
 * Invoke "du -sh -BMB" on the given location, to determine the total size in MB stored
```

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts
```diff
@@ -0,0 +1,337 @@
...
+    expect(usage).toBe(2);
+  });
+
+});
+
+describe('Testing padFilesystem macro behavior', () => {
+
+  var tempDirectory: string;
+
+  beforeEach(async () => {
```
jericht, line 191: Looks like we're using these `beforeEach` and `afterEach` functions for all of these `describe` blocks. Should we move these functions outside of the `describe` blocks (and the `tempDirectory` variable) so they only need to be defined once? Could also wrap all the `describe` blocks in one parent `describe` and put these functions in that parent scope

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts
```diff
@@ -0,0 +1,337 @@
...
+    const dirContents = (await fsp.readdir(tempDirectory)).sort();
+    expect(preDirContents).toEqual(['00000']);
+    expect(dirContents).toEqual(preDirContents);
+    for (var file of dirContents) {
+      const stat = await fsp.stat(join(tempDirectory, file));
+      expect(stat.size).toBe(10000000);
+    }
+  });
+
+});
```
jericht:
```suggestion
});

```

Missing newline

jericht (changes requested):
Just some minor comments, otherwise this looks really good!

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
```diff
@@ -0,0 +1,316 @@
...
+   */
+  readonly securityGroup?: ISecurityGroup;
+
+  /**
+   * The desired total size, in GB, of files stored in the access point directory.
+   */
+  readonly desiredPaddingGB: number;
```
jusiskin: nit: we *could* use the [`Size`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Size.html) type that CDK provides here. This has methods that deal in the *bibytes family of units - which conveniently are the same family of units that [EFS uses for computing baseline throughput](https://docs.aws.amazon.com/efs/latest/ug/performance.html#efs-burst-credits):

> The baseline rate is 50 MiB/s per TiB of storage

It provides an expressive API for specifying storage sizes in Gibibytes and higher-order units and also is able to enforce that they be multiples of a particular unit (gibibytes in our case):

```ts
import { Size } from '@aws-cdk/core';

// ...
const size = Size.pebibyte(2);
const sizeInGiB = size.toGibibytes({ rounding: SizeRoundingBehavior.FAIL });
```

What do you think?

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/storage_tier.py
```diff
@@ -91,6 +92,34 @@ def __init__(self, scope: Construct, stack_id: str, *, props: StorageTierProps,
...
+        pad_access_point = AccessPoint(
+            self, 
+            'PaddingAccessPoint',
+            file_system=file_system,
+            path='/PaddingFiles',
+            # TODO - We set the padding files to be owned by root (uid/gid = 0) by default. You may wish to change this.
+            create_acl=Acl(
+                owner_gid='0',
+                owner_uid='0',
+                permissions='744',
```
jusiskin: Was the intention to make this directory readable by all users mounting the EFS? If so, I believe we want to set this to `755`, which is required to enter the directory, list its contents, and access its files. Otherwise, we should simply set this to `750` or `700`.

Same feedback applies to the TS example and defaults in the `Repository` construct.

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
```diff
@@ -0,0 +1,316 @@
...
+ *   or enable AWS CloudTrail on your account to assist in post-incident analysis of compromised production
+ *   environments.
+ * - By default, the network interfaces created by this construct's AWS Lambda Functions have Security Groups
+ *   that restrict egress access from the Lambda Function into your VPC such that the Lambda Functions can
+ *   access only the given EFS Access Point.
+ */
+export class PadEfsStorage extends Construct {
+  // Testing shows about 70-80 MB/s throughput from the lambda to the EFS,
+  // so 15 minutes gives us a maximum of about 63GB - 72GB of writing in one invocation.
+  protected static readonly MAX_GB_PER_CALL = 65;
```
jusiskin: as discussed offline, this is no longer used.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -807,6 +822,7 @@ test('repository creates filesystem if none provided', () => {
   // THEN
   expectCDK(stack).to(haveResource('AWS::EFS::FileSystem'));
   expectCDK(stack).to(haveResource('AWS::EFS::MountTarget'));
+  expect(repo.node.tryFindChild('PadEfsStorage')).toBeDefined();
```
jusiskin, line 825: We should also assert the EFS access point for padding files is created here.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -506,6 +514,22 @@ export class Repository extends Construct implements IRepository {
         securityGroup: props.securityGroupsOptions?.fileSystem,
       });
 
+      const paddingAccess = fs.addAccessPoint('PaddingAccessPoint', {
+        createAcl: {
+          ownerGid: '0',
+          ownerUid: '0',
+          permissions: '744',
```
jusiskin, line 522: same feedback about using one of `755` / `750` / `700`

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/filesystem-ops.ts
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * List all of the names in the given directory that are numeric.
+ * @param location Path of the directory to list. Assumed to be a directory.
+ * @returns Array of the names of numeric contents in the directory, sorted into increasing order.
+ */
+export async function listNumberedFiles(location: string): Promise<string[]> {
+  const dirContents = await fsp.readdir(location);
+  const digitsRegex = /\d+/;
+  const numericFiles = dirContents.filter(name => digitsRegex.test(name)).sort();
```
jusiskin: nit: the regex is not anchored, so it will match any string containing a single digit. I believe the intend was to match entirely numeric strings, which would be:

```ts
const digitsRegex = /^\d+$/;
```

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/handlers.ts
```diff
@@ -0,0 +1,192 @@
...
+          console.error(`Unable to delete: ${filename} -- Error: ${err.message}`);
+        }
+      }
+    } catch (err) {
+      console.warn(`Warning: Unable to stat file '${filename}'`);
+    }
+    index -= 1;
+  }
+  console.log(`Deleted ${numFilesDeleted} numbered files`);
+}
```
jusiskin: It looks like errors are never surfaced to the caller here. I understand we want to be robust here and attempt to skip a file if we encounter errors and try deleting other files.

Can we change the code to throw an error at the end if we were unable to delete the desired number of files by the end of the loop? That way the state machine would actually show an error for this state and transition to the fail state.

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/handlers.ts
```diff
@@ -0,0 +1,192 @@
...
+  const mountPoint = getMountPoint(event);
+
+  try {
+    // Make sure that the given directory has been mounted before continuing.
+    await ensureIsDirectory(mountPoint);
+  } catch (err) {
+    throw new Error(`Mount point '${mountPoint}' is not a directory. Please ensure that the EFS is mounted to this directory.`);
+  }
+
+  const duGB = Math.floor(await diskUsage(mountPoint) / FILESIZE);
```
jusiskin: Perhaps it is a variable naming issue (see my comment above), but I'm trying to figure out why `duGB` would be a function of the size of the files we create

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
ddneilson: Heh. Yup. Right you are!

packages/aws-rfdk/lib/core/test/pad-efs-storage.test.ts
ddneilson: We could. I was less concerned about the precise timeout of this one. The other is writing a tonne of big files, so I wanted to make sure that there was an assertion that the timeout was maximal (900s).

packages/aws-rfdk/lib/core/test/pad-efs-storage.test.ts
ddneilson: That's sensible.

packages/aws-rfdk/lib/core/test/pad-efs-storage.test.ts
ddneilson: Right you are!

packages/aws-rfdk/lib/deadline/lib/repository.ts
ddneilson: I like it. Good idea

packages/aws-rfdk/lib/deadline/test/repository.test.ts
ddneilson: Yes. That would be consistent. Whoops...

For the record. I had to change these 'cause you cannot create a Lambda in a public subnet without a particular option on the Lambda definition. Seemed silly to provide that option just for a test, and customers shouldn't be deploying their EFS with a public IP.

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/filesystem-ops.ts
ddneilson: Good attention to detail. I originally had the code as `-BM` but EFS seems to use MB units for billing, so I went for staying consistent.

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/handlers.ts
ddneilson: Good point!

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts
ddneilson: I was going for grouping tests by top-level theme: "Filesystem mods", "getDiskUsage" handler, and "padFilesystem" handler. It *could* just all be under the same describe, or not use a describe at all, but I like the grouping. Makes my OCD a little happier. :-)

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts
ddneilson: Thanks!

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts
```diff
@@ -0,0 +1,336 @@
...
+      logStreamName: '',
+    });
+
+    // THEN
+    const dirContents = (await fsp.readdir(tempDirectory)).sort();
+    expect(dirContents).toEqual(['00000']);
+    for (var file of dirContents) {
+      const stat = await fsp.stat(join(tempDirectory, file));
+      expect(stat.size).toBe(10000000);
+    }
```
jusiskin: I was a little bit confused about this expected result here (a single 10MB file) given we're passing in a `desiredPadding` of 1GB. The call in the `GIVEN` section appears to change the file size (10MB), but we only expect one of those 10MB files. It occurred to me that may have been by design to reduce storage requirements and reduce IO for running the tests.

Perhaps a comment above the file size expectation would help. What do you think?

jusiskin (changes requested):
Overall this is great. Just a few suggestions here - great work!

packages/aws-rfdk/lib/core/test/pad-efs-storage.test.ts
jericht: Sounds good to me

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts
jericht: Sure, works for me

jericht (dismissed):

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/storage_tier.py
ddneilson: Changed to 700

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
ddneilson: Yup. Nor `LAMBDA_DURATION` -- both removed.

packages/aws-rfdk/lib/deadline/lib/repository.ts
ddneilson: Done

packages/aws-rfdk/lib/deadline/test/repository.test.ts
ddneilson: Good idea.

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/filesystem-ops.ts
ddneilson: Good catch!

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/handlers.ts
ddneilson: Renamed: `scaledUsage`

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts
ddneilson: Added a comment to the top of the test where we set `setDefaultFilesize(10)`

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
ddneilson: Great idea. Looking into it also surfaced that I had the units wrong. EFS is all on GiB rather than GB. So, I fixed that throughout as well.

jericht (dismissed):

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
```diff
@@ -0,0 +1,321 @@
...
+ * provided to give you a way to increase the baseline throughput of an Amazon EFS filesystem
+ * that has been deployed in bursting throughput mode (see: https://docs.aws.amazon.com/efs/latest/ug/performance.html#throughput-modes).
+ * This is most useful for your Amazon EFS filesystems that contain a very small amount of data and
+ * have a baseline throughput that exceeds the throughput provided by the size of the filesystem.
+ *
+ * When deployed in bursting throughput mode, an Amazon EFS filesystem provides you with a baseline
+ * throughput that is proportional to the amount of data stored in that filesystem. However, usage
+ * of that filesystem is allowed to burst above that throughput; doing so consumes burst credits that
+ * are associated with the filesystem. When all burst credits have been expended, then your filesystem
+ * is no longer allowed to burst throughput and you will be limited in throughput to the greater of 1MB/s
```
jusiskin: nit: one leftover here? `1MB/s` &rarr; `1MiB/s`

ddneilson: Fixed.

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
jusiskin: Looks great!

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
jusiskin: resolved

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts
jusiskin: Much better - thanks!

packages/aws-rfdk/lib/core/lib/pad-efs-storage.ts
```diff
@@ -103,8 +103,9 @@ export interface PadEfsStorageProps {
  * throughput that is proportional to the amount of data stored in that filesystem. However, usage
  * of that filesystem is allowed to burst above that throughput; doing so consumes burst credits that
  * are associated with the filesystem. When all burst credits have been expended, then your filesystem
- * is no longer allowed to burst throughput and you will be limited in throughput to the greater of 1MB/s
- * or the throughput dictated by the amount of data stored in your filesystem.
+ * is no longer allowed to burst throughput and you will be limited in throughput to the greater of 1MiB/s
+ * or the throughput dictated by the amount of data stored in your filesystem; the filesystem will be able
+ * to burst again if it is able to acrue burst credits by staying below its baseline throughput for a time.
```
jericht:
```suggestion
 * to burst again if it is able to accrue burst credits by staying below its baseline throughput for a time.
```

ddneilson: Fixed.

jericht (approved):

packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/handlers.ts
jusiskin: Resolved

jusiskin (approved):
Great job, Daniel! Approved

PR #367, opened 2021-03-29
feat(deadline): add option to the RenderQueue to use cachefilesd

### LINKED ISSUE #366: Enable cachefilesd on RenderQueue
Linux provides FS-Cache ( https://www.kernel.org/doc/Documentation/filesystems/caching/fscache.txt ) as a mechanism for transparently caching network file accesses. Filesystem driver that support it can interface with this subsystem via `cachefilesd`. Using this can reduce network traffic to networked filesystems when the majority of operations are read-ops -- as is the case with the RenderQueue.

### Use Case

Nominally, when using Amazon EFS configured for bursting throughput as the backing filesystem for the Deadline Remote Connection Server. Reducing the read traffic to the filesystem reduces the total reads to the filesystem, which reduces the throughput that must be indirectly provisioned, by adding data to the filesystem, to reduce burst credit consumption rate. 

### Proposed Solution

Option on the RenderQueue (off by default) that will install & start the `cachefilesd` service on the ECS container host.

### Other

N/A

---

ddneilson:
Implements: https://github.com/aws/aws-rfdk/issues/366

### Testing

Started up a basic render farm using the RFDK examples, but modified the example to set the `fsc` mount option on the Amazon EFS and enable `cachefilesd` on the RenderQueue.

Used systems manager connection to remote in to the RenderQueue's ECS container host, and then:
1. Verified that cachefilesd is installed and running.
2. Verified that the `/mnt/repo` filesystem was mounted using the EFS helper which in-turn uses the NFSv4 mount driver, and that the `fsc` option was properly set on the NFSv4 mount point.
```bash
[root@ip-10-0-126-223 ~]# mount | grep nfs
sunrpc on /var/lib/nfs/rpc_pipefs type rpc_pipefs (rw,relatime)
127.0.0.1:/ on /mnt/repo type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,hard,noresvport,proto=tcp,port=20312,timeo=600,retrans=2,sec=sys,clientaddr=127.0.0.1,fsc,local_lock=none,addr=127.0.0.1,_netdev)
```
3. Ran a series of full copies of the `/mnt/repo` filesystem (approx 1.3 GB of data) to the local device both with and without cachefilesd running. The image below shows the throughput metrics on the Amazon EFS filesystem through these tests; as expected, the read throughput with cachefilesd is much lower than a full copy once the local cache has been populated.

RepoCopy

### Performance Testing

Test setup:
1. RFDK Basic all-in application with an added bastion host set up as a Workstation (installed Deadline & DCV). So... single RCS (c5.large), EFS burst-mode filesystem with ~1GB of data (encrypted), DocDB (db.r5.large), Linux-based Worker ASG (t3.medium), and TLS enabled throughout

summary

So, cachefilesd shaves about 30% of metered throughput when starting 200 workers (9.73MB/s -> 6.9MB/s) and 26% of metered throughput when rendering 1k 'sleep 10' jobs (1.9MB/s -> 1.6MB/s). With the standard caveats, of course, about this being a single data point so there will be sizable error bars on those.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
+++ b/packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
@@ -11,7 +11,7 @@ then
   # The yum install can fail if there is no route to the yum repository.
   # Since this is an AL2 instance, that is equivalent to requiring a route to
   # the regional S3 endpoint on port 80.
-  sudo yum install -y cachefilesd || exit 0
+  sudo yum install -y cachefilesd || ( echo 'ERROR -- Failed to install cachefilesd' && exit 0 )
 
   cat << EOF | sudo tee -a /etc/cachefilesd.conf
 # Allow 16k cull table entries
--- a/packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
+++ b/packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
@@ -11,7 +11,7 @@ then
   # The yum install can fail if there is no route to the yum repository.
   # Since this is an AL2 instance, that is equivalent to requiring a route to
   # the regional S3 endpoint on port 80.
-  sudo yum install -y cachefilesd || ( echo 'ERROR -- Failed to install cachefilesd' && exit 0 )
+  sudo yum install -y cachefilesd || echo 'ERROR -- Failed to install cachefilesd' && exit 0
 
   cat << EOF | sudo tee -a /etc/cachefilesd.conf
 # Allow 16k cull table entries
--- a/packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
+++ b/packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+# Script to install and configure cachefilesd onto a Linux host.
+# This script is presently narrowly written for use on the ECS container host
+# instance(s) that are running the Deadline RCS containers in the RenderQueue construct.
+
+set -xeu
+
+if ! test -f /sbin/cachefilesd
+then
+  # The yum install can fail if there is no route to the yum repository.
+  # Since this is an AL2 instance, that is equivalent to requiring a route to
+  # the regional S3 endpoint on port 80.
+  sudo yum install -y cachefilesd || exit 0
+
+  cat << EOF | sudo tee -a /etc/cachefilesd.conf
+# Allow 16k cull table entries
+culltable 14
+EOF
+
+fi
+
+# Make sure cachefilesd is running
+sudo systemctl enable cachefilesd
+sudo systemctl start cachefilesd
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/service_tier.py
@@ -171,7 +171,11 @@ class ServiceTier(Stack):
             # TODO - Evaluate deletion protection for your own needs. This is set to false to
             # cleanly remove everything when this stack is destroyed. If you would like to ensure
             # that this resource is not accidentally deleted, you should set this to true.
-            deletion_protection=False
+            deletion_protection=False,
+            # Enable a local transparent filesystem cache of the Repository filesystem to reduce
+            # data traffic from the Repository's filesystem.
+            # For an EFS and NFS filesystem, this requires the 'fsc' mount option.
+            enable_local_file_caching=True,
         )
         self.render_queue.connections.allow_default_port_from(self.bastion)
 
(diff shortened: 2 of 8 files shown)


RandomInsano:
Looks great! My only concern is with the namespace for this option and underlying function names being generic as it has the following two effects:
1. Those aware of cachefilesd have to check the docs to know what the underlying tech is
2. If we find an *even better* caching option or choose to change the Repo's backing protocol (say to SMB) we'll have problems.

Maybe we could swap "enable_local_file_caching" for "enable_fsc_caching", etc?

packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
```diff
@@ -0,0 +1,25 @@
...
+# instance(s) that are running the Deadline RCS containers in the RenderQueue construct.
+
+set -xeu
+
+if ! test -f /sbin/cachefilesd
+then
+  # The yum install can fail if there is no route to the yum repository.
+  # Since this is an AL2 instance, that is equivalent to requiring a route to
+  # the regional S3 endpoint on port 80.
+  sudo yum install -y cachefilesd || exit 0
```
jericht: Should we echo a warning message if the yum install fails?

jericht (dismissed):
Just one really minor comment, otherwise this LGTM

ddneilson:
> 
> 
> Looks great! My only concern is with the namespace for this option and underlying function names being generic as it has the following two effects:
> 
>     1. Those aware of cachefilesd have to check the docs to know what the underlying tech is
> 
>     2. If we find an _even better_ caching option or choose to change the Repo's backing protocol (say to SMB) we'll have problems.
> 
> 
> Maybe we could swap "enable_local_file_caching" for "enable_fsc_caching", etc?

@RandomInsano Thanks for the feedback, Edwin. Variable/property naming is hard. ;-) On one hand... generic for those that don't know the details, or don't want to know/care... On the other hand... it is nice to know what is going on under the hood. I'm trying to strike that balance by going generic with the property name and providing details in the docstring (which shows up under quick-info-type IDE functionality). My thinking is also that if there's a different caching tech for a different filesystem, then we'd want to support that under the same option but detect the filesystem type if we can; it's all the same high-level functionality, regardless of how it's accomplished.

You're thinking that I'm missing the mark with it?

packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
ddneilson: Yeah, that would be sensible.

packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
```diff
@@ -0,0 +1,25 @@
...
+# instance(s) that are running the Deadline RCS containers in the RenderQueue construct.
+
+set -xeu
+
+if ! test -f /sbin/cachefilesd
+then
+  # The yum install can fail if there is no route to the yum repository.
+  # Since this is an AL2 instance, that is equivalent to requiring a route to
+  # the regional S3 endpoint on port 80.
+  sudo yum install -y cachefilesd || ( echo 'ERROR -- Failed to install cachefilesd' && exit 0 )
```
jericht:
```suggestion
  sudo yum install -y cachefilesd || echo 'ERROR -- Failed to install cachefilesd' && exit 0
```
Looks like we are now creating a subshell and running `exit` from there, so the script will continue to run even if the yum install fails. This is different from what we had before. Should we get rid of the subshell?

packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
```diff
@@ -0,0 +1,25 @@
...
+
+  cat << EOF | sudo tee -a /etc/cachefilesd.conf
+# Allow 16k cull table entries
+culltable 14
+EOF
```
jusiskin, line 19: Maybe I'm missing something, but should we apply this configuration regardless of whether `cachefilesd` was present (move outside the `if` block)?

jusiskin (changes requested):
Just one minor question, but otherwise looks great.

packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
ddneilson: Drat. Right you are.

packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
ddneilson: I was taking the approach that if it's already installed, then it has been configured by someone and I shouldn't mess with it. It's not normally installed on an ECS container host...

jericht (approved):

packages/aws-rfdk/lib/deadline/scripts/bash/enableCacheFilesd.sh
jusiskin: That makes sense, resolving!

jusiskin (approved):

PR #373, opened 2021-03-31
feat(examples): Demonstrate how to send an email alarm when EFS burst credits below a threshold

ddneilson:
When EFS is deployed in bursting throughput mode it will consume credits whenever throughput is
greater than a baseline throughput dictated by the amount of data stored in EFS. If the average
throughput over time is greater than the baseline, then burst credits will continuously decrease
until they eventually run out.

This adds to the Basic example some CDK code that demonstrates the good practice of setting up
an email alarm when the available burst credits drops below some thresholds; giving the operator
time to increase the amount of data stored on the EFS to increase baseline throughput.

Bonus: Fixes some math errors in the PadEfsStorage lambda function.

### Testing

1. Deployed the Basic Typescript app up to the ServiceTier. Checked Lambda logs to ensure that the files are writing via the padder.
2. Deployed a modified version of (1) with different thresholds and then used the RCS container host to write files to the EFS until I breached threshold. Email alarm was received.
3. cdk synth of the Python example modifications.

----

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #381, opened 2021-04-05
fix(examples): make alarm email optional and add README instructions

jusiskin:
## Problem

#373 added code in the AWS-All-in-Basic example to allow configuring an email address to which operational alarm notifications are sent. The intention was to have this optional and when not specified, no alarms would be configured. In the TypeScript version of the example, this was not made optional and anyone using it would need to  modify the code. This new configuration setting was not documented in the `README.md` files.

## Solution

*   Made the property optional in the TypeScript example
*   Added a step to both the Python and TypeScript `README.md` to optionally populate this configuration setting

----


ddneilson (approved):
Thanks for filling the gap that I left behind, Josh.

jericht (approved):

PR #389, opened 2021-04-13
test(integ): add test for RCS process owner

jusiskin:
## Problem

The RFDK integration tests does not have any coverage for the owning user of the process that the Deadline RCS runs as within the ECS task container. This means that we have no assurance that RFDK follows the principle of least-privilege with respect to the process permissions within these containers.

## Solution

Add a test that checks the OS user of the RCS process running within the ECS task containers deployed by RFDK's `RenderQueue` construct. The Deadline RCS already reports it's process' user to Deadline's database. This tests relies on:

```
deadlinecommand GetProxyServerInfos
```

which returns this information as reported by the RCS.

## Testing

Ran the integration tests and confirmed that the new test was run and passed:

```
Complete!
Pretest setup runtime: 0m 4s
Infrastructure stack deploy runtime: 4m 1s
Infrastructure stack cleanup runtime: 2m 0s
Results for test component deadline_02_renderQueue:
  -Tests ran: 10
  -Tests passed: 10
  -Tests failed: 0
  -Deploy runtime:     24m 12s
  -Test suite runtime: 0m 57s
  -Cleanup runtime:    16m 18s
Cleaning up folders...
```

----

### DIFF
--- a/integ/components/deadline/deadline_02_renderQueue/scripts/bastion/testing/RQ-query-rcs-user.sh
+++ b/integ/components/deadline/deadline_02_renderQueue/scripts/bastion/testing/RQ-query-rcs-user.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Script to return the username for the process running the RCS as reported by Deadline
+#
+# Input:
+#   None
+# Output:
+#   Non-zero return code on failure.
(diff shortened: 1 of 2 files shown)


jericht (approved):

horsmand (approved):
I discussed with Josh about using something like `ps` to determine what the OS is reporting the process owner as, but he explained how since we're only running these SSM commands against the bastion instance, there is a lot of overhead to running the command against the RCS.

This method of using Deadline Command to query for the list of proxy servers and then read the user from that info relies on Deadline reporting the correct user, but I don't see any reason for that to report incorrectly in the farm setups we use in our integration tests, so I believe this is fine.

PR #392, opened 2021-04-16
feat(core): add ability to resolve mount targets using EFS API

### LINKED ISSUE #391: Add ability to mount EFS in VPCs without Amazon-supplied DNS resolver
The RFDK code to mount EFS file-systems on an instance assumes that the instance belongs to a VPC with Amazon-supplied DNS (see [VPC DNS documentation](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html)). When a VPC is configured to disable the Amazon-supplied DNS resolver, then the DNS names used by both the EFS mount helper and the NFS mounting code cannot be resolved and the mount will fail with an error showing up in the `cloud-init` log file, such as:

```
[   32.413572] cloud-init[3739]: Failed to resolve "fs-a1b2c3d4.efs.us-west-2.amazonaws.com" - check that your file system ID is correct.
```

### Use Case

Users that want to supply their own DNS servers for the VPC(s) used with RFDK should be able to mount EFS but cannot given the current code.

### Proposed Solution

The DNS names in the error message above correspond to [EFS mount  targets](https://docs.aws.amazon.com/efs/latest/ug/manage-fs-access.html). These provide the endpoints that serve the networked file-system traffic. EFS automatically adds DNS entries into your VPC's Amazon-supplied Route 53 resolver, but when this feature is disabled, those names cannot be resolved using DNS.

Fortunately, the IP address of each mount target can be obtained by calling the EFS [`DescribeMountTargets` API](https://docs.aws.amazon.com/efs/latest/ug/API_DescribeMountTargets.html).

We can add optional logic in our EFS mounting script that will perform this lookup and bake the IP address into the host machine's `/etc/hosts` file to bypass DNS lookups.

---

jusiskin:
Resolves #391

Taken from that issue:

> The RFDK code to mount EFS file-systems on an instance assumes that the instance belongs to a VPC with Amazon-supplied DNS (see [VPC DNS documentation](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html)). When a VPC is configured to disable the Amazon-supplied DNS resolver, then the DNS names used by both the EFS mount helper and the NFS mounting code cannot be resolved and the mount will fail with an error showing up in the `cloud-init` log file, such as:

> ```
> [   32.413572] cloud-init[3739]: Failed to resolve "fs-a1b2c3d4.efs.us-west-2.amazonaws.com" - check that your file system ID is correct.
> ```

## Solution

The DNS names in the error message above correspond to [EFS mount  targets](https://docs.aws.amazon.com/efs/latest/ug/manage-fs-access.html). These provide the endpoints that serve the networked file-system traffic. EFS automatically adds DNS entries into your VPC's Amazon-supplied Route 53 resolver, but when this feature is disabled, those names cannot be resolved using DNS.

Fortunately, the IP address of each mount target can be obtained by calling the EFS [`DescribeMountTargets` API](https://docs.aws.amazon.com/efs/latest/ug/API_DescribeMountTargets.html).

Added an optional boolean flag when EFS mounting script that will perform this lookup and bake the IP address into the host machine's `/etc/hosts` file to bypass DNS lookups. Using this looks like:

**TypeScript:**

```ts
new MountableEfs(scope, {
  // ...
  resolveMountTargetDnsWithApi: true,
});
```

**Python:**

```py
MountableEfs(scope,
  # ...
  resolve_mount_target_dns_with_api=True,
)
```

## Testing

Modified the `All-In-AWS-Infrastructure-Basic` example, enabling the above flag. Connected to the mounting instances using SSM Session Manager and confirmed:

*   The `/etc/hosts` file contains an entry for the mount target:

        10.0.71.79 fs-a1b2c1d2.efs.us-west-2.amazonaws.com # Added by RFDK

*   The EFS file-system driver establishes a connection to the specified IP address:

    ```sh
    $ sudo netstat -alnp | grep 10.0.71.79
    tcp        0      0 10.0.0.13:55102         10.0.71.79:2049         ESTABLISHED 2306/stunnel
    ```

*   Deployment succeeds
*   Am able to connect to the farm, submit a job, view successful render logs (requires proper EFS mounting on the `RenderQueue`)

----

### DIFF
--- a/packages/aws-rfdk/lib/core/lib/mountable-efs.ts
+++ b/packages/aws-rfdk/lib/core/lib/mountable-efs.ts
@@ -188,7 +188,7 @@ export class MountableEfs implements IMountableLinuxFilesystem {
       'TMPDIR=$(mktemp -d)',
       'pushd "$TMPDIR"',
       `unzip ${mountScript}`,
-      `bash ./mountEfs.sh ${this.props.filesystem.fileSystemId} ${mountDir} ${mountOptionsStr} ${resolveMountTargetDnsWithApi}`,
+      `bash ./mountEfs.sh ${this.props.filesystem.fileSystemId} ${mountDir} ${resolveMountTargetDnsWithApi} ${mountOptionsStr}`,
       'popd',
       `rm -f ${mountScript}`,
     );
--- a/packages/aws-rfdk/lib/core/lib/mountable-efs.ts
+++ b/packages/aws-rfdk/lib/core/lib/mountable-efs.ts
@@ -66,6 +66,18 @@ export interface MountableEfsProps {
    * @default No extra options.
    */
   readonly extraMountOptions?: string[];
+
+  /**
+   * If enabled, RFDK will add user-data to the instances mounting this EFS file-system that obtains the mount target
+   * IP address using AWS APIs and writes them to the system's `/etc/hosts` file to not require DNS lookups.
+   *
+   * If mounting EFS from instances in a VPC configured to not use the Amazon-provided DNS Route 53 Resolver server,
+   * then the EFS mount targets will not be resolvable using DNS (see
+   * https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html) and enabling this will work around that issue.
+   *
+   * @default false
+   */
+  readonly resolveMountTargetDnsWithApi?: boolean;
 }
 
 /**
@@ -157,11 +169,26 @@ export class MountableEfs implements IMountableLinuxFilesystem {
     }
     const mountOptionsStr: string = mountOptions.join(',');
 
+    const resolveMountTargetDnsWithApi = this.props.resolveMountTargetDnsWithApi ?? false;
+    if (resolveMountTargetDnsWithApi) {
+      const describeMountTargetResources = [
+        (this.props.filesystem.node.defaultChild as efs.CfnFileSystem).attrArn,
+      ];
+      if (this.props.accessPoint) {
+        describeMountTargetResources.push(this.props.accessPoint.accessPointArn);
+      }
+
+      target.grantPrincipal.addToPrincipalPolicy(new PolicyStatement({
+        resources: describeMountTargetResources,
+        actions: ['elasticfilesystem:DescribeMountTargets'],
+      }));
+    }
+
     target.userData.addCommands(
       'TMPDIR=$(mktemp -d)',
       'pushd "$TMPDIR"',
       `unzip ${mountScript}`,
-      `bash ./mountEfs.sh ${this.props.filesystem.fileSystemId} ${mountDir} ${mountOptionsStr}`,
+      `bash ./mountEfs.sh ${this.props.filesystem.fileSystemId} ${mountDir} ${mountOptionsStr} ${resolveMountTargetDnsWithApi}`,
       'popd',
       `rm -f ${mountScript}`,
     );
--- a/packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
+++ b/packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
@@ -13,12 +13,15 @@
 #  $1 -- EFS Identifier (ex: efs-00000000000)
 #  $2 -- Mount path; directory that we mount the EFS to.
 #  $3 -- (optional) NFSv4 mount options for the EFS. 
+#  $4 -- (optional) whether to obtain the EFS mount target's IP address using the EFS API and persist this to
+#        the /etc/hosts file on the system. This allows the script to work when the mounting instance cannot resolve the
+#        mount target using DNS. This defaults to being disabled, specify "true" to enable this feature.
 
 set -xeu
 
 if test $# -lt 2
 then
-  echo "Usage: $0 <file system ID> <mount path> [<mount options>]"
+  echo "Usage: $0 FILE_SYSTEM_ID MOUNT_PATH [MOUNT_OPTIONS] [RESOLVE_MOUNT_POINT_USING_API]"
   exit 1
 fi
 
@@ -31,10 +34,12 @@ authenticate_identity_document
 
 METADATA_TOKEN=$(get_metadata_token)
 AWS_REGION=$(get_region "${METADATA_TOKEN}")
+AVAILABILITY_ZONE_NAME=$(get_availability_zone "${METADATA_TOKEN}")
 
 FILESYSTEM_ID=$1
 MOUNT_PATH=$2
 MOUNT_OPTIONS="${3:-}"
+RESOLVE_MOUNTPOINT_IP_VIA_API="${4:-false}"
 
 sudo mkdir -p "${MOUNT_PATH}"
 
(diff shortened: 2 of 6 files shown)


packages/aws-rfdk/lib/core/lib/mountable-efs.ts
```diff
@@ -66,6 +66,18 @@ export interface MountableEfsProps {
    * @default No extra options.
    */
   readonly extraMountOptions?: string[];
+
+  /**
```
ddneilson, line 70: There's a downside here that we should mention... If you have mount targets in multiple AZs, then the Route53 resolver will always give you the mount target IP that's in the same AZ as the requester (or random if the request isn't coming from an IP that shares an AZ with a resolution).

When resolving via IP address like this you don't get that AZ-matching. Thus, cross-AZ data charges may be incurred.

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
```diff
@@ -58,6 +63,68 @@ function use_nfs_mount() {
...
+
+${MOUNT_POINT_IP} ${DNS_NAME} # Added by RFDK
+EOF
+}
+
+# Optionally resolve DNS using the EFS API
+if [[ $RESOLVE_MOUNTPOINT_IP_VIA_API == "true"  ]]
+then
+  # jq is used to query the JSON API response
+  sudo "${PACKAGE_MANAGER}" install -y jq
```
ddneilson: Should check whether `jq` is present, and only hit the package manager if it is not. Use-case is an isolated machine with no network connectivity to the package repository.

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
```diff
@@ -58,6 +63,68 @@ function use_nfs_mount() {
...
+function resolve_mount_target_ip_via_api() {
+  local EFS_FS_ID=$1
+  local MNT_TARGET_RESOURCE_ID=$2
+  local AVAILABILITY_ZONE_NAME=$3
+  local AWS_REGION=$4
+  local MOUNT_POINT_IP=""
+
+  if [[ $MNT_TARGET_RESOURCE_ID == fs-* ]]; then
+    # Mounting without an access point
+    MOUNT_POINT_IP=$(aws efs describe-mount-targets \
```
ddneilson: Missing failure handling -- we're not guaranteed that there is a MountTarget in the same AZ as the instance. Route53 would resolve the DNS with a randomly chosen IP in that case.

Same in the `fsap-*` case

packages/aws-rfdk/lib/core/lib/mountable-efs.ts
```diff
@@ -157,11 +169,26 @@ export class MountableEfs implements IMountableLinuxFilesystem {
...
+        actions: ['elasticfilesystem:DescribeMountTargets'],
+      }));
+    }
+
     target.userData.addCommands(
       'TMPDIR=$(mktemp -d)',
       'pushd "$TMPDIR"',
       `unzip ${mountScript}`,
-      `bash ./mountEfs.sh ${this.props.filesystem.fileSystemId} ${mountDir} ${mountOptionsStr}`,
+      `bash ./mountEfs.sh ${this.props.filesystem.fileSystemId} ${mountDir} ${mountOptionsStr} ${resolveMountTargetDnsWithApi}`,
```
ddneilson: I suggest reversing the ordering of mount options & the new arg. Reason being that mount options could be empty, but the new arg will never be.

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
```diff
@@ -13,12 +13,15 @@
...
+#  $4 -- (optional) whether to obtain the EFS mount target's IP address using the EFS API and persist this to
+#        the /etc/hosts file on the system. This allows the script to work when the mounting instance cannot resolve the
+#        mount target using DNS. This defaults to being disabled, specify "true" to enable this feature.
 
 set -xeu
 
 if test $# -lt 2
 then
-  echo "Usage: $0 <file system ID> <mount path> [<mount options>]"
+  echo "Usage: $0 FILE_SYSTEM_ID MOUNT_PATH [MOUNT_OPTIONS] [RESOLVE_MOUNT_POINT_USING_API]"
```
ddneilson: Makes sense to me to make the new arg required. Two optional positional args can lead to confusion.

ddneilson (changes requested):

packages/aws-rfdk/lib/core/lib/mountable-efs.ts
jusiskin: This code actually does performs AZ matching. See:

https://github.com/jusiskin/aws-rfdk/blob/daa2fd714859be3cf31c45a394ec883f8dfa7f94/packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh#L78

and

https://github.com/jusiskin/aws-rfdk/blob/daa2fd714859be3cf31c45a394ec883f8dfa7f94/packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh#L85

ddneilson: Yeah. I saw that, but didn't come back to remove the comment. My bad.

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
jusiskin: Given the concern you mentioned about cross-AZ data charges, do you think we should replicate this behavior? Or is it better to fail? I'm learning towards failing with a user-friendly warning over using a random mount target. That way anyone using this could avoid the cross-AZ costs and simply add the missing mount target.

The calling code has a catch-all error handler, but you're right - currently this will not fail gracefully if there's no matching mount target. I'll fix that - thanks!

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
jusiskin: Good call! I'll add that check first

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
ddneilson: My opinion... We want to make sure that the customer knows that there could be cross-AZ data charges, but not prevent them from deploying in a cross-AZ manner. Throwing an error for cross-AZ would make a cross-AZ mount impossible, and there may be valid use-cases for that.

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
```diff
@@ -49,31 +55,104 @@ else
...
+then
+  # jq is used to query the JSON API response
+  if ! where jq > /dev/null 2>&1
+  then
+    "${PACKAGE_MANAGER}" install -y jq
+  fi
+
+  # Get access point ID if available, otherwise file system ID
+  MNT_TARGET_RESOURCE_ID=$FILESYSTEM_ID
+  ACCESS_POINT_MOUNT_OPT=$(echo "${MOUNT_OPTIONS}" | sed -e 's#,#\n#g' | grep 'accesspoint=') || true
```
horsmand, line 149: We're creating a dependency on the format of the mount options as we define them in the MountableEfs construct here and I could see that format getting changed without testing this specific use case. Maybe we should just add a comment where we define those mount options mentioning this?

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
jusiskin: I've added logic to fall-back to using any available mount target with a warning being output that should get logged to CloudWatch. Let me know what you think.

packages/aws-rfdk/lib/core/lib/mountable-efs.ts
jusiskin: done

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
jusiskin: applied

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
jusiskin: This is parsing a standard options string as passed to the mount system call (see `-o` option in the [docs for the CLI command equivalent](https://man7.org/linux/man-pages/man8/mount.8.html)).

Discussion offline, the confusion was around the `accesspoint=...` format, which is what the [EFS mount helper expects](https://docs.aws.amazon.com/efs/latest/ug/efs-mount-helper.html#mounting-access-points) (see docs). Added a comment here with a link to those docs.

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
```diff
@@ -12,13 +12,17 @@
...
-#  $3 -- (optional) NFSv4 mount options for the EFS. 
+#  $3 -- whether to obtain the EFS mount target's IP address using the EFS API and persist this to
+#        the /etc/hosts file on the system. This allows the script to work when the mounting instance cannot resolve the
+#        mount target using DNS. A value of "true" (case sensitive) will  enable this feature. Any other value will
+#        is treated as being disabled.
+#  $4 -- (optional) NFSv4 mount options for the EFS. 
 
 set -xeu
 
 if test $# -lt 2
```
ddneilson:
```suggestion
if test $# -lt 3
```

ddneilson (changes requested):
One small nit, and then I think we're good.

horsmand (dismissed):

packages/aws-rfdk/lib/core/scripts/bash/mountEfs.sh
jusiskin: good catch - fixed

ddneilson (approved):

horsmand (approved):

PR #393, opened 2021-04-19
chore(deadline): adding retries for http requests and timeouts for custom resources

ryyakobe:
This PR fixes 2 potential issues in RFDK:
1. SEP configuration uses HTTP/HTTPS requests to connect to the RenderQueue. We only send one request, so in case of a failure, the stack will be rolled back. This PR adds possibility to retry the requests in case of `503 Service Temporarily Unavailable error` and other `5xx` errors.
2. Our `SimpleCustomResource implementation` doesn't handle Lambda timeouts properly. When Lambda runs out of time, the execution is stopped without sending a failure signal to the CloudFormation. This causes the deployment to hang for a long time (1 hour) before indicating the error. With the changes from this PR, we will always stop the execution earlier then Lambda times out, so the signal will be always sent to the CloudFormation.
----

### DIFF
--- a/packages/aws-rfdk/lib/lambdas/nodejs/lib/aws-lambda/types.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/lib/aws-lambda/types.ts
@@ -13,4 +13,5 @@
 export interface LambdaContext {
   readonly logGroupName: string;
   readonly logStreamName: string;
+  getRemainingTimeInMillis?(): number;
 }
--- a/packages/aws-rfdk/lib/lambdas/nodejs/lib/aws-lambda/types.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/lib/aws-lambda/types.ts
@@ -13,5 +13,5 @@
 export interface LambdaContext {
   readonly logGroupName: string;
   readonly logStreamName: string;
-  getRemainingTimeInMillis?(): number;
+  getRemainingTimeInMillis(): number;
 }
--- a/packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
@@ -90,11 +90,18 @@ export abstract class SimpleCustomResource {
         ]).finally(() => clearTimeout(timer));
       };
 
-      // getRemainingTimeInMillis() should always be available. It is optional only in our code for backwards compatibility.
-      const remainingTime = context.getRemainingTimeInMillis?.() ?? 1000;
+      const defaultReserveTimeMs = 3000;
+      const remainingTimeMs = context.getRemainingTimeInMillis();
+      let reserveTimeMs = Math.min(0.2 * remainingTimeMs, defaultReserveTimeMs);
+      if (reserveTimeMs < defaultReserveTimeMs) {
+        console.debug(`The remaining Lambda execution time of ${reserveTimeMs} ` +
+        `ms might not be sufficient to send a CloudFormation response. At least ${defaultReserveTimeMs} ms is required. ` +
+        'Please increase the Lambda timeout.');
+      }
+
       cfnData = await timeout(
         this.handleEvent(event, context, resourceProperties, physicalId),
-        remainingTime - 1000,
+        remainingTimeMs - reserveTimeMs,
         new Error('Timeout error'),
       );
     } catch (e) {
--- a/packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
@@ -78,32 +78,25 @@ export abstract class SimpleCustomResource {
     let cfnData: object | undefined;
 
     console.log(`Handling event: ${JSON.stringify(event)}`);
-    const requestType: string = event.RequestType;
     const resourceProperties: object = event.ResourceProperties ?? {};
     const physicalId: string = calculateSha256Hash(resourceProperties);
 
     try {
-      if (requestType === 'Create') {
-        if (!this.validateInput(resourceProperties)) {
-          throw Error(`Input did not pass validation check. Check log group "${context.logGroupName}" ` +
-            `for log stream ${context.logStreamName} for additional information.`);
-        }
-        cfnData = await this.doCreate(physicalId, resourceProperties);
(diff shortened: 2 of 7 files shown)


packages/aws-rfdk/lib/lambdas/nodejs/lib/aws-lambda/types.ts
```diff
@@ -13,4 +13,5 @@
 export interface LambdaContext {
   readonly logGroupName: string;
   readonly logStreamName: string;
+  getRemainingTimeInMillis?(): number;
```
ddneilson: Why optional? afaik, it shouldn't break anything to make this required -- it is actually required... we just never used it before.

packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
```diff
@@ -78,32 +78,25 @@ export abstract class SimpleCustomResource {
...
+          prom,
+          new Promise((_r, rej) => timer = setTimeout(rej, time, exception)),
+        ]).finally(() => clearTimeout(timer));
+      };
+
+      // getRemainingTimeInMillis() should always be available. It is optional only in our code for backwards compatibility.
+      const remainingTime = context.getRemainingTimeInMillis?.() ?? 1000;
+      cfnData = await timeout(
+        this.handleEvent(event, context, resourceProperties, physicalId),
+        remainingTime - 1000,
```
ddneilson: Is 1s enough time to reliably resolve the S3 put-object request that signals success/failure of the custom resource? Would we be better off with a timeout that's a bit more generous just in case there's a service blip in S3? e.g. 5s?

packages/aws-rfdk/lib/lambdas/nodejs/lib/deadline-client/deadline-client.ts
```diff
@@ -150,38 +179,51 @@ export class DeadlineClient {
     return options;
   }
 
+  private async performRequestWithRetry(options: https.RequestOptions, retriesLeft: number, retryDelayMs: number, data?: string): Promise<Response> {
+    return this.performRequest(options, data)
```
ddneilson: Just curious... why this style of syntax instead of something like:
```ts
try {
  return await this.performRequest(options, data);
} catch (...) {
 ... exception handling
}
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/aws-lambda/types.ts
ryyakobe: Making it not optional breaks the unit-tests of [pad-efs-storage](https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/lambdas/nodejs/pad-efs-storage/test/handlers.test.ts#L129). I could add something like `getRemainingTimeInMillis: () => 1000,` to the unit-tests and remove optional. Do you agree with such a fix?

ddneilson: Yeah, that sounds perfectly reasonable to me.

packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
ryyakobe: The default Lambda timeout is actually 3 seconds. I could throw an exception with a message `Please increase Lambda timeout` if the `remainingTime` is less than 5 seconds. This will not give us a strong guarantee, but most likely will be caught during the development. What do you think?

packages/aws-rfdk/lib/lambdas/nodejs/lib/deadline-client/deadline-client.ts
ryyakobe: Just because of the historical reasons. I like your style more, so will use it instead.

packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
ddneilson: We could do some math maybe and do the warning for during development? Just to make sure that it'll still timeout and try to send the S3 response if the Lambda timeout is too low.
e.g. Something like `min(0.2 * remainingTime, 3000)` ?

ryyakobe: Sounds good. I've decided to keep 3 seconds instead of 5, used the math you proposed and also added printing a debug message if the remaining time is less than 3 seconds.

packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
```diff
@@ -78,32 +78,32 @@ export abstract class SimpleCustomResource {
...
+      };
+
+      const defaultReserveTimeMs = 3000;
+      const remainingTimeMs = context.getRemainingTimeInMillis();
+      let reserveTimeMs = Math.min(0.2 * remainingTimeMs, defaultReserveTimeMs);
+      if (reserveTimeMs < defaultReserveTimeMs) {
+        console.debug(`The remaining Lambda execution time of ${reserveTimeMs} ` +
+        `ms might not be sufficient to send a CloudFormation response. At least ${defaultReserveTimeMs} ms is required. ` +
+        'Please increase the Lambda timeout.');
       }
```
horsmand, line 108: It's not obvious where the `3000` is coming from or how we chose the values in `0.2 * remainingTimeMs`. Could you add a comment to this section about what we're doing?

horsmand (approved):

packages/aws-rfdk/lib/lambdas/nodejs/lib/custom-resource/simple-resource.ts
ryyakobe: I've added explanation for the purpose of reserved time in a new commit. The values don't have much background behind them. We simply want to finish earlier than Lambda's timeout, so I first used just 1 second. But we also want to be sure, that we will have enough time to notify CloudFormation about the failure, thus we increased that time to 3 seconds.

The smallest timeout in our code is `30` seconds. So we would keep `27` seconds for the actual function execution. That's why I didn't want to increase the reserved time from `3 seconds` to an even bigger number.

ddneilson (approved):

yashda (approved):
Looks good to me. Nice generic solution to resolve the timeout issues in CFN. Thanks Roman!

PR #395, opened 2021-04-19
feat(deadline): add ability to import repository settings

### LINKED ISSUE #397: Import Deadline Repository settings
Deadline has the ability to import/export repository settings (see [Import/Export feature](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/repository-settings-importer-exporter.html)). It would be great if RFDK could:

1. Import repository settings when installing the Repository
2. Import repository settings post-install

### Use Case

To configure the Deadline Repository through RFDK. For example, setting up path mapping rules can be a required step before being able to successfully render with the render farm.

### Proposed Solution

1. Implemented by https://github.com/aws/aws-rfdk/pull/395
2. Create a custom resource that uses the Deadline RCS API to configure the repository settings post-install

### Other

---

jericht:
Partial fix for https://github.com/aws/aws-rfdk/issues/397

### Notes
Adds ability to import Deadline Repository settings with a repository settings file (see https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/repository-settings-importer-exporter.html)

### Testing
Creating a JSON file that conformed to the repository settings importer format with a region and a path mapping rule. Modified the example app to use the file in the `Repository` construct and deployed it. Verified that:
- The Repository installer group succeeds in importing the settings with the Repository installer.
- Ran the `ExportRepositorySettings` Deadline command from the `RenderQueue` container and verified the settings I imported were in the output
Also deployed without the settings and verified the installer still succeeds.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/repository.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/repository.ts
@@ -43,6 +43,9 @@ import {
   PolicyStatement,
 } from '@aws-cdk/aws-iam';
 import {
+  Asset,
+} from '@aws-cdk/aws-s3-assets';
+import {
   Annotations,
   Construct,
   Duration,
@@ -376,6 +379,12 @@ export interface RepositoryProps {
    * Options to add additional security groups to the Repository.
    */
   readonly securityGroupsOptions?: RepositorySecurityGroupsOptions;
+
+  /**
+   * The Deadline Repository settings file to import.
+   * @see https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/repository-settings-importer-exporter.html
+   */
+  readonly repositorySettings?: Asset;
 }
 
 /**
@@ -657,6 +666,7 @@ export class Repository extends Construct implements IRepository {
       this.installerGroup,
       repositoryInstallationPath,
       props.version,
+      props.repositorySettings,
     );
 
     this.configureSelfTermination();
@@ -887,7 +897,8 @@ export class Repository extends Construct implements IRepository {
   private configureRepositoryInstallerScript(
     installerGroup: AutoScalingGroup,
     installPath: string,
-    version: IVersion) {
+    version: IVersion,
+    settings?: Asset) {
     const installerScriptAsset = ScriptAsset.fromPathConvention(this, 'DeadlineRepositoryInstallerScript', {
       osType: installerGroup.osType,
       baseName: 'installDeadlineRepository',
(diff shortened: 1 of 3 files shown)


ddneilson:
General question... do we want to be doing this via Deadline's Repository installer? We have the ability to hit RCS endpoints in our Lambdas now, so we could do up a CustomResource that imports Repository settings.

I suppose the two are not mutually exclusive... it makes sense to have the ability to import settings to the repository during an initial installation of the repository, but it also may be desirable to be able to programmatically build out repository settings JSON from deployed infrastructure and import those settings post-Repository-install.

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -69,7 +71,17 @@ set +x
 
 INSTALLER_DB_ARGS_STRING=''
 for key in "${!INSTALLER_DB_ARGS[@]}"; do INSTALLER_DB_ARGS_STRING=$INSTALLER_DB_ARGS_STRING"${key} ${INSTALLER_DB_ARGS[$key]} "; done
-$REPO_INSTALLER --mode unattended --setpermissions false --prefix "$PREFIX" --installmongodb false --backuprepo false ${INSTALLER_DB_ARGS_STRING}
+
+REPOSITORY_SETTINGS_ARG_STRING=''
+if [ ! -z "$DEADLINE_REPOSITORY_SETTINGS_FILE" ]; then
+  if [ ! -f "$DEADLINE_REPOSITORY_SETTINGS_FILE" ]; then
+    echo "WARNING: Repository settings file was specified but is not a file: $DEADLINE_REPOSITORY_SETTINGS_FILE. Repository settings will not be imported."
```
ddneilson: I wonder if we should make this case a fail. The warning's in the logs, but it looks like a successful deploy from the CDK CLI (and CloudFormation), but the customer would only know it's a fail if they look at the repository's cloud-init logs.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -376,6 +379,12 @@ export interface RepositoryProps {
    * Options to add additional security groups to the Repository.
    */
   readonly securityGroupsOptions?: RepositorySecurityGroupsOptions;
+
+  /**
+   * The Deadline Repository settings file to import.
+   * @see https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/repository-settings-importer-exporter.html
+   */
+  readonly repositorySettings?: Asset;
```
horsmand: We should describe the default behaviour here too.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -887,7 +897,8 @@ export class Repository extends Construct implements IRepository {
   private configureRepositoryInstallerScript(
     installerGroup: AutoScalingGroup,
     installPath: string,
-    version: IVersion) {
+    version: IVersion,
+    settings?: Asset) {
```
horsmand: The closing parenthasis shoud be on a new line and the `settings` line should have a comma dangle.

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
jericht: That's a good point, changed this to fail the script

packages/aws-rfdk/lib/deadline/lib/repository.ts
jericht: Added

packages/aws-rfdk/lib/deadline/lib/repository.ts
jericht: Fixed

jericht: Fixed

ddneilson (approved):

horsmand (approved):

PR #396, opened 2021-04-19
feat(deadline): add security group property to ubl

### LINKED ISSUE #394: Accept Security Group as a property on UBL construct
The `UsageBasedLicensing` construct should have a `securityGroup` field in its construct props that would accept a security group to be applied to the capacity created by the ECS Cluster for hosting the container running the license forwarder. Currently this capacity will always create a new security group.

A benefit to allowing the security group to be taken as a construct prop is that a user could create and configure their own security groups in a parent stack and they would be persisted and reused if the stack containing the UBL construct was torn down.

---

horsmand:
Fixes #394 

## Notes
This PR adds `securityGroup` as a construct property on the UBL construct to be used by the ASG that gets created to host the container running the license forwarder, similarly to how we accept a `securityGroup` on the RenderQueue construct for the same purpose.

## Testing
A unit test was added to pass a security group into the UBL constructor and confirm that the security group's ID was present in the LaunchConfiguration of the ASG that the UBL construct creates.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/render-queue.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/render-queue.ts
@@ -329,6 +329,8 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
       }],
       updateType: undefined, // Workaround -- See: https://github.com/aws/aws-cdk/issues/11581
       updatePolicy: UpdatePolicy.rollingUpdate(),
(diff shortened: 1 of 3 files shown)


jericht (approved):

ddneilson (approved):
Gotta love the simple ones. Thanks for adding this David.

PR #408, opened 2021-05-03
fix(core): Add security group property to HealthMonitor

### LINKED ISSUE #407: Allow setting pre-defined security group in Health Monitor construct
The health monitor construct creates a application load balancer. This construct creates a new security group(SG) implicitly if it is not provided in constructor properties. This means, a new SG is always created and ingress and egress rules are added automatically based on the target groups created. 

Customer may want to customize this option by setting a SG which was created using console or created in a separate stack.
### Use Case

This is required to set a pre-defined SG to the health monitor construct.

### Proposed Solution

The proposed solution is to first add an optional constructor property to `HealthMonitorProps` ([here](https://github.com/aws/aws-rfdk/blob/a0d7b0aa25be8d957d54f2324bc9c2f7e01438a2/packages/aws-rfdk/lib/core/lib/health-monitor.ts#L174)) and then if this value is provided, pass it to the ALB creation ([here](https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/core/lib/load-balancer-manager.ts#L198-L203))

---

kozlove-aws:
### Problem 
The health monitor construct creates a application load balancer. This construct creates a new security group(SG) implicitly if it is not provided in constructor properties. This means, a new SG is always created and ingress and egress rules are added automatically based on the target groups created.

### Solution
Extended `HealthMonitorProps` with new optional property `securityGroup`.
Passing this new property to `ApplicationLoadBalancer` construct.

### Testing
Added unit test.
Deployed fleet with defined security group and tested that:
- instance was raised and running at least 30 minutes
- after terminating launcher on instance it received status `unhealthy`
- new instance was invoked to reach target capacity
- target capacity set to 0 when both instances had status `unhealthy`

----

### DIFF
--- a/packages/aws-rfdk/lib/core/lib/health-monitor.ts
+++ b/packages/aws-rfdk/lib/core/lib/health-monitor.ts
@@ -14,6 +14,7 @@ import {
 import {SnsAction} from '@aws-cdk/aws-cloudwatch-actions';
 import {
   IConnectable,
+  ISecurityGroup,
   IVpc,
   Port,
   SubnetSelection,
@@ -211,6 +212,13 @@ export interface HealthMonitorProps {
    * @default: The VPC default strategy
    */
   readonly vpcSubnets?: SubnetSelection;
+
+  /**
+   * Security group for the health monitor. This is security group is associated with the health monitor's load balancer.
+   *
+   * @default: A security group is created
+   */
+  readonly securityGroup?: ISecurityGroup;
 }
 
 /**
--- a/packages/aws-rfdk/lib/core/test/health-monitor.test.ts
+++ b/packages/aws-rfdk/lib/core/test/health-monitor.test.ts
@@ -32,6 +32,7 @@ import {
   InstanceSize,
   InstanceType,
   IVpc,
+  SecurityGroup,
   SubnetType,
   Vpc,
 } from '@aws-cdk/aws-ec2';
@@ -565,6 +566,31 @@ describe('HealthMonitor', () => {
         {'Fn::ImportValue': notMatching(stringLike('*PrivateSubnet*'))},
       ],
     }));
+  });
+
+  test('specifying a security group', () => {
+    // GIVEN
+    const securityGroup = new SecurityGroup(infraStack, 'LBSecurityGroup', { vpc });
+    const fleet = new TestMonitorableFleet(wfStack, 'workerFleet', {
+      vpc,
+    });
+
+    // WHEN
+    healthMonitor = new HealthMonitor(hmStack, 'healthMonitor2', {
+      vpc,
+      securityGroup,
+    });
+    healthMonitor.registerFleet(fleet, {});
+
+    // THEN
+    // Make sure it has the security group
+    expectCDK(hmStack).to(haveResourceLike('AWS::ElasticLoadBalancingV2::LoadBalancer', {
+      SecurityGroups: arrayWith(
+        hmStack.resolve(securityGroup.securityGroupId),
+      ),
+    }));
+    // HealthMonitor should not create its own security group
+    expectCDK(hmStack).notTo(haveResource('AWS::EC2::SecurityGroup'));
   });
 
   describe('tagging', () => {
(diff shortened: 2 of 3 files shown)


packages/aws-rfdk/lib/core/test/health-monitor.test.ts
```diff
@@ -567,6 +568,27 @@ describe('HealthMonitor', () => {
...
+    healthMonitor.registerFleet(fleet, {});
+
+    // THEN
+    // Make sure it has the security group
+    expectCDK(hmStack).to(haveResourceLike('AWS::ElasticLoadBalancingV2::LoadBalancer', {
+      SecurityGroups: [
+        {'Fn::GetAtt': arrayWith(stringLike('LBSecurityGroup*'))},
+      ],
+    }));
+  });
```
jusiskin: Just a few suggestions for the test:

1. I think it might make sense to add a `// GIVEN` section and move the `// WHEN` section.
2. We should assert that the `HealthMonitor` does not create its own security group. For this we might need to create the passed-in security group and `fleet` in separate stacks and then check that `hmStack` contains no security groups.
3. The assertion that the SG is associated with the load balancer could be tightened up a little using [`stack.resolve(...)`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html#resolveobj)

Putting that together, how about this:

```ts
// GIVEN
const securityGroup = new SecurityGroup(infraStack, 'LBSecurityGroup', { vpc });
const fleet = new TestMonitorableFleet(wfStack, 'workerFleet', {
  vpc,
});

// WHEN
healthMonitor = new HealthMonitor(hmStack, 'healthMonitor2', {
  vpc,
  securityGroup: new SecurityGroup(hmStack, 'LBSecurityGroup', { vpc }),
});
healthMonitor.registerFleet(fleet, {});

// THEN
// Make sure the health monitor uses the passed-in security group
expectCDK(hmStack).to(haveResourceLike('AWS::ElasticLoadBalancingV2::LoadBalancer', {
  SecurityGroups: arrayWith(
    hmStack.resolve(securityGroup.securityGroupId),
  ),
}));
// HealthMonitor should not create its own security group
expectCDK(hmStack).notTo(haveResource('AWS::EC2::SecurityGroup'));
```

packages/aws-rfdk/lib/core/lib/health-monitor.ts
```diff
@@ -211,6 +212,13 @@ export interface HealthMonitorProps {
    * @default: The VPC default strategy
    */
   readonly vpcSubnets?: SubnetSelection;
+
+  /**
+   * Security group to associate with the load balancer if it will be created.
```
jusiskin: I looked and we do document that the `HealthMonitor` uses a load balancer, but for anyone who skips over that documentation might be confused when seeing this documentation. How about:

```suggestion
   * Security group for the health monitor. This is security group is associated with the health monitor's load
   * balancer.
```

jusiskin (changes requested):
Looks good! just minor feedback for the documentation and test

packages/aws-rfdk/lib/core/lib/health-monitor.ts
kozlove-aws: It is good but I just wanted to highlight here that this security group is used only when we create load balancer. 
We have option to pass LB and this SG will not be used then.

jusiskin: If that is the case, perhaps we want to warn or error when a user supplies both properties to the `HealthMonitor`. As for the TSDoc, how about:

```suggestion
   * Security group for the health monitor. This is security group is associated with the health monitor's load
   * balancer. This security group is only used when a load balancer is not supplied to the health monitor.
```

jusiskin: After discussion offline, we confirmed that `HealthMonitor` does not accept a load balancer as input.

packages/aws-rfdk/lib/core/test/health-monitor.test.ts
jusiskin: Looks good. Resovled

packages/aws-rfdk/lib/core/lib/health-monitor.ts
jusiskin: Resolved

jusiskin (approved):

yashda (approved):

yashda (commented):
Changes looks good to me. Thanks Eugene!

PR #446, opened 2021-05-26
chore(lambdas): add retry logic to ACM certificate importer

jericht:
### Notes
- Adds a class that can be used to back off exponentially, optionally with jitter
- Adds (& refactors) exponential back off & retry logic in the ACM certificate importer lambda, to avoid failures due to the limit of importing 1 certificate per second (see [ACM limits](https://docs.aws.amazon.com/acm/latest/userguide/acm-limits.html)). 

### Testing
- Wrote unit tests
- Built RFDK, deployed the example app, and ensured the ACM importer lambda function for the `RenderQueue` ran successfully
    - Treat this as a sanity check to make sure the regular case still works, since reproducing the error is difficult. The unit test added that tests the retry logic should be sufficient.

----

### DIFF
--- a/packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
@@ -1,0 +1,139 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Properties for BackoffGenerator.
+ */
+export interface BackoffGeneratorProps {
+  /**
+   * The base duration, in milliseconds, used to calculate exponential backoff.
+   *
+   * For example, when not using jitter, the backoff time per attempt will be calculated as:
+   * 1. `base` * 2^0
+   * 2. `base` * 2^1
+   * 3. `base` * 2^2, etc.
+   *
+   * @default 200 milliseconds
+   */
+  readonly base?: number;
+
+  /**
+   * The maximum amount of time, in milliseconds, allowed between backoffs.
+   * @default Number.MAX_SAFE_INTEGER
+   */
+  readonly maxIntervalMs?: number;
+
+  /**
+   * The divisor used to calculate the portion backoff time that will be subject to jitter.
+   * Higher values indicate lower jitters (backoff times will differ by a smaller amount).
+   *
+   * For example, given a calculated `backoff` value, applying jitter would look like:
+   * ```
+   * backoffJitter = (backoff - backoff / jitterDivisor) + jitter(backoff / jitterDivisor)
+   * ```
+   * @default 1 - The entire backoff time is subject to jitter (i.e. "full jitter")
+   */
+  readonly jitterDivisor?: number;
+
+  /**
+   * The maximum amount of time, in milliseconds, to backoff before quitting.
+   * @default No limit on how long this object can backoff for
+   */
+  readonly maxBackoffTimeMs?: number;
+
+  /**
+   * The maximum number of times to backoff before quitting.
+   * @default No limit on how many times this object can backoff
+   */
+  readonly maxAttempts?: number;
+}
+
+/**
+ * Class to handle sleeping with exponential backoff.
+ *
+ * Reference: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+ */
+export class BackoffGenerator
+{
+  /**
+   * Calculates the number of milliseconds to sleep based on the attempt count.
+   * @param b The base value for the calculation.
+   * @param attempt The attempt count.
+   * @param maxIntervalMs The maximum interval between backoffs, in milliseconds.
+   * @returns The number of milliseconds to sleep.
+   */
+  private static calculateSleepMs(b: number, attempt: number, maxIntervalMs: number): number {
+    return Math.min(b * Math.pow(2, attempt), maxIntervalMs, Number.MAX_SAFE_INTEGER);
+  }
+
+  private readonly base: number;
+  private readonly maxIntervalMs: number;
+  private readonly jitterDivisor: number;
+  private readonly maxBackoffTimeMs: number | undefined;
+  private readonly maxAttempts: number | undefined;
+
+  private sleepTime: number = 0;
+  private attempt: number = 0;
+
+  constructor(props?: BackoffGeneratorProps)
+  {
+    this.maxBackoffTimeMs = props?.maxBackoffTimeMs;
+    this.maxAttempts = props?.maxAttempts;
+    this.base = props?.base ?? 200;
+    this.maxIntervalMs = props?.maxIntervalMs ?? Number.MAX_SAFE_INTEGER;
+
+    this.jitterDivisor = props?.jitterDivisor ?? 1;
+    if (this.jitterDivisor <= 0) {
+      throw new Error(`jitterDivisor must be a postive integer, got: ${this.jitterDivisor}`);
+    }
+
+    // Initialize internal counters
+    this.restart();
+  }
+
+  /**
+   * Restarts the internal counters used by this class.
+   */
+  public restart(): void {
+    this.sleepTime = 0;
+    this.attempt = 0;
+  }
+
+  /**
+   * Sleeps for an exponentially increasing time depending on how many times this class has backed off.
+   */
+  public async backoff(): Promise<void> {
+    const interval = BackoffGenerator.calculateSleepMs(this.base, this.attempt, this.maxIntervalMs);
+
+    await sleep(interval);
+    this.sleepTime += interval;
+    this.attempt++;
+  }
+
+  /**
+   * Sleeps for an exponentially increasing time (with jitter) depending on how many times this class has backed off.
+   */
+  public async backoffJitter(): Promise<void> {
+    const interval = BackoffGenerator.calculateSleepMs(this.base, this.attempt, this.maxIntervalMs);
+    const intervalJitter = (interval - interval / this.jitterDivisor) + (Math.floor(interval / this.jitterDivisor * Math.random()));
+
+    await sleep(intervalJitter);
+    this.sleepTime += intervalJitter;
+    this.attempt++;
+  }
+
+  /**
+   * Returns true if either the maximum number of attempts or maximum time span has been reached.
+   * If neither are specified, this will always return true.
+   */
+  public shouldContinue(): boolean {
+    return ( this.maxAttempts === undefined || this.attempt < this.maxAttempts ) &&
+           ( this.maxBackoffTimeMs === undefined || this.sleepTime < this.maxBackoffTimeMs );
+  }
+}
+
+async function sleep(ms: number): Promise<void> {
+  return new Promise(res => setTimeout(res, ms));
+}
--- a/packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
@@ -21,6 +21,7 @@ export interface BackoffGeneratorProps {
 
   /**
    * The maximum amount of time, in milliseconds, allowed between backoffs.
+   * Each backoff will have its length clamped to a maximum of this value.
    * @default Number.MAX_SAFE_INTEGER
    */
   readonly maxIntervalMs?: number;
@@ -33,15 +34,15 @@ export interface BackoffGeneratorProps {
    * ```
    * backoffJitter = (backoff - backoff / jitterDivisor) + jitter(backoff / jitterDivisor)
    * ```
-   * @default 1 - The entire backoff time is subject to jitter (i.e. "full jitter")
+   * @default No jitter
    */
   readonly jitterDivisor?: number;
 
   /**
-   * The maximum amount of time, in milliseconds, to backoff before quitting.
+   * The maximum cumulative time, in milliseconds, to backoff before quitting.
    * @default No limit on how long this object can backoff for
    */
-  readonly maxBackoffTimeMs?: number;
+  readonly maxCumulativeBackoffTimeMs?: number;
 
   /**
    * The maximum number of times to backoff before quitting.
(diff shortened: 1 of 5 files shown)


ddneilson (dismissed):
Looks solid to me. Thanks Jericho!

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
```diff
@@ -0,0 +1,139 @@
...
+   * ```
+   * @default 1 - The entire backoff time is subject to jitter (i.e. "full jitter")
+   */
+  readonly jitterDivisor?: number;
+
+  /**
+   * The maximum amount of time, in milliseconds, to backoff before quitting.
+   * @default No limit on how long this object can backoff for
+   */
+  readonly maxBackoffTimeMs?: number;
```
jusiskin: I believe this is meant as cumulative backoff time, right? It might help to spell that out here and even rename the variable to reflect that. What do you think?

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
```diff
@@ -0,0 +1,139 @@
...
+   *
+   * @default 200 milliseconds
+   */
+  readonly base?: number;
+
+  /**
+   * The maximum amount of time, in milliseconds, allowed between backoffs.
+   * @default Number.MAX_SAFE_INTEGER
+   */
+  readonly maxIntervalMs?: number;
```
jusiskin: From this doc string, it's not clear how this maximum is imposed. Based on the code, it looks like the backoff time is clamped to a maximum of this value (as opposed to aborting the entire backoff attempts). Can we mention this?

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
```diff
@@ -0,0 +1,139 @@
...
+    return Math.min(b * Math.pow(2, attempt), maxIntervalMs, Number.MAX_SAFE_INTEGER);
+  }
+
+  private readonly base: number;
+  private readonly maxIntervalMs: number;
+  private readonly jitterDivisor: number;
+  private readonly maxBackoffTimeMs: number | undefined;
+  private readonly maxAttempts: number | undefined;
+
+  private sleepTime: number = 0;
```
jusiskin: nit: how about renaming this to `cumulativeBackoffTimeMs`?

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
```diff
@@ -0,0 +1,139 @@
...
+   * Sleeps for an exponentially increasing time (with jitter) depending on how many times this class has backed off.
+   */
+  public async backoffJitter(): Promise<void> {
+    const interval = BackoffGenerator.calculateSleepMs(this.base, this.attempt, this.maxIntervalMs);
+    const intervalJitter = (interval - interval / this.jitterDivisor) + (Math.floor(interval / this.jitterDivisor * Math.random()));
+
+    await sleep(intervalJitter);
+    this.sleepTime += intervalJitter;
+    this.attempt++;
+  }
```
jusiskin: From an API perspective, perhaps it makes sense unify these methods and have the presence of the `jitterDivisor` constructor property dictate whether to apply jitter.

Right now, someone could use this in unforeseen ways:

```ts
const backoffGen = new BackoffGenerator(...);
backoffGen.backoff();
backoffGen.backoffJitter();
backoffGen.backoff();
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Returns true if either the maximum number of attempts or maximum time span has been reached.
+   * If neither are specified, this will always return true.
+   */
+  public shouldContinue(): boolean {
+    return ( this.maxAttempts === undefined || this.attempt < this.maxAttempts ) &&
+           ( this.maxBackoffTimeMs === undefined || this.sleepTime < this.maxBackoffTimeMs );
+  }
+}
```
jusiskin: Could this be returned by the promise returned by `backoff()` as a convenience? I'm thinking the API could be simplified to:

```ts
const backoffGenerator = new BackoffGenerator(...);
do {
  const result = attempt();
} while (!result.succeeded && await backoffGenerator.backoff());
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/test/backoff-generator.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    // WHEN
+    for (let i = 0; i < maxAttempts; i++) {
+      const promise = backoffGenerator.backoff();
+      jest.advanceTimersByTime(base * Math.pow(2, i));
+      await promise;
+    }
+
+    // THEN
+    expect(backoffGenerator.shouldContinue()).toBe(false);
```
jusiskin, line 53: looks like we don't have any tests for when `.shouldContinue()` to be `true`

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/test/backoff-generator.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    // This reaches maxAttempts, shouldContinue() will return false
+    const promise = backoffGenerator.backoff();
+    jest.advanceTimersByTime(base);
+    await promise;
+
+    // WHEN
+    backoffGenerator.restart();
+
+    // THEN
+    expect(backoffGenerator.shouldContinue()).toBe(true);
```
jusiskin: There is another behavior that we should assert for the restart: that the max cumulative backoff is properly reset.

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
```diff
@@ -0,0 +1,459 @@
...
+      X509CertificatePem: {
+        Cert: 'cert',
+        CertChain: 'certChain',
+        Key: 'key',
+        Passphrase: 'passphrase',
+      },
+    };
+
+    beforeEach(() => {
+      sinon.stub(Certificate, 'decryptKey').returns(new Promise((res, _rej) => res('key')));
```
jusiskin: nit: minor simplification

```suggestion
      sinon.stub(Certificate, 'decryptKey').returns(Promise.resolve('key'));
```

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
```diff
@@ -0,0 +1,459 @@
...
+      let error = undefined;
+      try {
+        await importer.doCreate(physicalId, doCreateProps);
+      } catch(e) {
+        error = e;
+      }
+
+      // THEN
+      expect(error).toBeDefined();
+      expect(error.message).toMatch(/Secret .* did not contain a SecretString as expected/);
```
jusiskin: nit: we can use some jest sugar for this. I think it could look like:

```ts
// WHEN
const promise = importer.doCreate(physicalId, doCreateProps);

// THEN
expect(promise).rejects.toThrow(/Secret .* did not contain a SecretString as expected/);
```

Same suggestion for other error message assertions below. Not critical, but what do you think?

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
```diff
@@ -0,0 +1,459 @@
...
+        expect(getItemStub.calledOnce).toBe(true);
+      });
+
+      test('throws if certificate not found in ACM', async () => {
+        // GIVEN
+        const resourceTable = new MockCompositeStringIndexTable();
+        const getItemStub = sinon.stub(resourceTable, 'getItem').resolves({ ARN: certArn });
+
+        const getCertificateFake = sinon.fake.resolves({});
+        AWSMock.mock('ACM', 'getCertificate', getCertificateFake);
```
jusiskin: Does `ACM.getCertificate(...)` return a resolved promise with an empty object when no certificate is found? What is the behavior our logic is using to determine if a certificate exists using this API?

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
```diff
@@ -0,0 +1,459 @@
...
+        });
+
+        // WHEN
+        await expect(importer.doCreate(physicalId, doCreateProps))
+
+        // THEN
+          .resolves.toEqual({ CertificateArn: certArn });
+        expect(getItemStub.calledOnce).toBe(true);
+        expect(getCertificateFake.calledOnce).toBe(true);
+        expect(importCertificateFake.calledOnce).toBe(true);
```
jusiskin: I'm probably missing something, but I am confused by this assertion. Why do we import a cert if it already exists in ACM?

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
```diff
@@ -0,0 +1,459 @@
...
+    AWSMock.setSDKInstance(AWS);
+  });
+
+  afterEach(() => {
+    process.env = oldEnv;
+    sinon.restore();
+    AWSMock.restore();
+  });
+
+  describe('doCreate', () => {
```
jusiskin, line 42: can we add a test for the return value of this function?

jusiskin (changes requested):
Looks great. Main things are a some missing test coverage, but I also threw in some minor optional code clean-up suggestions.

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jericht: Yes, I've renamed this and updated the description as well.

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jericht: Good catch, updated the docstring.

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jericht: Yes, I've updated the variable name and docstring

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jericht: Renamed

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jericht: Good point, I've removed the `backoffJitter` method and now conditionally apply jitter in `backoff()`

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jericht: Good idea! Added this in.

This also made me think about how the class should handle calls to `backoff()` when the conditions to stop backing off have already been met. Right now, it will continue backing off regardless of those conditions and the only way to know if it should stop is to call `shouldContinue()`, This probably isn't what most users would expect to happen by default, so I've also changed `backoff()` to first check if it should even backoff according to the "stop" conditions, with an option to force backoffs regardless.

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/test/backoff-generator.test.ts
jericht: Thanks for catching this, added it in.

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/test/backoff-generator.test.ts
jericht: Added a test for the max cumulative backoff time as well.

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
jericht: We do assert the return value of this function in `new > imports certificate` and `existing > imports certificate`. I'm not sure if there's a way to create a separate test that focuses on that return value without it being very similar to one of those existing tests. Do you have something in mind?

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
jericht: Neat, thanks!

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
jericht: That's a lot nicer, changed these tests to use this.

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
jericht: We check that the returned object has a `Certificate` field in it, otherwise we throw this error.

However, after testing this locally, it turns out that was incorrect. If the ARN does not reference a valid ACM certificate, `getCertificate` will reject the promise with an error object like:

```jsonc
{
  "message": "Could not find certificate <cert-arn>.",
  "code": "ResourceNotFoundException",
  // etc.
}
````

We should instead be catching that error in a try/catch, so I've updated the ACM handler to properly handle this error case.

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
jericht: This is to [replace/update the certificate](https://docs.aws.amazon.com/acm/latest/APIReference/API_ImportCertificate.html#ACM-ImportCertificate-request-CertificateArn), which is useful when we want to rotate certs.

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jusiskin: Thanks!

jusiskin: Thanks!

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jusiskin: 👍

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jusiskin: Looks great! Thanks

packages/aws-rfdk/lib/lambdas/nodejs/lib/backoff-generator/backoff-generator.ts
jusiskin: Much cleaner! looks great

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
jusiskin: Great - thanks for confirming that!

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
jusiskin: Ah thanks for clarifying that. Can we add a comment or rename the test to help clarify this?

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/test/acm-handlers.test.ts
jusiskin: I missed those. Thanks for pointing that out. Resolved

jusiskin (dismissed):
Great work, @jericht! Approved

jusiskin (approved):

ddneilson (approved):

PR #451, opened 2021-05-27
fix(deadline): allow zero-sized WorkerInstanceFleet

### LINKED ISSUE #442: On-Demand worker fleets time out during deployment if min_capacity is set to 0
Deploying a CDK application with an On-Demand worker fleet that has its min_capacity set to zero will result in the ASG timing out and the deployment failing.

### Reproduction Steps
1. Launch an EC2 instance with the latest version of Nimble StudioBuilder
2. Create a valid configuration for a studio. Add an On-Demand worker fleet (OS shouldn't matter) and set its minimum instance capacity to 0
3. Deploy the studio

(Launching a more limited application with just the worker fleet would almost certainly repro the same issue, but I haven't tested that myself)

### Error Log

```
2021-05-24 13:25:45 UTC-0700 | WorkerFleetondemandlinuxASG254059E0 | &nbsp;
CREATE_FAILED | Received 0 SUCCESS signal(s) out of 1.  Unable to satisfy 100% MinSuccessfulInstancesPercent requirement
```

### Environment

  - **CDK CLI Version  :** 1.99.0
  - **CDK Framework Version:**
  - **RFDK Version:** 0.30.0
  - **Deadline Version:**
  - **Node.js Version:** 
  - **OS               :**
  - **Language (Version):** 

### Other

I believe I've found the root cause of the issue:

[Here](https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/deadline/lib/worker-fleet.ts#L461) is where the RFDK defines on-demand worker fleets. Note that it will always set `resourceSignalTimeout`

But as the [CDK documentation](https://docs.aws.amazon.com/cdk/api/latest/python/aws_cdk.aws_autoscaling/AutoScalingGroup.html) notes for the `resourceSignalCount` field:
>  How many ResourceSignal calls CloudFormation expects before the resource is considered created. Default: 1 if resourceSignalTimeout is set, 0 otherwise

Because the timeout is always being set, the ASG will *always* expect one instance to send a resource signal when it completes. But if min capacity is zero, then there's no instance, and no signal, and the ASG times out.

--- 

This is :bug: Bug Report

ddneilson:
Fixes: https://github.com/aws/aws-rfdk/issues/442

### Testing
* Deployed the all-in basic example with worker fleet min capacity of 0 and 1. Both successful.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
+++ b/packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
@@ -7,7 +7,6 @@
 
 import {
   ABSENT,
-  arrayWith,
   expect as expectCDK,
   haveResource,
   haveResourceLike,
@@ -1890,7 +1889,7 @@ describe('tagging', () => {
 
 test('worker fleet signals when non-zero minCapacity', () => {
   // WHEN
-  new WorkerInstanceFleet(wfstack, 'workerFleet', {
+  const fleet = new WorkerInstanceFleet(wfstack, 'workerFleet', {
     vpc,
     workerMachineImage: new GenericWindowsImage({
       'us-east-1': 'ami-any',
@@ -1899,18 +1898,11 @@ test('worker fleet signals when non-zero minCapacity', () => {
     minCapacity: 1,
   });
 
+  // WHEN
+  const userData = fleet.fleet.userData.render();
+
   // THEN
-  // Make sure the `cfn-signal` is present
-  expectCDK(wfstack).to(haveResourceLike('AWS::AutoScaling::LaunchConfiguration', {
-    UserData: {
-      'Fn::Base64': {
-        'Fn::Join': [
-          '',
-          arrayWith('<powershell>trap {\n$success=($PSItem.Exception.Message -eq "Success")\ncfn-signal --stack workerFleetStack --resource workerFleetASG25520D69 --region us-east-1 --success ($success.ToString().ToLower())\nbreak\n}\nmkdir (Split-Path -Path \'C:/temp/'),
-        ],
-      },
-    },
-  }));
+  expect(userData).toContain('cfn-signal');
   expectCDK(wfstack).to(haveResourceLike('AWS::AutoScaling::AutoScalingGroup', {
     CreationPolicy: {
       ResourceSignal: {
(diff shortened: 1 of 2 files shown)


packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
```diff
@@ -1884,3 +1887,63 @@ describe('tagging', () => {
...
+  new WorkerInstanceFleet(wfstack, 'workerFleet', {
+    vpc,
+    workerMachineImage: new GenericWindowsImage({
+      'us-east-1': 'ami-any',
+    }),
+    renderQueue,
+    minCapacity: 0,
+  });
+
+  // THEN
```
ddneilson, line 1931: Hrm. I should make sure the warning is emitted as well... revision coming.

jericht (approved):

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
```diff
@@ -1884,3 +1886,58 @@ describe('tagging', () => {
...
+  expect(userData).not.toContain('cfn-signal');
+  // Make sure we don't have a CreationPolicy
+  expectCDK(wfstack).notTo(haveResourceLike('AWS::AutoScaling::AutoScalingGroup', {
+    CreationPolicy: objectLike({}),
+  }, ResourcePart.CompleteDefinition));
+  // There should be a warning in the construct's metadata about deploying with no capacity.
+  expect(fleet.node.metadataEntry).toHaveLength(3);
+  // [0] = warning about block devices. [2] = warning about no health monitor
+  expect(fleet.node.metadataEntry[1].type).toMatch(ArtifactMetadataEntryType.WARN);
+  expect(fleet.node.metadataEntry[1].data).toMatch(/Deploying with 0 minimum capacity./);
```
jusiskin, line 1942: While the comments help, hard-coding the expected indexes of `metadataEntry` makes the tests fragile. The  assertions depend on side-effects of other behavior in the `WorkerInstanceFleet` construct. I believe we can use the following jest matchers to make this more robust:

*   [`exepect.arrayContaining(...)`](https://jestjs.io/docs/expect#expectarraycontainingarray)
*   [`expect.stringMatching(...)`](https://jestjs.io/docs/expect#expectstringmatchingstring--regexp)

Which would look like:

```ts
expect(fleet.node.metadataEntry).toEqual(expect.arrayContaining([
  {
    type: ArtifactMetaDataEntryType.WARN,
    data: expect.stringMatching(/Deploying with 0 minimum capacity\./),
  },
]));
```

In the case of ensuring the warning is not emitted, we could do:

```ts
expect(fleet.node.metadataEntry).not.toEqual(expect.arrayContaining([
  {
    type: ArtifactMetaDataEntryType.WARN,
    data: expect.stringMatching(/Deploying with 0 minimum capacity\./),
  },
]));
```

This being said, the tests currently pass, so I won't hold up the merge on it. We can change this later if these tests ever break.

jusiskin (approved):
Approved. Just a little testing advice for the future, but I'm fine with merging this as-is.

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
ddneilson: Oh, I like that!

PR #456, opened 2021-05-31
fix(integ): make exit code reflect test success/failure

jusiskin:
# Problem

## 1. Exit Code

The exit code returned by the `integ/scripts/bash/rfdk-integ-e2e.sh` script does not properly reflect the result of the integration test. When the test fails, the script can return a exit code of `0`. RFDK release automation relies on the exit code reflecting the result of the test.

## 2.  Error Handling

The error handling in the bash scripting of the RFDK e2e test framework does not work as expected. When there is a synthesis or deploy failure, the scripting continues processing and attempts to execute the tests. This produces strange errors in the test output that can cause confusion and cause readers to misinterpret why a test failed.

## 3. Pre-interaction Hook

The RFDK e2e integration test framework allows specifying a `PRE_AWS_INTERACTION_HOOK` environment variable that points to a bash function. It is supposed to be invoked by the framework before any AWS APIs are interacted with. This hook was not being invoked before synthesizing each test component's CDK app. CDK makes some API calls during synthesis if the CDK app makes use of [CDK context methods](https://docs.aws.amazon.com/cdk/latest/guide/context.html#context_methods), which the apps do use.

# Solution

## 1. Exit Code

The various bash scripts were adjusted so that exit codes are bubbled up to the component driver script (`component-e2e-driver.sh`) and returned by that script. This uses the first non-zero exit code in the following order:

*   Synthesis failure
*   Deploy failure
*   Test failure
*   Destroy failure

If all of these steps succeed, then `component-e2e-driver.sh` will have an exit code of `0`.

The entry-point script (`rfdk-integ-e2e.sh`), uses `xargs` to call `component-e2e-driver.sh`. This command returns an exit code of `123` if any sub-process it spawns returns an exit code between 1-125. If a sub-process returns a different non-zero exit code, it returns an exit code of `1` ([docs](https://man7.org/linux/man-pages/man1/xargs.1.html#EXIT_STATUS)).

## 2. Error Handling

After some debugging, it was determined that the following statements were the cause of error:

https://github.com/aws/aws-rfdk/blob/0cc67238645e24805590412776d262c8e6b9ec49/integ/components/deadline/common/scripts/bash/component_e2e.sh#L32-L33

Each of the above lines is considered a bash "list". The LHS of these lists is a call to a bash function (`deploy_component_stacks` and `execute_component_test` respectively). Those functions had assumed that the bash `set -e` option was set and any error would short-circuit the function execution and return a non-zero exit code. According to the [`bash` documentation](https://man7.org/linux/man-pages/man1/bash.1.html) for the `-e` option:

>   Exit immediately if a pipeline (which may consist of a single simple command), a list, or a compound command (see `SHELL GRAMMAR` above), exits with a non-zero status.  The shell does not exit if the command that fails is part of the command list immediately following a `while` or `until` keyword, part of the test following the `if` or `elif` reserved words, part of any command executed in a `&&` or `||` list except the command following the final `&&` or `||`, any command in a pipeline but the last, or if the command's return value is being inverted with `!`.

A minimal reproducer was created:

```sh
function bar() {
  set +e
  echo "shell opts: $-"
  false
  echo "I'm immortal"
}
```

Then:

```sh
$ # With outer error-handling enabled
$ set -e
$ bar || echo 'I'm mortal'
shell opts: himBH
I'm immortal
$ # With outer error-handling disabled
$ set +e
$ bar || echo 'I'm mortal' 
shell opts: himBH
I'm immortal
```

Based on this conclusion, the `component-e2e.sh` was modified to move away from a bash "list" and instead disable the `-e` flag and handling errors manually.

## 3. Pre-interaction Hook

Added a call to `run_aws_interaction_hook` before each CDK app is synthesized.

## Other Changes

- output timestamps for live progress logging
- make output more consistent
- clean up format of test report
- show synth log on synth failures

# Testing

- Ran the tests and confirmed that both synth and deploy failures cause both:
    - Short-circuit the test execution
    - Cause the outer script to return a non-zero exit code
- Ran the tests and confirmed that a fully successful test synth/deploy/test/destroy returns a zero exit code

----

### DIFF
--- a/integ/components/deadline/common/scripts/bash/deploy-utils.sh
+++ b/integ/components/deadline/common/scripts/bash/deploy-utils.sh
@@ -110,12 +110,12 @@ function destroy_component_stacks () {
 
     echo "$(timestamp) [${COMPONENT_NAME}] -> [${stack}] stack destroy started"
     npx cdk destroy --app cdk.out -e -f "${stack}" &>> "${destroy_log_path}"
-    STACK_DEPLOY_EXIT_CODE=$?
-    if [[ $STACK_DEPLOY_EXIT_CODE -ne 0 ]]
+    STACK_DESTROY_EXIT_CODE=$?
+    if [[ $STACK_DESTROY_EXIT_CODE -ne 0 ]]
     then
       echo "$(timestamp) [${COMPONENT_NAME}] -> [${stack}] stack destroy failed"
       echo "$(timestamp) [${COMPONENT_NAME}] app destroy failed"
-      return $STACK_DEPLOY_EXIT_CODE
+      return $STACK_DESTROY_EXIT_CODE
     fi
     echo "$(timestamp) [${COMPONENT_NAME}] -> [${stack}] stack destroy complete"
   done
--- a/integ/components/deadline/common/scripts/bash/deploy-utils.sh
+++ b/integ/components/deadline/common/scripts/bash/deploy-utils.sh
@@ -21,17 +21,30 @@ function ensure_component_artifact_dir () {
 function deploy_component_stacks () {
   COMPONENT_NAME=$1
 
-  echo "[${COMPONENT_NAME}] started"
+  echo "$(timestamp) [${COMPONENT_NAME}] app deployment started"
 
   ensure_component_artifact_dir "${COMPONENT_NAME}"
 
   # Generate the cdk.out directory which includes a manifest.json file
   # this can be used to determine the deployment ordering
-  echo "[${COMPONENT_NAME}] synthesizing started"
-  npx cdk synth &> "${INTEG_TEMP_DIR}/${COMPONENT_NAME}/synth.log"
-  echo "[${COMPONENT_NAME}] synthesizing complete"
+  echo "$(timestamp) [${COMPONENT_NAME}] synthesizing started"
+  # Synthesis requires AWS API calls for the context methods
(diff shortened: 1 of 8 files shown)


ddneilson (dismissed):
Great writeup. Thanks for fixing this, Josh.

integ/components/deadline/common/scripts/bash/deploy-utils.sh
```diff
@@ -68,45 +91,38 @@ function execute_component_test () {
...
   # Empty the destroy log file in case it was non-empty
   cp /dev/null "${destroy_log_path}"
   for stack in $(cdk_stack_destroy_order); do
     run_aws_interaction_hook
 
-    echo "[${COMPONENT_NAME}] -> [${stack}] stack destroy started"
+    echo "$(timestamp) [${COMPONENT_NAME}] -> [${stack}] stack destroy started"
     npx cdk destroy --app cdk.out -e -f "${stack}" &>> "${destroy_log_path}"
-    echo "[${COMPONENT_NAME}] -> [${stack}] stack destroy complete"
+    STACK_DEPLOY_EXIT_CODE=$?
```
jericht: nit: variable name
```suggestion
    STACK_DESTROY_EXIT_CODE=$?
```

jericht (changes requested):
Just one tiny comment, otherwise it looks great!

ddneilson (approved):

jericht (approved):

PR #459, opened 2021-06-02
fix: remove support for Node.JS 10.x

### LINKED ISSUE #418: Drop Node10, add Node16
Node.JS 10 is officially EOL, and Node 16 has been officially released -- https://nodejs.org/en/about/releases/

We should drop Node10 from our CI and support, and add Node16.

kozlove-aws:
### Problem
Node.js 10 is not supported anymore since April 10, 2021
We need to update it with newer library.

### Solution 
Replaced nodejs10 with nodejs14 in Docker file for lambda layer.
Also was changed to take `amazonlinux` from `public.ecr` for avoiding limit hit.

### Testing
Was build image and deployed on local account.
Was verified that lambda layer is using `nodejs12.x, nodejs14.x`

----


ddneilson (changes requested):
Looks like Node 10 references here that should also be updated

* https://github.com/aws/aws-rfdk/blob/8a55f32124e2cfbadb33437d2d0494580a9eebac/integ/package.json#L110
* https://github.com/aws/aws-rfdk/blob/8a55f32124e2cfbadb33437d2d0494580a9eebac/packages/aws-rfdk/package.json#L210
* https://github.com/aws/aws-rfdk/blob/8a55f32124e2cfbadb33437d2d0494580a9eebac/tools/pkglint/lib/rules.ts#L632
* https://github.com/aws/aws-rfdk/blob/8a55f32124e2cfbadb33437d2d0494580a9eebac/tools/cdk-build-tools/package.json#L73

ddneilson (dismissed):

tools/pkglint/lib/rules.ts
```diff
@@ -376,11 +376,11 @@ export class NodeCompatibility extends ValidationRule {
 
   public validate(pkg: PackageJson): void {
     const atTypesNode = pkg.getDevDependency('@types/node');
-    if (atTypesNode && !atTypesNode.startsWith('^10.')) {
+    if (atTypesNode && !atTypesNode.startsWith('^12.')) {
       pkg.report({
         ruleName: this.name,
         message: `packages must support node version 10 and up, but ${atTypesNode} is declared`,
```
horsmand:
```suggestion
        message: `packages must support node version 12 and up, but ${atTypesNode} is declared`,
```

horsmand (changes requested):
Looks good, I only found one spot where we're referencing node 10 instead of node 12 in a message that pkglint prints out. I'm also wondering if we should be running `npx yarn install` so it can update the `yarn.lock` file. I see this line in there:
```
"@types/node" "^10.17.60"
```

tools/pkglint/lib/rules.ts
kozlove-aws: Good catch. `yarn.lock` is updated now.

horsmand: Did you include your changes in that most recent push? I don't see this rule or the yarn.lock updated.

kozlove-aws: You are right. Forgot to add files to commit.

horsmand (changes requested):
This looks good, I don't see anything that needs updating, but it seems you have a conflict in your yarn.lock so you'll need to rebase your changes and push them. I'll approve after that.

horsmand (approved):

jericht (approved):
Looks good!

I only see the one reference to `"@types/node@^10.17.60"` that @horsmand [mentioned](https://github.com/aws/aws-rfdk/pull/459#pullrequestreview-685709529) in our `yarn.lock` file and it looks like it's a [dependency of `@aws-cdk/cloudformation-diff`](https://github.com/aws/aws-rfdk/blob/18a8098944f3599446c0f319eaf46b383724cfc3/yarn.lock#L780-L791). We can force it to resolve to `@types/node@12.x.x` by adding an entry in the [resolutions in the root `package.json`](https://github.com/aws/aws-rfdk/blob/18a8098944f3599446c0f319eaf46b383724cfc3/package.json#L112-L119):
```
  "resolutions": {
    // ...
    "**/@aws-cdk/cloudformation-diff/**/@types/node": "^12.8.3"
  },
```

These are just the typings though, so not really necessary. I checked out `@aws-cdk/cloudformation-diff`, changed it to depend on `"@types/node": "^12.18.3"`, then built and tested it and it succeeded, so I think it's safe to assume it's not relying on things that are incompatible between Node 10 and 12.

jusiskin (approved):

PR #461, opened 2021-06-07
feat(core): add FSx for Lustre integration

jericht:
### Notes
- Adds support to mount FSx for Lustre filesystems onto Linux instances, and integration of FSxL with the Deadline `Repository` construct, as described here: https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html

### Testing
- Deployed most Linux distribution/versions (at least 1 per distribution) from the above link and verified that they are able to install the Lustre client and mount an FSxL.
- Deployed the `Repository` construct with an FSxL as the backing filesystem and ensured the repository files were owned by UID:GID=1000:1000 and that the RCS had access to them

I deployed & tested:
- Amazon Linux 1
- Amazon Linux 2
- RHEL 7.5
- RHEL 7.6
- RHEL 7.7
- RHEL 7.8
- RHEL 7.9
- RHEL 8.2
- RHEL 8.3
- CentOS 7.7
- Ubuntu 16.04
- Ubuntu 18.04
- Ubuntu 20.04
- SUSE 12 SP5
- Amazon Linux 2 (ARM)
- Ubuntu 18.04 (ARM)
- Ubuntu 20.04 (ARM)

I was not able to test the following:
- SUSE 12 SP3 & SP4
    - I wasn't able to find AMIs with these versions of SUSE Linux and wasn't able to find a way to downgrade from 12 SP5. Only alternative I know of is to install the operating system from scratch.
- RHEL 7.7, 7.8, and 7.9 (ARM)
    - I wasn't able to find AMIs with these versions and wasn't able to find a way to upgrade a RHEL 7.6 ARM-based AMI to the above versions.

----


packages/aws-rfdk/lib/core/lib/mountable-fsx-lustre.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as path from 'path';
+import {
+  OperatingSystemType,
+  Port,
+} from '@aws-cdk/aws-ec2';
+import * as fsx from '@aws-cdk/aws-fsx';
```
horsmand: Any reason to do this import like this? I don't think it's a problem, but the rest of the imports are more specific.

packages/aws-rfdk/lib/core/lib/mountable-fsx-lustre.ts
```diff
@@ -0,0 +1,125 @@
...
+    mountScriptAsset.grantRead(target.grantPrincipal);
+    const mountScript: string = target.userData.addS3DownloadCommand({
+      bucket: mountScriptAsset.bucket,
+      bucketKey: mountScriptAsset.s3ObjectKey,
+    });
+
+    const mountDir: string = path.posix.normalize(mount.location);
+    const mountOptions: string[] = [ MountPermissionsHelper.toLinuxMountOption(mount.permissions) ];
+    if (this.props.extraMountOptions) {
+      mountOptions.push( ...this.props.extraMountOptions);
```
horsmand:
```suggestion
      mountOptions.push(...this.props.extraMountOptions);
```

packages/aws-rfdk/lib/core/lib/mountable-fsx-lustre.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+    const mountOptionsStr: string = mountOptions.join(',');
+
+    let mountName = this.props.filesystem.mountName;
+    if (this.props.fileset) {
+      mountName = path.posix.join(mountName, this.props.fileset);
+    }
```
horsmand:
```suggestion
    const mountName = this.props.fileset ?  path.posix.join(this.props.filesystem.mountName, this.props.fileset) : this.props.filesystem.mountName;
```

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,264 @@
...
+        sudo sed -i "s#7#${replace_ver}#" /etc/yum.repos.d/aws-fsx.repo
+      fi
+
+      sudo yum clean all
+      sudo yum install -y kmod-lustre-client lustre-client
+    ;;
+
+    # RHEL 8.2 or newer
+    # -----------------
+    8.2|8.[1-9]*([0-9]))
```
horsmand: The first half of this matches only 8.2, but then the second half looks like it would match 8.1 and up. Was 8.1 meant to be included in this?
If we don't want 8.1, maybe it could be:
```suggestion
    8.[2-9]|8.[1-9]+([0-9]))
```

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,264 @@
...
+    # RHEL 7.5 and 7.6
+    # ----------------
+    7.5|7.6)
+      case $kernel_version in
+        ${rhel_kernel_versions[7.5]}) lustre_version="2.10.5";;
+        ${rhel_kernel_versions[7.6]}) lustre_version="2.10.8";;
+        *) echo "ERROR: Unsupported kernel version $kernel_version for RHEL $rhel_version"; exit 1;;
+      esac
+      sudo yum -y install "https://downloads.whamcloud.com/public/lustre/lustre-${lustre_version}/el7/client/RPMS/x86_64/kmod-lustre-client-${lustre_version}-1.el7.x86_64.rpm"
+      sudo yum -y install "https://downloads.whamcloud.com/public/lustre/lustre-${lustre_version}/el7/client/RPMS/x86_64/lustre-client-${lustre_version}-1.el7.x86_64.rpm"
```
horsmand, line 122: How come we're installing 2 clients? Maybe add a comment

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,264 @@
...
+      sudo yum clean all
+      sudo yum install -y kmod-lustre-client lustre-client
+    ;;
+
+    # RHEL 8.2 or newer
+    # -----------------
+    8.2|8.[1-9]*([0-9]))
+      case $kernel_version in
+        ${rhel_kernel_versions[8.2]}) replace_ver="8.2";;
+        ${rhel_kernel_versions[8.3]}  replace_ver="8.3";;
```
horsmand: There is no 8.3 in the array. I also see the RHEL article you linked to has an entry for 8.4, should we add that too? Will we need to manually update this whenever a new version of RHEL is released?

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,264 @@
...
+  fi
+
+  sudo yum install -y lustre-client
+}
+
+function install_on_rhel() {
+  # Create map of RHEL release to kernel version shell patterns
+  # See https://access.redhat.com/articles/3078
+  declare -A rhel_kernel_versions
+  rhel_kernel_versions=(
```
horsmand, line 85: It isn't clear what we're going to use this for, can you add a comment explaining?

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,264 @@
...
+          fi
+        ;;
+      esac
+
+      curl https://fsx-lustre-client-repo-public-keys.s3.amazonaws.com/fsx-rpm-public-key.asc -o /tmp/fsx-rpm-public-key.asc
+      sudo rpm --import /tmp/fsx-rpm-public-key.asc
+      sudo curl https://fsx-lustre-client-repo.s3.amazonaws.com/el/8/fsx-lustre-client.repo -o /etc/yum.repos.d/aws-fsx.repo
+      if [[ -n "${replace_ver+x}" ]]; then
+        sudo sed -i "s#8#${replace_ver}#" /etc/yum.repos.d/aws-fsx.repo
+      fi
```
horsmand: Can you add a comment on what is going on here?

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,264 @@
...
+    *) echo "ERROR: Unsupported CentOS/RHEL version: $rhel_version"; exit 1;;
+  esac
+}
+
+function install_on_ubuntu() {
+  ubuntu_version=$(grep ^VERSION_ID= /etc/os-release | awk -F "=" '{print $2}' | sed s/\"//g)
+  kernel_version=$(uname -r)
+  architecture=$(uname -m)
+
+  case $ubuntu_version in
```
horsmand, line 202: Can we only support LTS versions? What about versions like 20.10 or 21.04?

horsmand (changes requested):
Looks good for the most part, I just found the bash a little hard to follow at some points. Additional comments would help.

packages/aws-rfdk/lib/core/lib/mountable-fsx-lustre.ts
jericht: Nope, no reason for that. Changed it to be consistent with the rest of the imports.

packages/aws-rfdk/lib/core/lib/mountable-fsx-lustre.ts
jericht: Thanks, fixed

packages/aws-rfdk/lib/core/lib/mountable-fsx-lustre.ts
jericht: Updated

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
jericht: Sure, added a brief comment with a link to the related AWS docs

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
jericht: The AWS docs don't have an explanation for it, but doing a bit of searching and it looks like the `kmod-lustre-client` package contains kernel-specific modules for the Lustre client. Added a comment at the top of the function with this info, since we're installing `kmod` packages in multiple places here.

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
jericht: 8.1 was not meant to be captured here, thanks for catching that! Fixed

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
jericht: Ah yeah I missed that, added in an entry for 8.3. I went ahead and added an entry for 8.4, but we technically don't need it yet since it's the latest release, so we don't have to download a specific version yet.

Unfortunately, yes. We will need to manually update this whenever a new version of RHEL is released, assuming the FSx Lustre docs also update the [instructions](https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html#lustre-client-rhel) with the new kernel versions added.

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
jericht: Sure, added a brief comment explaining what it does. For more context, the contents of `aws-fsx.repo` will look something like this:
```
[aws-fsx]
name=AWS FSx Packages - $basearch
baseurl=https://fsx-lustre-client-repo.s3.amazonaws.com/el/8/$basearch/
enabled=1
gpgcheck=1

[aws-fsx-src]
name=AWS FSx Source
baseurl=https://fsx-lustre-client-repo.s3.amazonaws.com/el/8/SRPMS/
enabled=1
gpgcheck=1
```

This changes the `baseurl` to point to a URL that has the right versions of the lustre client packages for the kernel we're on

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
jericht: I've only added support for what's documented [on AWS](https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html#lustre-client-ubuntu). I'm assuming that only the versions that are documented are the versions officially supported by AWS (may also be influenced by this: [Lustre Support Matrix](https://wiki.whamcloud.com/display/PUB/Lustre+Support+Matrix)), but I'll double check that with the FSx Lustre team.

jericht: Checked with FSx Lustre team and they confirmed that the non-LTS Ubuntu releases aren't supported.

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,271 @@
...
+    [7.7]='3.10.0-1062.*'
+    [7.8]='3.10.0-1127.*'
+    [7.9]='3.10.0-1160.*'
+    [8.1]='4.18.0-147.*'
+    [8.2]='4.18.0-193.*'
+    [8.3]='4.18.0-240.*'
+    [8.4]='4.18.0-305.*'
+  )
+
+  rhel_version=$(grep ^VERSION_ID= /etc/os-release | awk -F "=" '{print $2}' | sed s/\"//g)
```
horsmand: What happens when this is run on centos? I see we have a special block for that use case that will replace the value here.  Would it make sense to move this to an else statement in that block?

packages/aws-rfdk/lib/core/lib/mountable-fsx-lustre.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * The fileset to mount.
+   * @see https://docs.aws.amazon.com/fsx/latest/LustreGuide/mounting-from-fileset.html
+   * @default Mounts the root of the filesystem.
+   */
+  readonly fileset?: string;
+
+  /**
+   * Extra NFSv4 mount options that will be added to /etc/fstab for the file system.
```
ddneilson: Is this a typo? Should it be Lustre mount options with a link to a different man page, or does the Lustre mount helper support all NFSv4 mount options?

packages/aws-rfdk/lib/core/scripts/bash/mountFsxLustre.sh
```diff
@@ -0,0 +1,75 @@
...
+# See https://docs.aws.amazon.com/fsx/latest/LustreGuide/install-lustre-client.html
+#
+# Note: This script uses get_metadata_token and get_region from ./metadataUtilities.sh
+#  Thus, the system must have applications pre-installed as outlined in that file.
+#
+# Script arguments:
+#  $1 -- file system Identifier (ex: fs-00000000000)
+#  $2 -- Mount path; directory that we mount the file system to.
+#  $3 -- Mount name
+#  $4 -- (optional) NFSv4 mount options for the file system.
```
ddneilson: These should be lustre mount target options rather than NFSv4, I think

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,271 @@
...
+      curl https://fsx-lustre-client-repo-public-keys.s3.amazonaws.com/fsx-rpm-public-key.asc -o /tmp/fsx-rpm-public-key.asc
+      sudo rpm --import /tmp/fsx-rpm-public-key.asc
+      sudo curl https://fsx-lustre-client-repo.s3.amazonaws.com/el/8/fsx-lustre-client.repo -o /etc/yum.repos.d/aws-fsx.repo
+      if [[ -n "${replace_ver+x}" ]]; then
+        # Change the baseurl referenced in aws-fsx.repo to so we pull the correct version of the lustre client for this kernel
+        sudo sed -i "s#8#${replace_ver}#" /etc/yum.repos.d/aws-fsx.repo
+      fi
+
+      sudo yum clean all
+      sudo yum install -y kmod-lustre-client lustre-client
```
ddneilson, line 189: I guess the `yum install`s run `modprobe` to load the kernel module?

ddneilson (changes requested):

packages/aws-rfdk/lib/core/lib/mountable-fsx-lustre.ts
jericht: Yeah this is a copy-paste I missed, thanks for catching it. Fixed

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
jericht: Yes, good catch, fixed

packages/aws-rfdk/lib/core/scripts/bash/mountFsxLustre.sh
jericht: Yeah another copy-paste, thanks for catching it! Updated

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
jericht: According to the [Lustre wiki](https://wiki.lustre.org/Installing_the_Lustre_Software), it looks like it's not necessary to load the kernel modules via `modprobe` before using the Lustre client since they have instructions to unload them after loading them in. Seems like loading them in is just a way to verify they were installed correctly.

I've pretty much scripted the instructions on the FSx Lustre docs word by word, but it probably doesn't hurt it add the check with `modprobe` recommended by the Lustre wiki. Should we add these steps in?

ddneilson: It's probably not *necessary* if the wiki is just putting in the modprobe as a verification step. But, it certainly doesn't hurt if you think it would be relatively quick to throw that together -- the sanity check is kind of nice.

jericht: Sounds good, I've added these checks in and verified they work

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,288 @@
...
+    echo "ERROR: Kernel version ($kernel_version) is lower than the minimum required version: $min_version"
+    exit 1
+  fi
+
+  sudo yum install -y lustre-client
+}
+
+function verify_lustre_kmod() {
+  modprobe -v --first-time lustre
+  if [[ $? -ne 0 ]]; then
```
ddneilson: The shell script uses `set -e`, so if the `modprobe` doesn't return 0 then we'll bail immediately and never get to this condition test, no?

i.e. Should these two lines instead be?
```
if modprobe -v --first-time lustre; then
```

packages/aws-rfdk/lib/core/scripts/bash/installLustreClient.sh
```diff
@@ -0,0 +1,288 @@
...
+  if is_kernel_version_lower "$kernel_version" "$min_version"; then
+    echo "ERROR: Kernel version ($kernel_version) is lower than the minimum required version: $min_version"
+    exit 1
+  fi
+
+  sudo yum install -y lustre-client
+}
+
+function verify_lustre_kmod() {
+  modprobe -v --first-time lustre
```
ddneilson: Technically, should be `sudo`, no?

ddneilson (approved):
LGTM

horsmand (approved):

PR #465, opened 2021-06-11
fix(core): Convert group & pool names in SpotEventPluginFleet to lowercase

### LINKED ISSUE #425: SpotEventPluginFleet must lowercase group & pool names
Groups & pools are always lowercase in Deadline. The Deadline dialogs automatically change Groups/Pools to lowercase, and we also have code in RFDK that changes given groups/pools to lowercase when configuring workers (see: https://github.com/aws/aws-rfdk/blob/f46d9fca8a82a700103b476696215165515ef410/packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts#L323 ). 

The SpotEventPluginFleet is passing the Group/Pool names through as given, without ensuring that they are in lower case. With the SpotEventPluginFleet currently not creating the required groups, this means that it is impossible for a customer to create the required group on their own if they ask for one that has uppercase characters in it.

### Reproduction Steps

Create a SpotEventPluginFleet with a Group/Pool name that contains uppercase characters. The Spot Event Plugin configuration in Deadline will show that it is configured for the Group/Pool that has the uppercase characters -- but that Group/Pool cannot exist.

### Environment

  - **CDK CLI Version      :** Any
  - **CDK Framework Version:** Any
  - **RFDK Version:** latest
  - **Deadline Version:** any
  - **Node.js Version:** any
  - **OS               :** any
  - **Language (Version):** any

--- 

This is :bug: Bug Report

kozlove-aws:
### Problem
Groups & pools are always lowercase in Deadline.
The SpotEventPluginFleet is passing the Group/Pool names through as given, without ensuring that they are in lower case. With the SpotEventPluginFleet currently not creating the required groups, this means that it is impossible for a customer to create the required group on their own if they ask for one that has uppercase characters in it.

### Solution
 - Convert group and pool names lo lowercase
 - Changed validation for group names

### Testing

 - Deployed farm with group and pool that names contain capital symbols.
 - Verified that spot configuration deployed with converted group name.
 - Created pool and group with lowercase names and send a job with this pool and group.
 - Make sure that worker was spawned and rendered this job.
----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
@@ -412,6 +412,7 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
   constructor(scope: Construct, id: string, props: SpotEventPluginFleetProps) {
     super(scope, id);
 
+    this.deadlineGroups = props.deadlineGroups.map(group => group.toLocaleLowerCase());
     this.validateProps(props);
 
     this.securityGroups = props.securityGroups ?? [ new SecurityGroup(this, 'SpotFleetSecurityGroup', { vpc: props.vpc }) ];
@@ -447,7 +448,6 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
     this.maxCapacity = props.maxCapacity;
     this.validUntil = props.validUntil;
     this.keyName = props.keyName;
-    this.deadlineGroups = props.deadlineGroups;
 
     const imageConfig = props.workerMachineImage.getImage(this);
     this.osType = imageConfig.osType;
@@ -462,8 +462,8 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
       },
       renderQueue: props.renderQueue,
       workerSettings: {
-        groups: props.deadlineGroups,
-        pools: props.deadlinePools,
+        groups: this.deadlineGroups,
+        pools: props.deadlinePools?.map(pool => pool.toLocaleLowerCase()),
         region: props.deadlineRegion,
       },
       userDataProvider: props.userDataProvider,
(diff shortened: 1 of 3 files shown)


ddneilson (dismissed):

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
```diff
@@ -463,7 +463,7 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
       renderQueue: props.renderQueue,
       workerSettings: {
         groups: props.deadlineGroups,
-        pools: props.deadlinePools,
+        pools: props.deadlinePools?.map(pool => pool.toLocaleLowerCase()),
```
jericht, line 466: nit: This is technically not necessary since we are already converting to lowercase in [WorkerInstanceConfiguration](https://github.com/aws/aws-rfdk/blob/18a8098944f3599446c0f319eaf46b383724cfc3/packages/aws-rfdk/lib/deadline/lib/worker-configuration.ts#L324), so we could remove it.

Alternatively, we can keep this here, but we should use `this.deadlineGroups` in the above line so that we're passing lowercase versions for both pools and groups, and not half relying on `WorkerInstanceConfiguration` to lowercase them.

jericht (changes requested):
Just one tiny comment otherwise this looks great!

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
kozlove-aws: We cannot relay on  WorkerInstanceConfiguration since it is running only when worker instance spawned.
There is another [PR](https://github.com/aws/aws-rfdk/pull/465) for creating Pools and groups so we have to have groups and pools names lowercased. 
Group names are converted to lowercase above but it is good catch that I used values from `props`. (Should be fixed)

jericht (approved):

ddneilson (approved):

PR #466, opened 2021-06-11
fix(lambdas): handle unknown region when retrieving lambda layer arns

jericht:
### Problem

When synthesizing a stack template that would deploy a resource with one of our lambda functions that uses the `openssl-al2` lambda layer, if the AWS credentials/region are not provided, our code fails to resolve the lambda layer version ARN it should use for the lambda function. This is because the code tries to [index into a plain Type/Javascript object with the value retrieved from `Stack.of(this).region`](https://github.com/aws/aws-rfdk/blob/5a9989c5792306ed1b50a011f53fa24c3da4e16f/packages/aws-rfdk/lib/core/lib/x509-certificate.ts#L184-L188), which is a CDK token in this case, so the indexing returns `undefined`. This results in the lambda function getting synthesized with no lambda layers associated.

### Solution
Added a class that creates a `CfnMapping` for the lambda layer version ARNs and changed our lambda functions that use the `openssl-al2` lambda layer to reference this mapping.

### Testing
- Synthesized a stack without providing any AWS credentials/region and verified the lambda functions referencing the `CfnMapping` (see below)
- Deployed a basic stack with a `Repository` and `RenderQueue` and verified the relevant lambda functions were using the correct layers
- Built & ran `lambda-layers/bin/write-ts.js` to verify the file it generates is valid Typescript.

**Synthesized Stack Template w/o creds/region**
```yaml
  RenderQueueTlsCaCertPemGenerator37207ABB:
    Type: AWS::Lambda::Function
    Properties:
      # other props...
      Layers:
        - Fn::FindInMap:
            - LambdaLayerVersionArnMapping7a9adb59d5d042fbbc5ba55bcbf6a6b2
            - Ref: AWS::Region
            - OpenSslAl2

# Further down in the template...
Mappings:
  LambdaLayerVersionArnMapping7a9adb59d5d042fbbc5ba55bcbf6a6b2:
    ap-northeast-1:
      OpenSslAl2: arn:aws:lambda:ap-northeast-1:224375009292:layer:openssl-al2:2
    ap-northeast-2:
      OpenSslAl2: arn:aws:lambda:ap-northeast-2:224375009292:layer:openssl-al2:2
    ap-northeast-3:
      OpenSslAl2: arn:aws:lambda:ap-northeast-3:224375009292:layer:openssl-al2:1
    ap-south-1:
      OpenSslAl2: arn:aws:lambda:ap-south-1:224375009292:layer:openssl-al2:2
    ap-southeast-1:
      OpenSslAl2: arn:aws:lambda:ap-southeast-1:224375009292:layer:openssl-al2:2
    ap-southeast-2:
      OpenSslAl2: arn:aws:lambda:ap-southeast-2:224375009292:layer:openssl-al2:2
    ca-central-1:
      OpenSslAl2: arn:aws:lambda:ca-central-1:224375009292:layer:openssl-al2:2
    eu-central-1:
      OpenSslAl2: arn:aws:lambda:eu-central-1:224375009292:layer:openssl-al2:2
    eu-north-1:
      OpenSslAl2: arn:aws:lambda:eu-north-1:224375009292:layer:openssl-al2:2
    eu-west-1:
      OpenSslAl2: arn:aws:lambda:eu-west-1:224375009292:layer:openssl-al2:2
    eu-west-2:
      OpenSslAl2: arn:aws:lambda:eu-west-2:224375009292:layer:openssl-al2:2
    eu-west-3:
      OpenSslAl2: arn:aws:lambda:eu-west-3:224375009292:layer:openssl-al2:2
    sa-east-1:
      OpenSslAl2: arn:aws:lambda:sa-east-1:224375009292:layer:openssl-al2:2
    us-east-1:
      OpenSslAl2: arn:aws:lambda:us-east-1:224375009292:layer:openssl-al2:2
    us-east-2:
      OpenSslAl2: arn:aws:lambda:us-east-2:224375009292:layer:openssl-al2:2
    us-west-1:
      OpenSslAl2: arn:aws:lambda:us-west-1:224375009292:layer:openssl-al2:2
    us-west-2:
      OpenSslAl2: arn:aws:lambda:us-west-2:224375009292:layer:openssl-al2:2
```

----


packages/aws-rfdk/lib/lambdas/lambdaLayerVersionArns.ts
```diff
@@ -3,5 +3,108 @@
...
+
+export enum LambdaLayer {
+  OPEN_SSL_AL2 = 'OpenSslAl2',
+}
+
+type RegionMap<T> = { [region: string]: T };
+type LambdaLayerVersionArnMap<T> = { [layer: string]: T };
+
+const arns: LambdaLayerVersionArnMap<RegionMap<string>>= {
+  [LambdaLayer.OPEN_SSL_AL2]: {
```
ddneilson: These ARNs are autogenerated during the release publish of the RFDK -- https://github.com/aws/aws-rfdk/blob/mainline/lambda-layers/bin/write-ts.ts

We can't hard-code them in the construct like this.

jericht: Ah, I didn't notice that. I'll get another revision up to address that

horsmand (dismissed):
Looks good!

packages/aws-rfdk/lib/lambdas/lambda-layer-version-arn-mapping.ts
```diff
@@ -0,0 +1,103 @@
...
+   * Gets the singleton instance (per-stack) of this class.
+   * @param scope The parent construct.
+   */
+  public static getSingletonInstance(scope: Construct) {
+    const uuid = '7a9adb59-d5d0-42fb-bc5b-a55bcbf6a6b2';
+    const uniqueId = 'LambdaLayerVersionArnMapping' + uuid.replace(/[-]/g, '');
+    const stack = Stack.of(scope);
+
+    return (stack.node.tryFindChild(uniqueId) as LambdaLayerVersionArnMapping) ?? new LambdaLayerVersionArnMapping(stack, uniqueId, ARNS);
+  }
```
jusiskin: Can this be made `private` or `protected`?

Seems like we only need a static `getLambdaLayerVersion` method that could implicitly create the singleton. Then the API simply becomes:

```ts
const lambdaLayerVersion = LambdaLayerVersionArnMapping.getLambdaLayerVersion(this, 'LambdaLayerVersion', LambdaLayer.OPEN_SSL_AL2);
```

and the caller does not have to worry about calling `LambdaLayerVersionArnMapping.getSingletonInstance`

jusiskin (changes requested):
Looks good to me. Just a minor API simplification suggestion.

packages/aws-rfdk/lib/lambdas/lambda-layer-version-arn-mapping.ts
jericht: That's much cleaner, thanks! updated

jusiskin: Looks good! thanks

jusiskin (dismissed):

packages/aws-rfdk/lib/lambdas/lambda-layer-version-arn-mapping.ts
```diff
@@ -33,11 +33,25 @@ type LambdaLayerVersionArnMap<T> = { [layer: string]: T };
...
+    const instance = this.getSingletonInstance(scope);
+    return LayerVersion.fromLayerVersionArn(scope, id, instance.findInMap(instance.stack.region, layer));
+  }
+
   /**
    * Gets the singleton instance (per-stack) of this class.
    * @param scope The parent construct.
    */
-  public static getSingletonInstance(scope: Construct) {
+  protected static getSingletonInstance(scope: Construct) {
```
horsmand: Shouldn't this be private?

jericht: Yeah, no need for it to be `protected` yet, changed

horsmand (dismissed):

PR #470, opened 2021-06-15
feat(deadline): create Deadline Groups and Pools on deploy for ConfigureSpotEventPlugin

### LINKED ISSUE #358: Create deadline group/pool implicitly while configuring SEP
After creating and configuring a spot fleet via RFDK constructs `SpotEventPluginFleet` & `ConfigureSpotEventPlugin`, there is a requirement for manually creating deadline groups and pools before the spot fleet request can be created.
Details [here](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.ConfigureSpotEventPlugin.html)

This can be done by either
1. Running a deadline command to create group/pool, or,
2. Creating group/pool using deadline monitor.

This manual stop of creating the deadline groups/pools can be automated while deployment.

### Use Case

With this feature, once a farm is deployed by the user, the only step required by an user to start using the spot fleet is to submit a job for this deadline group.

### Proposed Solution

In order to eliminate this manual step of creating group/pool, an additional call can be made to RCS to automatically create the deadline group for which the spot request is being configured.

---

kozlove-aws:
### Problem 
After creating and configuring a spot fleet via RFDK constructs `SpotEventPluginFleet` & `ConfigureSpotEventPlugin`, there is a requirement for manually creating deadline groups and pools before the spot fleet request can be created.

### Solution
 - Add requests to create Groups/Pools during SEP configuration

### Testing
 - Deployed fleet with two new groups and two new pools
 - Validated that groups and pools were created
 - Deployed updated configuration with two fleets where groups are different but with same one pool
 - Validated that new groups and pool created.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
@@ -459,6 +459,8 @@ export class ConfigureSpotEventPlugin extends Construct {
     };
     const spotFleetRequestConfigs = this.mergeSpotFleetRequestConfigs(props.spotFleets);
 
+    const deadlineGroups = Array.from(new Set(props.spotFleets?.map(fleet => fleet.deadlineGroups).reduce((p, c) => p.concat(c), [])));
+    const deadlinePools = Array.from(new Set(props.spotFleets?.map(fleet => fleet.deadlinePools).reduce((p, c) => p?.concat(c ?? []), [])));
     const properties: SEPConfiguratorResourceProps = {
       connection: {
         hostname: props.renderQueue.endpoint.hostname,
@@ -468,6 +470,8 @@ export class ConfigureSpotEventPlugin extends Construct {
       },
       spotFleetRequestConfigurations: spotFleetRequestConfigs,
       spotPluginConfigurations: pluginConfig,
+      deadlineGroups,
+      deadlinePools,
     };
 
     const resource = new CustomResource(this, 'Default', {
--- a/packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
@@ -317,8 +317,6 @@ export interface ConfigureSpotEventPluginProps {
  * Logs for all AWS Lambdas are automatically recorded in Amazon CloudWatch.
  *
  * This construct will configure the Spot Event Plugin, but the Spot Fleet Requests will not be created unless you:
- * - Create the Deadline Group associated with the Spot Fleet Request Configuration. See [Deadline Documentation](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/pools-and-groups.html).
- * - Create the Deadline Pools to which the fleet Workers are added. See [Deadline Documentation](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/pools-and-groups.html).
  * - Submit the job with the assigned Deadline Group and Deadline Pool. See [Deadline Documentation](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/job-submitting.html#submitting-jobs).
  *
  * Important: Disable 'Allow Workers to Perform House Cleaning If Pulse is not Running' in the 'Configure Repository Options'
--- a/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
@@ -388,6 +388,13 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
   public readonly deadlineGroups: string[];
 
   /**
+   * Deadline pools the workers need to be assigned to.
+   *
+   * @default - Workers are not assigned to any pool
+   */
+  public readonly deadlinePools?: string[];
+
+  /**
    * Name of SSH keypair to grant access to instances.
    *
    * @default - No SSH access will be possible.
@@ -413,6 +420,7 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
     super(scope, id);
 
     this.deadlineGroups = props.deadlineGroups.map(group => group.toLocaleLowerCase());
+    this.deadlinePools = props.deadlinePools?.map(pool => pool.toLocaleLowerCase());
     this.validateProps(props);
 
     this.securityGroups = props.securityGroups ?? [ new SecurityGroup(this, 'SpotFleetSecurityGroup', { vpc: props.vpc }) ];
@@ -463,7 +471,7 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
       renderQueue: props.renderQueue,
       workerSettings: {
         groups: this.deadlineGroups,
-        pools: props.deadlinePools?.map(pool => pool.toLocaleLowerCase()),
+        pools: this.deadlinePools,
         region: props.deadlineRegion,
       },
       userDataProvider: props.userDataProvider,
--- a/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
@@ -90,9 +90,6 @@ export interface SpotEventPluginFleetProps {
   /**
    * Deadline groups these workers need to be assigned to.
    *
-   * Note that you will have to create the groups manually using Deadline before submitting jobs.
-   * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/pools-and-groups.html
-   *
    * Also, note that the Spot Fleet configuration does not allow using wildcards as part of the Group name
    * as described here https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/event-spot.html#wildcards
    */
@@ -100,9 +97,6 @@ export interface SpotEventPluginFleetProps {
 
   /**
    * Deadline pools these workers need to be assigned to.
-   *
-   * Note that you will have to create the pools manually using Deadline before submitting jobs.
-   * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/pools-and-groups.html
    *
    * @default - Workers are not assigned to any pool.
    */
--- a/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/handler.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/handler.ts
@@ -47,6 +47,14 @@ export class SEPConfiguratorResource extends SimpleCustomResource {
   public async doCreate(_physicalId: string, resourceProperties: SEPConfiguratorResourceProps): Promise<object|undefined> {
     const spotEventPluginClient = await this.spotEventPluginClient(resourceProperties.connection);
 
+    if (!await spotEventPluginClient.addGroups(resourceProperties.deadlineGroups)) {
+      throw new Error('Failed to add group collection.');
+    }
+
+    if (!await spotEventPluginClient.addPools(resourceProperties.deadlinePools)) {
+      throw new Error('Failed to add pool collection.');
+    }
+
     if (resourceProperties.spotFleetRequestConfigurations) {
       const convertedSpotFleetRequestConfigs = convertSpotFleetRequestConfiguration(resourceProperties.spotFleetRequestConfigurations);
       const stringConfigs = JSON.stringify(convertedSpotFleetRequestConfigs);
--- a/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/handler.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/handler.ts
@@ -48,11 +48,11 @@ export class SEPConfiguratorResource extends SimpleCustomResource {
     const spotEventPluginClient = await this.spotEventPluginClient(resourceProperties.connection);
 
     if (!await spotEventPluginClient.addGroups(resourceProperties.deadlineGroups)) {
-      throw new Error('Failed to add group collection.');
+      throw new Error(`Failed to add Deadline group(s) ${resourceProperties.deadlineGroups}`);
     }
 
     if (!await spotEventPluginClient.addPools(resourceProperties.deadlinePools)) {
-      throw new Error('Failed to add pool collection.');
+      throw new Error(`Failed to add Deadline pool(s) ${resourceProperties.deadlinePools}`);
     }
 
     if (resourceProperties.spotFleetRequestConfigurations) {
--- a/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
@@ -487,12 +487,12 @@ describe('SEPConfiguratorResource', () => {
       mockSpotEventPluginClient.addGroups = jest.fn( (a) => returnFalse(a) );
 
       // WHEN
-      const promise = handler.doCreate('physicalId', noConfigs);
+      const promise = handler.doCreate('physicalId', allConfigs);
 
       // THEN
       await expect(promise)
         .rejects
-        .toThrowError(/Failed to add group collection./);
+        .toThrowError(`Failed to add Deadline group(s) ${allConfigs.deadlineGroups}`);
     });
 
     test('throw when cannot add pools', async () => {
@@ -500,12 +500,12 @@ describe('SEPConfiguratorResource', () => {
       mockSpotEventPluginClient.addPools = jest.fn( (a) => returnFalse(a) );
 
       // WHEN
-      const promise = handler.doCreate('physicalId', noConfigs);
+      const promise = handler.doCreate('physicalId', allConfigs);
 
       // THEN
       await expect(promise)
         .rejects
-        .toThrowError(/Failed to add pool collection./);
+        .toThrowError(`Failed to add Deadline pool(s) ${allConfigs.deadlinePools}`);
     });
 
     test('throw when cannot save spot fleet request configs', async () => {
--- a/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
@@ -205,24 +205,44 @@ describe('SEPConfiguratorResource', () => {
     connection: validConnection,
   };
 
+  const deadlineGroups =  ['group_name'];
+  const deadlinePools =  ['pool_name'];
+
   const allConfigs: SEPConfiguratorResourceProps = {
     spotPluginConfigurations: validSpotEventPluginConfig,
     connection: validConnection,
     spotFleetRequestConfigurations: validSpotFleetRequestConfig,
+    deadlineGroups,
+    deadlinePools,
   };
 
   const noConfigs: SEPConfiguratorResourceProps = {
     connection: validConnection,
   };
 
+  async function returnTrue(_v1: any): Promise<boolean> {
+    return true;
+  }
+
+  async function returnFalse(_v1: any): Promise<boolean> {
+    return false;
+  }
+
   describe('doCreate', () => {
     let handler: SEPConfiguratorResource;
-    let mockSpotEventPluginClient: { saveServerData: jest.Mock<any, any>; configureSpotEventPlugin: jest.Mock<any, any>; };
+    let mockSpotEventPluginClient: {
+      saveServerData: jest.Mock<any, any>;
+      configureSpotEventPlugin: jest.Mock<any, any>;
+      addGroups: jest.Mock<any, any>;
+      addPools: jest.Mock<any, any>;
+    };
 
     beforeEach(() => {
       mockSpotEventPluginClient = {
-        saveServerData: jest.fn(),
-        configureSpotEventPlugin: jest.fn(),
+        saveServerData: jest.fn( (a) => returnTrue(a) ),
+        configureSpotEventPlugin: jest.fn( (a) => returnTrue(a) ),
+        addGroups: jest.fn( (a) => returnTrue(a) ),
+        addPools: jest.fn( (a) => returnTrue(a) ),
       };
 
       handler = new SEPConfiguratorResource(new AWS.SecretsManager());
@@ -242,9 +262,6 @@ describe('SEPConfiguratorResource', () => {
 
     test('with no configs', async () => {
       // GIVEN
-      async function returnTrue(_v1: any): Promise<boolean> {
-        return true;
-      }
       const mockSaveServerData = jest.fn( (a) => returnTrue(a) );
       mockSpotEventPluginClient.saveServerData = mockSaveServerData;
       const mockConfigureSpotEventPlugin = jest.fn( (a) => returnTrue(a) );
@@ -261,9 +278,6 @@ describe('SEPConfiguratorResource', () => {
 
     test('save spot fleet request configs', async () => {
       // GIVEN
-      async function returnTrue(_v1: any): Promise<boolean> {
-        return true;
-      }
       const mockSaveServerData = jest.fn( (a) => returnTrue(a) );
       mockSpotEventPluginClient.saveServerData = mockSaveServerData;
 
@@ -281,9 +295,6 @@ describe('SEPConfiguratorResource', () => {
 
     test('save spot fleet request configs without BlockDeviceMappings', async () => {
       // GIVEN
-      async function returnTrue(_v1: any): Promise<boolean> {
-        return true;
-      }
       const mockSaveServerData = jest.fn( (a) => returnTrue(a) );
       mockSpotEventPluginClient.saveServerData = mockSaveServerData;
 
@@ -326,9 +337,6 @@ describe('SEPConfiguratorResource', () => {
 
     test('save spot fleet request configs without Ebs', async () => {
       // GIVEN
-      async function returnTrue(_v1: any): Promise<boolean> {
-        return true;
-      }
       const mockSaveServerData = jest.fn( (a) => returnTrue(a) );
       mockSpotEventPluginClient.saveServerData = mockSaveServerData;
 
@@ -375,9 +383,6 @@ describe('SEPConfiguratorResource', () => {
 
     test('save spot event plugin configs', async () => {
       // GIVEN
-      async function returnTrue(_v1: any): Promise<boolean> {
-        return true;
-      }
       const mockConfigureSpotEventPlugin = jest.fn( (a) => returnTrue(a) );
       mockSpotEventPluginClient.configureSpotEventPlugin = mockConfigureSpotEventPlugin;
 
@@ -407,14 +412,22 @@ describe('SEPConfiguratorResource', () => {
       expect(mockConfigureSpotEventPlugin.mock.calls[0][0]).toEqual([...configs, ...securitySettings]);
     });
 
-    test('save both configs', async () => {
+    test('save server data', async () => {
       // GIVEN
-      async function returnTrue(_v1: any): Promise<boolean> {
-        return true;
-      }
       const mockSaveServerData = jest.fn( (a) => returnTrue(a) );
       mockSpotEventPluginClient.saveServerData = mockSaveServerData;
 
+      // WHEN
+      const result = await handler.doCreate('physicalId', allConfigs);
+
+      // THEN
+      expect(result).toBeUndefined();
+      expect(mockSaveServerData.mock.calls.length).toBe(1);
+      expect(mockSaveServerData.mock.calls[0][0]).toEqual(JSON.stringify(validConvertedSpotFleetRequestConfig));
+    });
+
+    test('configure spot event plugin', async () => {
+      // GIVEN
       const mockConfigureSpotEventPlugin = jest.fn( (a) => returnTrue(a) );
       mockSpotEventPluginClient.configureSpotEventPlugin = mockConfigureSpotEventPlugin;
 
@@ -436,22 +449,67 @@ describe('SEPConfiguratorResource', () => {
       }];
 
       // WHEN
-      const result = await handler.doCreate('physicalId', allConfigs);
+      await handler.doCreate('physicalId', allConfigs);
 
       // THEN
-      expect(result).toBeUndefined();
-      expect(mockSaveServerData.mock.calls.length).toBe(1);
-      expect(mockSaveServerData.mock.calls[0][0]).toEqual(JSON.stringify(validConvertedSpotFleetRequestConfig));
-
       expect(mockConfigureSpotEventPlugin.mock.calls.length).toBe(1);
       expect(mockConfigureSpotEventPlugin.mock.calls[0][0]).toEqual([...configs, ...securitySettings]);
     });
 
+    test('create groups', async () => {
+      // GIVEN
+      const mockAddGroups = jest.fn( (a) => returnTrue(a) );
+      mockSpotEventPluginClient.addGroups = mockAddGroups;
+
+      // WHEN
+      await handler.doCreate('physicalId', allConfigs);
+
+      // THEN
+      expect(mockAddGroups.mock.calls.length).toBe(1);
+      expect(mockAddGroups).toHaveBeenCalledWith(deadlineGroups);
+    });
+
+    test('create pools', async () => {
+      // GIVEN
+      const mockAddPools = jest.fn( (a) => returnTrue(a) );
+      mockSpotEventPluginClient.addPools = mockAddPools;
+
+      // WHEN
+      await handler.doCreate('physicalId', allConfigs);
+
+      // THEN
+      expect(mockAddPools.mock.calls.length).toBe(1);
+      expect(mockAddPools).toHaveBeenCalledWith(deadlinePools);
+    });
+
+    test('throw when cannot add groups', async () => {
+      // GIVEN
+      mockSpotEventPluginClient.addGroups = jest.fn( (a) => returnFalse(a) );
+
+      // WHEN
+      const promise = handler.doCreate('physicalId', noConfigs);
+
+      // THEN
+      await expect(promise)
+        .rejects
+        .toThrowError(/Failed to add group collection./);
+    });
+
+    test('throw when cannot add pools', async () => {
+      // GIVEN
+      mockSpotEventPluginClient.addPools = jest.fn( (a) => returnFalse(a) );
+
+      // WHEN
+      const promise = handler.doCreate('physicalId', noConfigs);
+
+      // THEN
+      await expect(promise)
+        .rejects
+        .toThrowError(/Failed to add pool collection./);
+    });
+
     test('throw when cannot save spot fleet request configs', async () => {
       // GIVEN
-      async function returnFalse(_v1: any): Promise<boolean> {
-        return false;
-      }
       const mockSaveServerData = jest.fn( (a) => returnFalse(a) );
       mockSpotEventPluginClient.saveServerData = mockSaveServerData;
 
@@ -466,9 +524,6 @@ describe('SEPConfiguratorResource', () => {
 
     test('throw when cannot save spot event plugin configs', async () => {
       // GIVEN
-      async function returnFalse(_v1: any): Promise<boolean> {
-        return false;
-      }
       const mockConfigureSpotEventPlugin = jest.fn( (a) => returnFalse(a) );
       mockSpotEventPluginClient.configureSpotEventPlugin = mockConfigureSpotEventPlugin;
 
--- a/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/types.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/types.ts
@@ -23,6 +23,16 @@ export interface SEPConfiguratorResourceProps {
    * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/event-spot.html#event-plugin-configuration-options
    */
   readonly spotPluginConfigurations?: PluginSettings;
+
+  /**
+   * Deadline groups that are used by these fleets.
+   */
+  readonly deadlineGroups?: string[];
+
+  /**
+   * Deadline pools that are used by these fleets.
+   */
+  readonly deadlinePools?: string[];
 }
 
 /**
--- a/packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
+++ b/packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
@@ -26,6 +26,30 @@ interface DescribeServerDataResponse {
 }
 
 /**
+ * A response from get pool/group request
+ */
+export interface PoolGroupCollections {
+  /**
+   * The collection of user-created Pools/Groups that are currently active
+   */
+  readonly Pools: string [];
+
+  /**
+   * The collection of Pools/Groups that are currently obsolete
+   */
+  readonly ObsoletePools: string [];
+}
+
+/**
+ * A type of collection to get/recive from Deadline.
+ */
+export enum CollectionType {
+  Pool = 'pool',
+
+  Group = 'group',
+}
+
+/**
  * Provides a simple interface to send requests to the Render Queue API related to the Deadline Spot Event Plugin.
  */
 export class SpotEventPluginClient {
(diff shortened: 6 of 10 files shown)


packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
```diff
@@ -459,6 +459,8 @@ export class ConfigureSpotEventPlugin extends Construct {
     };
     const spotFleetRequestConfigs = this.mergeSpotFleetRequestConfigs(props.spotFleets);
 
+    const deadlineGroups = props.spotFleets ? [...new Set(([] as string[]).concat(...props.spotFleets?.map(fleet => fleet.deadlineGroups)))] : [];
```
jericht: It looks like we want to get a set of all groups/pools we want to configure here. I think this could simplify to something like:
```ts
const deadlineGroups = Array.from(new Set(props.spotFleets?.flatMap(fleet => fleet.deadlineGroups)));
```

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
```diff
@@ -448,6 +455,7 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
     this.validUntil = props.validUntil;
     this.keyName = props.keyName;
     this.deadlineGroups = props.deadlineGroups;
+    this.deadlinePools = props.deadlinePools;
```
jericht: If https://github.com/aws/aws-rfdk/pull/465 is merged in before this, we should remember to lowercase the pools here.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/handler.ts
```diff
@@ -147,6 +150,23 @@ export class SEPConfiguratorResource extends SimpleCustomResource {
       },
     ];
   }
+
+  private async addGroupPoolCollection(spotEventPluginClient: SpotEventPluginClient, isGroup: boolean, instances?: string[]) {
```
jericht: nit: `instances` for the parameter name is a bit confusing here. Similar to my suggestions for refactoring the changes in `SpotEventPluginClient`, I think we should make separate methods for groups and pools here.

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
```diff
@@ -94,6 +109,49 @@ export class SpotEventPluginClient {
     }
   }
 
+  public async getGroupPoolCollection(isGroup: boolean): Promise<string[]> {
```
jericht: I like that you've used these methods as a way to avoid duplicating code, I have a couple suggestions though:

1. We should create separate methods for groups and pools. Right now, we only need groups and pools, but if we ever need to add a third API that goes to a `/db/settings/collections/<thing>` endpoint, we would still need to duplicate the code below.
2. We can actually move these methods out of this file, since Deadline groups and pools don't always have to be related SEP. I'm thinking we move this into some new `getGroups`/`saveGroups` and `getPools`/`savePools` methods either in the `DeadlineClient` class or in a new class that wraps `DeadlineClient`. 

How does that sound?

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
```diff
@@ -25,6 +25,21 @@ interface DescribeServerDataResponse {
   readonly ServerData: DescribedServerData[];
 }
 
+/**
+ * A response from get pool/group request
+ */
+export interface PoolCollections {
```
jericht: Along with my suggestion to move the group/pool methods to `DeadlineClient`, maybe we should put interfaces like these in a `types.ts` file in the `deadline-client` folder. This way all of our types that interface with Deadline APIs are in the same spot. What do you think?

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -418,6 +438,15 @@ describe('SEPConfiguratorResource', () => {
       const mockConfigureSpotEventPlugin = jest.fn( (a) => returnTrue(a) );
       mockSpotEventPluginClient.configureSpotEventPlugin = mockConfigureSpotEventPlugin;
 
+      async function returnEmptyArray(_v1: any): Promise<string[]> {
+        return [];
+      }
+      const mockGetGroupPoolCollection = jest.fn( (a) => returnEmptyArray(a) );
```
jericht: nit: Could we do this instead of defining an inner function?
```suggestion
      const mockGetGroupPoolCollection = jest.fn( async (_a) => [] );
```

I see that we're already using inner functions elsewhere in this file, so I'm okay with leaving this as is.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -480,6 +517,59 @@ describe('SEPConfiguratorResource', () => {
         .rejects
         .toThrowError(/Failed to save Spot Event Plugin Configurations/);
     });
+
+    test('throw when cannot save group collection', async () => {
```
jericht: I think these tests can be improved by using something like `sinon` to stub out or mock methods for us in a nicer way. For example, we could rewrite the mocking portion of this test to something like:
```ts
    test('throw when cannot save group collection', async () => {
      // GIVEN
      const stubGetGroupPoolCollection = sinon.stub(SpotEventPluginClient.prototype, 'getGroupPoolCollection').resolves([]);
      const stubSaveGroupPoolCollection = sinon.stub(SpotEventPluginClient.prototype, 'saveGroupPoolCollection').resolves(false);
      const stubConfigureSpotEventPlugin = sinon.stub(SpotEventPluginClient.prototype, 'configureSpotEventPlugin').resolves(false);

      // WHEN
      // ...the rest of the test...
    });
```

With the way it's written right now, we won't be able to do this without changing how all the tests work, so I'd say we can leave this as is. Just wanted to make note of this so that when we need to add more to these tests, we can do a bit of cleanup then.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -205,10 +205,22 @@ describe('SEPConfiguratorResource', () => {
...
+
   const allConfigs: SEPConfiguratorResourceProps = {
     spotPluginConfigurations: validSpotEventPluginConfig,
     connection: validConnection,
     spotFleetRequestConfigurations: validSpotFleetRequestConfig,
+    deadlineGroups,
+    deadlinePools,
+  };
+
+  const noGroupsButPools: SEPConfiguratorResourceProps = {
```
jericht: Instead of creating this variable up here, in the test we need to remove groups, we can do something like this instead:
```ts
const configNoGroups: SEPConfiguratorResourceProps = {
  ...allConfigs,
  deadlinePools: [],
}
```
This way we're keeping this test case input inside the test it's used in.

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -8,10 +8,13 @@ import { Socket } from 'net';
 import { DeadlineClient, Response } from '../../deadline-client';
 import { SpotEventPluginClient } from '../spot-event-plugin-client';
 
+const deadlinePools =  ['pool_name'];
```
jericht: nit: Can we define this inside the `describe` block where it's being used?

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +250,111 @@ describe('SpotEventPluginClient', () => {
...
+
+  test('successful getGroupPoolCollection for groups', async () => {
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    spotEventPluginClient['deadlineClient'].GetRequest = jest.fn().mockResolvedValue(successfulPoolResponse);
+    const result = await spotEventPluginClient.getGroupPoolCollection(true);
+
+    // THEN
+    expect(result).toEqual(deadlinePools);
+    // expect(result).toBeNaN();
```
jericht:
```suggestion

```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +250,111 @@ describe('SpotEventPluginClient', () => {
...
+
+  test('successful getGroupPoolCollection for pools', async () => {
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    spotEventPluginClient['deadlineClient'].GetRequest = jest.fn().mockResolvedValue(successfulPoolResponse);
+    const result = await spotEventPluginClient.getGroupPoolCollection(false);
+
+    // THEN
+    expect(result).toEqual(deadlinePools);
+    // expect(result).toBeNaN();
```
jericht:
```suggestion

```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +250,111 @@ describe('SpotEventPluginClient', () => {
...
   });
+
+  test('successful getGroupPoolCollection for groups', async () => {
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    spotEventPluginClient['deadlineClient'].GetRequest = jest.fn().mockResolvedValue(successfulPoolResponse);
+    const result = await spotEventPluginClient.getGroupPoolCollection(true);
+
+    // THEN
+    expect(result).toEqual(deadlinePools);
```
jericht: nit: variable name is misleading. While they are both just string arrays, I think it's worth creating a separate `deadlineGroups` variable so that our tests are clear about what they are testing.

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +250,111 @@ describe('SpotEventPluginClient', () => {
     // THEN
     await expect(promise).rejects.toEqual(statusMessage);
   });
+
+  test('successful getGroupPoolCollection for groups', async () => {
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    spotEventPluginClient['deadlineClient'].GetRequest = jest.fn().mockResolvedValue(successfulPoolResponse);
```
jericht: Similar to my comment as below about `deadlineGroups`, I think it's worth creating a corresponding `successfulGroupResponse` to use here.

jericht (changes requested):
Overall nice work! Just added some suggestions to reorganize the code a bit.

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
kozlove-aws: Pools and groups are very similar in Deadline, they even use the same object type `PoolCollections`.
I can't see any reasons to duplicate code for these two instances.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
kozlove-aws: `allConfigs` contains group, so your example just create config with empty Pools but with Groups. This is not what I need for test.

jericht: Oh yeah, woops, I got mixed up, what I meant to suggest was:
```ts
const configNoGroups: SEPConfiguratorResourceProps = {
  ...allConfigs,
  deadlineGroups: [],
}
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
jericht: I think we can take this opportunity to establish a pattern in our code that wraps the `DeadlineClient.GetRequest/PostRequest` methods. While groups and pools may be (coincidentally) returning the same data structure, from an API perspective they are different resources, so I think we should reflect that in our interfaces.

To me, this code expresses its intent a bit better
```ts
const pools = await client.getPools();
pools.push('my_new_pool');
await client.postPools(pools);
```
than
```ts
const pools = await client.getGroupPoolCollection(false);
pools.push('my_new_pool');

// accidentally mixed up the boolean here, so this will overwrite the users groups with these pools
await client.saveGroupPoolCollection(pools, true);
```

I think the benefits of having a more intuitive interface outweigh having to split these methods up with some  code duplication. What do you think?

jericht (changes requested):

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -480,6 +513,57 @@ describe('SEPConfiguratorResource', () => {
...
+      // WHEN
+      const promise = handler.doCreate('physicalId', allConfigs);
+
+      // THEN
+      await expect(promise)
+        .rejects
+        .toThrowError(/Failed to save group collection/);
+    });
+
+    test('throw when cannot save pool collection', async () => {
```
jericht: We can remove this test now since we can assert the behavior of `getCollection` using the input `CollectionType` we provide (same goes for other tests that are similar but only different due to being groups/pools)

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -480,6 +513,57 @@ describe('SEPConfiguratorResource', () => {
         .rejects
         .toThrowError(/Failed to save Spot Event Plugin Configurations/);
     });
+
+    test('throw when cannot save group collection', async () => {
```
jericht:
```suggestion
    test('throw when cannot save collection', async () => {
```

I think we can generalize this test and just use a variable `const collectionType = CollectionType.Group` as our test case. Then we can assert the error message contains the `collectionType` we passed in.

Alternatively, we can parameterize this test for all the values in `CollectionType` using something like `test.each`:

```ts
test.each(
  Object.values(CollectionType)
)('throws when collection fails to save (%s)', async (ct: CollectionType) => {
  // Test code here
});
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
```diff
@@ -94,6 +118,47 @@ export class SpotEventPluginClient {
...
+  }
+
+  public async saveCollection(pools: string[], type: CollectionType): Promise<boolean> {
+    console.log(`Saving ${type.toString()} collection:`);
+    console.log(pools);
+
+    try {
+      await this.deadlineClient.PostRequest(`/db/settings/collections/${type.toString()}s/save`, {
+        Pools: pools,
+        ObsoletePools: [],
```
jericht: Just to make sure, would this overwrite what the old values of `ObsoletePools` are in Deadline with an empty array? or does it retain the old value for `ObsoletePools`?

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
```diff
@@ -94,6 +118,47 @@ export class SpotEventPluginClient {
     }
   }
 
+  public async getCollection(type: CollectionType): Promise<string[]> {
```
jericht: A couple of observations/suggestions, but they're not necessary changes:

Here, we're assuming Deadline collections are the same type (in this case `string[]`). This may not always be the case, but it should probably be fine for now seeing as there are only `/db/settings/collections/<type>`  endpoints for groups/pools.

Another thing I want to point out is that we might want to consider returning `Promise<PoolCollections>` from this method rather than just returning the value of the `.Pools` field. If callers of this method only happen to need the `.Pools` field, they can extract it themselves, but since this class is specifically for SEP, it's probably fine that we make the assumption that the `.ObsoletePools` field is not needed.

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +260,109 @@ describe('SpotEventPluginClient', () => {
...
+
+    // THEN
+    expect(result).toEqual(deadlineGroups);
+    // eslint-disable-next-line dot-notation
+    expect(spotEventPluginClient['deadlineClient'].GetRequest).toBeCalledTimes(1);
+    expect(consoleLogMock.mock.calls.length).toBe(1);
+    expect(consoleLogMock.mock.calls[0][0]).toMatch(/Getting group collection:/);
+  });
+
+  test('successful getCollection for pools', async () => {
```
jericht: Similar to my comments in `packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts`, we can unify tests that only differ based on whether they use groups or pools and possibly parameterize the tests based on `CollectionType` values.

jericht (changes requested):

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
kozlove-aws: These tests are for `doCreate` for make sure it throw correct errors.

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
kozlove-aws: Good catch. I have check and `ObsoletePools` is rewritten with empty values. Fixed.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
jericht: I still think we can do away with this variable and just override `allConfigs` as necessary in each test. It might get pretty messy up here as we add more test cases to this file. But it's more of a nitpick, so resolving.

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
jericht: This is just a nit, so not worth holding up the PR for it. Resolving.

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
```diff
@@ -94,6 +118,45 @@ export class SpotEventPluginClient {
...
+        ObsoletePools: pools.ObsoletePools,
+      },
+      {
+        headers: {
+          'Content-Type': 'application/json; charset=utf-8',
+        },
+      });
+      return true;
+    } catch(e) {
+      console.error(`Failed to save ${type} collection. Reason: ${e}`);
```
jericht: The [code above](https://github.com/aws/aws-rfdk/pull/470/files#diff-153ff117c810c27af1307dcb1f5c3c171197e7d9a54c92fb5ca606c1c0ca9014R135) casts the type of `e` to `Error` and uses the `.message` field. Should we have the same thing here?

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,94 @@ describe('SpotEventPluginClient', () => {
...
+
+    // THEN
+    await expect(promise).rejects.toThrowError(`Failed to get group collection. Reason: Failed to receive a group collection. Invalid response: ${JSON.stringify(invalidGroupResponse.data)}.`);
+  });
+
+  test('successful saveCollection for groups', async () => {
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    spotEventPluginClient['deadlineClient'].PostRequest = jest.fn().mockResolvedValue({});;
+    const result = await spotEventPluginClient.saveCollection( groupsColection, CollectionType.Group);
```
jericht:
```suggestion
    const result = await spotEventPluginClient.saveCollection(groupsColection, CollectionType.Group);
```

jericht (changes requested):
Just a couple of tiny nitpicks, otherwise looks good

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,94 @@ describe('SpotEventPluginClient', () => {
...
+    // THEN
+    expect(result).toBeTruthy();
+    // eslint-disable-next-line dot-notation
+    expect(spotEventPluginClient['deadlineClient'].PostRequest).toBeCalledTimes(1);
+    expect(consoleLogMock.mock.calls.length).toBe(2);
+    expect(consoleLogMock.mock.calls[0][0]).toMatch(/Saving group collection:/);
+    expect(consoleLogMock.mock.calls[1][0]).toBe(groupsColection);
+  });
+
+  test('successful saveCollection for pools', async () => {
```
jericht: We can do the same thing we did for `Successful getCollection for %s` here, using `test.each` to combine this test and the groups version.

jericht (dismissed):

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/handler.ts
```diff
@@ -47,6 +51,37 @@ export class SEPConfiguratorResource extends SimpleCustomResource {
...
+      const deadlineGroups = await spotEventPluginClient.getCollection(CollectionType.Group);
+      const newDeadlineGroups = deadlineGroups.Pools
+        .concat(resourceProperties.deadlineGroups
+          .filter(group => !deadlineGroups.Pools.includes(group)));
+      const response = await spotEventPluginClient
+        .saveCollection({
+          Pools: newDeadlineGroups,
+          ObsoletePools: deadlineGroups.ObsoletePools,
+        } as PoolCollections,
+        CollectionType.Group);
```
jusiskin: I echo @jericht 's [feedback around the API](https://github.com/aws/aws-rfdk/pull/470#discussion_r654660983) we are introducing here. It does not seem intuitive for user of the `SpotEventPluginClient` unless you know about Deadline internals. We should strive for an API that makes sense for administrators/users of Deadline.

Currently, users must call `.getCollection(...)` and `.saveCollection(...)` which may not be very intuitive/discoverable for those looking to work with pools and groups or anyone reading the code. The use of the tern "pool" in this calling code is a sign that we probably need to refine the API. The implementation behind the API can make use of common code for dealing with "collections", but let's abstract that from the API/users. Similarly, since we are ignoring `ObsoletePools` (it is seemingly obsolete), let's cut it from the API.

Working backwards, can we change the API to look something like:

```ts
const deadlineGroups = await spotEventPluginClient.getGroups();
const newDeadlineGroups = Array.from(new Set(deadlineGroups.concat(resourceProperties.deadlineGroups)));
const response = await spotEventPluginClient.saveGroups(newDeadlineGroups);
```

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -445,6 +478,14 @@ describe('SEPConfiguratorResource', () => {
...
       expect(mockConfigureSpotEventPlugin.mock.calls.length).toBe(1);
       expect(mockConfigureSpotEventPlugin.mock.calls[0][0]).toEqual([...configs, ...securitySettings]);
+
+      expect(mockGetCollection.mock.calls.length).toBe(2);
+      expect(mockGetCollection.mock.calls[0][0]).toBe(CollectionType.Group);
+      expect(mockGetCollection.mock.calls[1][0]).toBe(CollectionType.Pool);
+
+      expect(mockSaveCollection.mock.calls.length).toBe(2);
+      expect(mockSaveCollection.mock.calls[0][0]).toEqual({ Pools: deadlineGroups, ObsoletePools: []});
+      expect(mockSaveCollection.mock.calls[1][0]).toEqual({ Pools: deadlinePools, ObsoletePools: []});
```
jusiskin: Here is where we are asserting that the handler is fetching and saving the groups and pools. Since this is an individual behavior that can potentially regress independent of other behaviors, can we separate it into it's own `test()` section? That way we'd have more fine-grained failures that are easier to reason about.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/types.ts
```diff
@@ -23,6 +23,16 @@ export interface SEPConfiguratorResourceProps {
    * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/event-spot.html#event-plugin-configuration-options
    */
   readonly spotPluginConfigurations?: PluginSettings;
+
+  /**
+   * Deadline groups are used by these fleets.
```
jusiskin: nit: missed word

```suggestion
   * Deadline groups that are used by these fleets.
```

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/types.ts
```diff
@@ -23,6 +23,16 @@ export interface SEPConfiguratorResourceProps {
...
+   */
+  readonly deadlineGroups?: string[];
+
+  /**
+   * Deadline pools are used by these fleets.
```
jusiskin: nit: missed word

```suggestion
   * Deadline pools that are used by these fleets.
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/spot-event-plugin-client.ts
```diff
@@ -25,6 +25,30 @@ interface DescribeServerDataResponse {
   readonly ServerData: DescribedServerData[];
 }
 
+/**
+ * A response from get pool/group request
+ */
+export interface PoolCollections {
```
jusiskin: nit: since this is generic to both groups and pools, can we generalize the interface name? How about `WorkerCollections`?

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -480,6 +521,31 @@ describe('SEPConfiguratorResource', () => {
         .rejects
         .toThrowError(/Failed to save Spot Event Plugin Configurations/);
     });
+
+    test.each([
+      ['group', allConfigs],
+      ['pool', noGroupsButPools],
+    ])('throws when collection fails to save %s', async (collectionType: string, config: any) => {
```
jusiskin: should we add a test for failing to get a collection?

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,82 @@ describe('SpotEventPluginClient', () => {
...
     await expect(promise).rejects.toEqual(statusMessage);
   });
+
+  test.each([
+    [CollectionType.Group, successfulGroupResponse, groupsColection],
+    [CollectionType.Pool, successfulPoolResponse, poolsColection],
+  ])('Successful getCollection for %s', async (type: CollectionType, response: Response, expectedResult: any) => {
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    spotEventPluginClient['deadlineClient'].GetRequest = jest.fn().mockResolvedValue(response);
```
jusiskin, line 266: nit: this should be in a `// GIVEN` section

jusiskin (changes requested):
Great work. I think it might help to make the new `SpotEventPluginClient` APIs a little more intuitive.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/handler.ts
jusiskin: The new API is more intuitive and encapsulates the preservation of `ObsoletePools`. Much better, thanks! Resolved

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -445,6 +465,10 @@ describe('SEPConfiguratorResource', () => {
 
       expect(mockConfigureSpotEventPlugin.mock.calls.length).toBe(1);
       expect(mockConfigureSpotEventPlugin.mock.calls[0][0]).toEqual([...configs, ...securitySettings]);
+
+      expect(mockAddGroups.mock.calls.length).toBe(1);
+
+      expect(mockAddPools.mock.calls.length).toBe(1);
```
jusiskin: we should probably assert what arguments are being passed to these functions. How about:

```ts
expect(mockAddGroups).toHaveBeenCalledWith(deadlineGroups);
expect(mockAddPools).toHaveBeenCalledWith(deadlinePools);
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,163 @@ describe('SpotEventPluginClient', () => {
...
+    // WHEN
+    await spotEventPluginClient.addGroups(groupsCollection);
+
+    // THEN
+    // eslint-disable-next-line dot-notation
+    expect(spotEventPluginClient['deadlineClient'].GetRequest).toBeCalledTimes(reuquestsCount);
+
+    // eslint-disable-next-line dot-notation
+    expect(spotEventPluginClient['deadlineClient'].PostRequest).toBeCalledTimes(reuquestsCount);
+  });
```
jusiskin: The only thing we are asserting in this test is that `deadlineClient.GetRequest` and `deadlineClent.PostRequest` are being called once. We should add more assertions here to confirm that the correct behavior is being applied. For this, we probably want to split this into a few tests:

*  A test to ensure groups are added if they don't already exist
*  A test to ensure existing groups are not duplicated
*  A test to ensure a mix of (non-)existing groups are correctly merged

Same applies for the `.addPools(...)` test below.

jusiskin (changes requested):
The API looks much better now. The only thing missing is more test coverage to ensure the new APIs behave as expected.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/handler.ts
```diff
@@ -47,6 +47,14 @@ export class SEPConfiguratorResource extends SimpleCustomResource {
   public async doCreate(_physicalId: string, resourceProperties: SEPConfiguratorResourceProps): Promise<object|undefined> {
     const spotEventPluginClient = await this.spotEventPluginClient(resourceProperties.connection);
 
+    if (!await spotEventPluginClient.addGroups(resourceProperties.deadlineGroups)) {
+      throw new Error('Failed to add group collection.');
```
jusiskin: nit: I'm thinking this might benefit from a slightly clearer error message so that anyone encountering it understands this is a Deadline interaction. I'm thinking:

> Failed to add Deadline group(s)

what do you think? It might even help to include the group names in the message.

Same feedback applies to the pools message below

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,139 @@ describe('SpotEventPluginClient', () => {
...
+
+  test.each([
+    [poolsColection.Pools,1],
+    [[], 0],
+  ])('successful call addPool with %s', async (poolsCollection: string[], reuquestsCount: number) => {
```
jusiskin: nit: spelling should be `reuquestsCount` &rarr; `requestsCount`. This same spelling error is repeated in a few places in this file.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
jusiskin: looks good now - thanks!

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
jusiskin: This is much better. Thanks for separating these tests!

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
jusiskin: Looks good now 👍

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,139 @@ describe('SpotEventPluginClient', () => {
...
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    const result = await spotEventPluginClient['getCollection'](type);
+
+    // THEN
+    expect(result).toEqual(expectedResult);
+    // eslint-disable-next-line dot-notation
+    expect(spotEventPluginClient['deadlineClient'].GetRequest).toBeCalledTimes(1);
+    expect(consoleLogMock.mock.calls.length).toBe(1);
+    expect(consoleLogMock.mock.calls[0][0]).toMatch(`Getting ${type} collection:`);
```
jusiskin:
```suggestion
    expect(consoleLogMock).toBeCalledTimes(1);
    expect(consoleLogMock).toBeCalledWith(expect.stringMatching(`Getting ${type} collection:`));
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,139 @@ describe('SpotEventPluginClient', () => {
...
+
+    // eslint-disable-next-line dot-notation
+    spotEventPluginClient['deadlineClient'].GetRequest = jest.fn().mockRejectedValue(new Error(statusMessage));
+    // eslint-disable-next-line dot-notation
+    const result = await spotEventPluginClient['getCollection'](CollectionType.Group);
+
+    // THEN
+    expect(result).toBeUndefined();
+    expect(consoleErrorMock.mock.calls.length).toBe(1);
+    expect(consoleErrorMock.mock.calls[0][0]).toMatch(`Failed to get group collection. Reason: ${statusMessage}`);
```
jusiskin:
```suggestion
    expect(consoleErrorMock).toBeCalledTimes(1);
    expect(consoleErrorMock).toBeCalledWith(expect.stringMatching(`Failed to get group collection. Reason: ${statusMessage}`));
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,139 @@ describe('SpotEventPluginClient', () => {
...
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    spotEventPluginClient['deadlineClient'].GetRequest = jest.fn().mockResolvedValue(invalidGroupResponse);
+    // eslint-disable-next-line dot-notation
+    const result = await spotEventPluginClient['getCollection'](CollectionType.Group);
+
+    // THEN
+    expect(result).toBeUndefined();
+    expect(consoleErrorMock.mock.calls.length).toBe(1);
+    expect(consoleErrorMock.mock.calls[0][0]).toMatch(`Failed to receive a group collection. Invalid response: ${JSON.stringify(invalidGroupResponse.data)}.`);
```
jusiskin:
```suggestion
    expect(consoleErrorMock).toBeCalledTimes(1);
    expect(consoleErrorMock).toBeCalledWith(expect.stringMatching(`Failed to receive a group collection. Invalid response: ${JSON.stringify(invalidGroupResponse.data)}.`));
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,139 @@ describe('SpotEventPluginClient', () => {
...
+    // eslint-disable-next-line dot-notation
+    const result = await spotEventPluginClient['saveCollection'](expectedResult, type);
+
+    // THEN
+    expect(result).toBeTruthy();
+    // eslint-disable-next-line dot-notation
+    expect(spotEventPluginClient['deadlineClient'].PostRequest).toBeCalledTimes(1);
+    expect(consoleLogMock.mock.calls.length).toBe(2);
+    expect(consoleLogMock.mock.calls[0][0]).toMatch(`Saving ${type} collection:`);
+    expect(consoleLogMock.mock.calls[1][0]).toBe(expectedResult);
```
jusiskin:
```suggestion
    expect(consoleLogMock).toBeCalledTimes(2);
    expect(consoleLogMock).toBeCalledWith(expect.stringMatching(`Saving ${type} collection:`));
    expect(consoleLogMock).toBeCalledWith(expectedResult);
```

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,139 @@ describe('SpotEventPluginClient', () => {
...
+    spotEventPluginClient['deadlineClient'].PostRequest = jest.fn().mockRejectedValue(new Error(statusMessage));
+
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    const result = await spotEventPluginClient['saveCollection'](groupsColection, CollectionType.Group);
+
+    // THEN
+    expect(result).toBeFalsy();
+    expect(consoleErrorMock.mock.calls.length).toBe(1);
+    expect(consoleErrorMock.mock.calls[0][0]).toMatch(`Failed to save group collection. Reason: ${statusMessage}`);
```
jusiskin:
```suggestion
    expect(consoleErrorMock).toBeCalledTimes(1);
    expect(consoleErrorMock).toBeCalledWith(expect.stringMatching(`Failed to save group collection. Reason: ${statusMessage}`));
```

jusiskin (changes requested):

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
jusiskin: Great job. Resolved

jusiskin (changes requested):
Functionally and code quality-wise this looks ready to me. I just remembered that we will need to update some related places in our documentation:

1.  https://github.com/aws/aws-rfdk/blob/11e30f6125b06826e1465a21d7562baa980a00dc/packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts#L320-L321
2. https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/deadline/README.md#configure-spot-event-plugin

jusiskin (changes requested):
Found a few more places that mention creating the groups/pools manually:

*   https://github.com/aws/aws-rfdk/blob/11e30f6125b06826e1465a21d7562baa980a00dc/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts#L93-L94
*   https://github.com/aws/aws-rfdk/blob/11e30f6125b06826e1465a21d7562baa980a00dc/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts#L104-L105

jusiskin (dismissed):
Thanks for the great feature enhancement, @kozlove-aws. This looks ready to me!

packages/aws-rfdk/lib/lambdas/nodejs/lib/configure-spot-event-plugin/test/spot-event-plugin-client.test.ts
```diff
@@ -238,4 +256,208 @@ describe('SpotEventPluginClient', () => {
...
+          headers: { 'Content-Type': 'application/json; charset=utf-8' },
+        },
+      );
+    }
+  });
+
+  test.each([
+    [poolsColection.Pools,1],
+    [[], 0],
+  ])('successful call addPool with %s', async (poolsCollection: string[], requestsCount: number) => {
```
jericht: This test looks like it covers the same code path that the above test we're adding in is covering. Should we remove this test?

jericht (changes requested):

jericht (approved):
Looks great, thanks!

jusiskin (approved):

PR #477, opened 2021-06-18
chore(deps): add resolution for diff and remove nodeunit

jericht:
### Notes
- Addresses https://github.com/advisories/GHSA-h6ch-v84p-w6p9
- Removes `nodeunit`. Our unit tests are all using `jest`, so we don't need to keep `nodeunit` around for `cdk-build-tools`. This also means we don't depend on a vulnerable version of `diff`

### Testing
- Ran `build.sh` and ensured the tests passed

----

### DIFF
--- a/tools/cdk-build-tools/bin/cdk-test.ts
+++ b/tools/cdk-build-tools/bin/cdk-test.ts
@@ -1,7 +1,6 @@
-import * as fs from 'fs-extra';
 import * as yargs from 'yargs';
 import { shell } from '../lib/os';
-import { cdkBuildOptions, configFilePath, hasIntegTests, unitTestFiles } from '../lib/package-info';
+import { cdkBuildOptions, hasIntegTests } from '../lib/package-info';
 import { Timers } from '../lib/timer';
 
 async function main() {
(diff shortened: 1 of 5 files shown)


jusiskin (dismissed):

tools/cdk-build-tools/lib/package-info.ts
```diff
@@ -147,6 +140,7 @@ export interface CDKBuildOptions {
    * Whether the package uses Jest for tests.
    * The default is NodeUnit,
```
yashda: We can remove this comment for the default value now. It looks a little confusing.

tools/cdk-build-tools/bin/cdk-test.ts
```diff
@@ -40,40 +27,12 @@ async function main() {
     await shell(options.test, defaultShellOptions);
   }
 
-  const testFiles = await unitTestFiles();
-  const useJest = options.jest;
-
-  if (useJest) {
-    if (testFiles.length > 0) {
-      throw new Error(`Jest is enabled, but ${testFiles.length} nodeunit tests were found!: ${testFiles.map(f => f.filename)}`);
```
yashda, line 48: IMO, we should keep this check to detect if any NodeUnit file was created unintentionally. Such tests will be ignored silently now.

yashda (changes requested):
The changes looks good. Just a couple of minor comments, let me know what are your thoughts.

tools/cdk-build-tools/bin/cdk-test.ts
jericht: Good idea, I've added the check back.

tools/cdk-build-tools/lib/package-info.ts
jericht: Ah right, missed that. Removed

yashda (approved):
Looks good! Thanks Jericho.

jusiskin (approved):

PR #484, opened 2021-06-21
chore(deps): use jsii/superchain:node14 for rfdk development

### LINKED ISSUE #418: Drop Node10, add Node16

kozlove-aws:
### Problem
Image `jsii/superchain` by default is using NodeJS 10.
We need to use NodeJS 14 in building environment script

### Testing
Built and pack our last code with this changes.

----

### DIFF
--- a/scripts/rfdk_build_environment.sh
+++ b/scripts/rfdk_build_environment.sh
@@ -42,5 +42,5 @@ docker run --rm \
     ${USER_OPT} \
     --net=host -it \
     --env DOTNET_CLI_TELEMETRY_OPTOUT=1 \
-    jsii/superchain
+    jsii/superchain:node14
 


jusiskin (changes requested):
The change itself looks good, but I think we need to change the conventional commit. This is a developer-facing change and we're not fixing any bug in the construct library we deliver to users, so I'd argue that the following commit title makes more sense:

```
chore(deps): use jsii/superchain:node14 for rfdk development
```

what do you think?

jusiskin (approved):

horsmand (approved):

PR #491, opened 2021-06-29
feat(deadline): use TLS between RenderQueue and clients by default

### LINKED ISSUE #490: Enable TLS between Render Queue and clients by default
Switch the default behaviour in the Render Queue so that TLS between the ALB and clients is enabled by default.

### Use Case

This will make using TLS require minimal effort and cost for anyone creating an RFDK render farm, while giving them increased security.

### Proposed Solution

To prevent breaking changes to anyone already using TLS in the Render Queue, we will leave the interface as-is, and only add one new `enabled` field in the `RenderQueueExternalTLSProps` interface. You can see the interface and where it fits in to the Render Queue's construct props here:
```
export interface RenderQueueProps {
  readonly trafficEncryption?: RenderQueueTrafficEncryptionProps;
  ...
}
export interface RenderQueueTrafficEncryptionProps {
  readonly externalTLS?: RenderQueueExternalTLSProps;
  ...
}
export interface RenderQueueExternalTLSProps {
  readonly enabled?: boolean;
  readonly acmCertificate?: ICertificate;
  readonly acmCertificateChain?: ISecret;
  readonly rfdkCertificate?: IX509CertificatePem;
}
```

If no certificate or hostname are provided, they will be created by default as follows:
```
const hostname = 'renderqueue';
const zoneName = 'rfdk.internal';

const domainZone = new PrivateHostedZone(this, 'DnsZone', {
  vpc,
  zoneName,
});
const fullyQualifiedDomainName = `${hostname}.${domainZone.zoneName}`

const rootCa = new X509CertificatePem(this, 'RootCA', {
  subject: {
    cn: 'RenderQueueRootCA',
  },
});
const rfdkCert = new X509CertificatePem(this, 'RenderQueueCA', {
  subject: {
    cn: fullyQualifiedDomainName,
  },
  signingCertificate: rootCa,
});
```

### Other

This is a **breaking change** that will enable TLS on farms without it, unless they are modified to set the `enabled` flag to false.

* [x] :warning: This feature might incur a breaking change

---

horsmand:
fixes #490

BREAKING CHANGE: Farms currently not configured to use external TLS on
the Render Queue will be modified to have it enabled and using the
default certificate and hosted zone. To continue to keep external TLS
disabled, the `enabled` flag on the `RenderQueueExternalTLSProps` can be
set to false; however, we strongly encourage you to enable TLS.

## Testing
This has full unit test coverage and was also tested by deploying the render queue and worker fleet integration test suites.

----


examples/deadline/EC2-Image-Builder/python/package/lib/base_farm_stack.py
```diff
@@ -81,4 +83,5 @@ def __init__(self, scope: Construct, stack_id: str, *, props: BaseFarmStackProps
             images=images,
             repository=repository,
             deletion_protection=False,
+            traffic_encryption=RenderQueueTrafficEncryptionProps( external_tls=RenderQueueExternalTLSProps( enabled=False ) ),
```
jusiskin: Any reason we want to explicitly disable TLS for this example?

I imagine it was done to preserve the previous behavior, but if we can now concisely deploy a more secure TLS-enabled RQ, let's change it to promote best-practices.

Same question applies to the TypeScript version

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -148,9 +148,15 @@ export interface RenderQueueHealthCheckConfiguration {
...
+  /**
+   * Whether to enable TLS between the Render Queue and Deadline clients.
+   * @default true
+   */
+  readonly enabled?: boolean;
+
   /**
    * The ACM certificate that will be used for establishing incoming external TLS connections to the RenderQueue.
-   * @default If  not provided then the rfdkCertificate must be provided.
+   * @default: If not provided, rfdkCertificate will be used.
```
jusiskin: Two questions here:

1.  do we want the `:` after `@default`? The [generated docs](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.RenderQueueExternalTLSProps.html#acmcertificate) render properly without them. I'm not sure if the `:` would break anything there.
2. Can we modify this to align more with the doc for `rfdkCertificate`?

    >   If rfdkCertificate and acmCertificate are both not provided, an rfdkCertificate will be generated and used.

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -281,7 +287,7 @@ export interface RenderQueueProps {
   /**
    * Hostname to use to connect to the RenderQueue.
    *
-   * @default A hostname is generated by the Application Load Balancer that fronts the RenderQueue.
+   * @default: A private hosted host will be created and the default hostname will be used.
```
jusiskin: should we document the default hostname and the private hosted zone name here?

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -109,6 +113,42 @@ export interface IRenderQueue extends IConstruct, IConnectable {
...
+}
+
+/**
+ * Interface for information about the render queue's TLS configuration
+ */
+interface TlsInfo {
+  /**
+   * The certificate to use for the TLS connection.
+   */
+  readonly clientCert: ICertificate;
```
jusiskin: I realize this name was copied from the previous code, but it has been a source of confusion whenever looking at the code. Unless I'm mistaken, I believe this contains the server certificate as presented by the Application Load Balancer. If so, can we take the opportunity to rename this to be more semantically correct? Perhaps `serverCert`?

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -692,4 +713,120 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+    });
+    const label = hostname?.hostname ?? RenderQueue.DEFAULT_HOSTNAME;
+    const domainInfo = this.createDomainInfo(label, domainZone);
+
+    const rootCa = new X509CertificatePem(this, 'RootCA', {
+      subject: {
+        cn: 'RenderQueueRootCA',
+      },
+    });
+    const rfdkCert = new X509CertificatePem(this, 'RenderQueueCA', {
```
jusiskin: perhaps `RenderQueuePemCert` is more accurate than `RenderQueueCA`. What do you think?

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -738,99 +740,219 @@ describe('RenderQueue', () => {
...
+            Ref: 'RenderQueueRenderQueueCAPassphrase95F29CE0',
+          },
+          CertChain: {
+            'Fn::GetAtt': [
+              'RenderQueueRenderQueueCA5EEAD7C4',
+              'CertChain',
+            ],
+          },
+        },
+      }));
```
jusiskin: So far this only asserts that we create an ACM import custom resource. A few things:

1.  We can clean this up by reaching into the `RenderQueue` using `.node.findChild(...)` to get a reference to the `X509CertificatePem` instance and using [`stack.resolve(...)`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html#resolveobj) to make this code smaller and more readable:

    ```ts
    const rq = new RenderQueue(isolatedStack, 'RenderQueue', props);

    const certGen = rq.node.findChild('RenderQueueCA') as X509CertificatePem;
    expect(isolatedStack).to(haveResourceLike('Custom::RFDK_AcmImportedCertificate', {
      Cert: isolatedStack.resolve(certGen.GetAtt('Cert')),
      Key: isolatedStack.resolve(certGen.GetAtt('Key')),
      Passphrase: isolatedStack.resolve(certGen.passphrase),
      CertChain: isolatedStack.resolve(certGen.GetAtt('CertChain')),
    }));
    ```
2.  We should also assert the `Custom::RFDK_X509Generator` resource is being created
3.  We should assert that the `ImportedAcmCertificate` is being wired up correctly to the ALB listener

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -1904,6 +2014,42 @@ describe('RenderQueue', () => {
       });
     });
 
+    test('not specified with TLS', () => {
```
jusiskin: It took me a moment to figure out the semantics of the test name here. Perhaps a well-placed comma would help and change to `not specified, with TLS`

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -1904,6 +2014,42 @@ describe('RenderQueue', () => {
...
+
+      const renderQueue = new RenderQueue(isolatedStack, 'RenderQueue', props);
+
+      const loadBalancerLogicalId = dependencyStack.getLogicalId(
+        renderQueue.loadBalancer.node.defaultChild as CfnElement,
+      );
+
+      expectCDK(isolatedStack).to(haveResource('AWS::Route53::RecordSet', {
+        // eslint-disable-next-line dot-notation
+        Name: `${RenderQueue['DEFAULT_HOSTNAME']}.${RenderQueue['DEFAULT_DOMAIN_NAME']}.`,
```
jusiskin: I think this is a scenario where we would duplicate the values of the expected FQDN here rather than referencing the internals of the `RenderQueue` class. This makes the tests anchor to the expected DNS record and protects someone from changing the default and breaking things for users.

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -1134,7 +1227,14 @@ describe('RenderQueue', () => {
...
                 ],
               },
-              `" --render-queue "http://renderqueue.${ZONE_NAME}:8080" \n` +
+              '" --render-queue "http://',
+              {
+                'Fn::GetAtt': [
+                  'RenderQueueLB235D35F4',
+                  'DNSName',
+                ],
+              },
```
jusiskin: this looks like it could this be converted to a more readable `isolatedStack.resolve(...)` expression. What do you think?

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -1904,6 +2014,42 @@ describe('RenderQueue', () => {
...
+        // eslint-disable-next-line dot-notation
+        Name: `${RenderQueue['DEFAULT_HOSTNAME']}.${RenderQueue['DEFAULT_DOMAIN_NAME']}.`,
+        Type: 'A',
+        AliasTarget: objectLike({
+          HostedZoneId: {
+            'Fn::GetAtt': [
+              loadBalancerLogicalId,
+              'CanonicalHostedZoneID',
+            ],
+          },
```
jusiskin: Does this simplification work?

```suggestion
          HostedZoneId: isolatedStack.resolve(renderQueue.loadBalancer.loadBalancerCanonicalHostedZoneId),
```

jusiskin (changes requested):
Great work! Just a few minor suggestions

examples/deadline/EC2-Image-Builder/python/package/lib/base_farm_stack.py
horsmand: Removed the line to disable TLS for both python and TS

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
horsmand: Updated

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
horsmand: Updated

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
horsmand: Updated

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
horsmand: Yeah I'm not sure why I had CA in there, that doesn't make sense.

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
horsmand: Updated the `Custom::RFDK_AcmImportedCertificate` expectation and added the other two.

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
horsmand: Updated

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
horsmand: Makes sense

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
horsmand: Updated

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
horsmand: Yep this works

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -148,9 +151,15 @@ export interface RenderQueueHealthCheckConfiguration {
...
+  /**
+   * Whether to enable TLS between the Render Queue and Deadline clients.
+   * @default true
+   */
+  readonly enabled?: boolean;
+
   /**
    * The ACM certificate that will be used for establishing incoming external TLS connections to the RenderQueue.
-   * @default If  not provided then the rfdkCertificate must be provided.
+   * @default If rfdkCertificate and acmCertificate are both not provided, an rfdkCertificate will be generated and used.
```
kozlove-aws: It is obvious but maybe will be worth mention that these certificates will be used or created only when TLS is enabled?

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -295,38 +339,23 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
-          throw new Error('externalTLS.acmCertificateChain must be provided when using externalTLS.acmCertificate.');
-        }
-        this.clientCert = props.trafficEncryption.externalTLS.acmCertificate;
-        this.certChain = props.trafficEncryption.externalTLS.acmCertificateChain;
-      }
-    } else {
-      externalProtocol = ApplicationProtocol.HTTP;
+    const externalProtocol = props.trafficEncryption?.externalTLS?.enabled === false ? ApplicationProtocol.HTTP : ApplicationProtocol.HTTPS;
+    let loadBalancerFQDN: string | undefined;
+    let domainZone: IHostedZone | undefined;
```
kozlove-aws, line 334: As I can see now if TLS is disabled we leave `domainZone` and `domainName` undefined.
Should we use previous logic or this was done intentionally?

kozlove-aws (commented):
Looks good, just want to clarify a couple moments.

examples/deadline/EC2-Image-Builder/python/package/lib/base_farm_stack.py
jusiskin: Looks good. thanks!

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Thanks! much better

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
jusiskin: This is great! Thanks and resolvved

jusiskin (changes requested):
This is looking good to me. Only one more request...

Since this is a breaking change, do you mind adding upgrade documentation to `packages/aws-rfdk/docs/upgrade` like we have for 0.27.0? This should help guide a RFDK user through the process of upgrading and determining whether they need to take any action and if so, what action to take.

examples/deadline/EC2-Image-Builder/python/package/lib/base_farm_stack.py
```diff
@@ -15,6 +15,8 @@
 from aws_rfdk.deadline import (
     AwsThinkboxEulaAcceptance,
     RenderQueue,
+    RenderQueueExternalTLSProps,
+    RenderQueueTrafficEncryptionProps,
```
jusiskin: I suppose these are no longer needed, right?

horsmand: Correct, I've removed these.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
horsmand: Good catch, I added test cases for this and fixed it.

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
horsmand: Updated

packages/aws-rfdk/docs/upgrade/upgrading-0.37.md
```diff
@@ -0,0 +1,41 @@
...
+
+There will be a few new constructs deployed to your farm:
+1. A `PrivateHostedZone` will be created if you do not supply your own. We set the default domain to `aws-rfdk.com`, which we have registered and suggest that you use if you do not have your own registered domain. [RFC 6762](https://datatracker.ietf.org/doc/html/rfc6762#appendix-G) recommends against using any unregistered top-level domains.
+1. A self-signed X509 certificate will be generated using OpenSSL and that will then be used to sign a certificate that the Render Queue will use for TLS. Specifically, the certificate will be passed to the Application Listener for the Application Load Balancer that the Render Queue creates. Additional details about how RFDK uses TLS can the built-in certificate management can be found in the developer guide for [Encryption in transit](https://docs.aws.amazon.com/rfdk/latest/guide/security-encrypt-in-transit.html).
+
+These new constructs will require the Render Queue load balancer's listener to need replacing, but the load balancer itself and the backing services it redirects traffic to will not need to be changed.
+
+### WorkerInstanceFleet Changes
```
jusiskin: Good catch to think about this and the CDK ["deadly embrace" issue](https://github.com/aws/aws-cdk/issues/7602) that would arise.

I'm curious what happens with the [`ConfigureSpotEventPlugin` construct](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.ConfigureSpotEventPlugin.html) in this upgrade scenario. Has this been tested? My theory is that it would unfold as follows:

1. The deployment would replace the RQ ALB listener
2. The deployment would proceed to deploy the update to the `Custom::RFDK_ConfigureSpotEventPlugin` which would reconfigure the Deadline Spot Event Plugin with the new `RenderQueue` configuration. Changing the Spot Event Plugin configuration does not immediately apply changes to the fleets that it manages. The Spot Event Plugin evaluates its configuration and applies spot fleet request changes [at the next Deadline house cleaning event](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/event-spot.html#lifecycle).
3. Fork depending on if [the ResourceTracker is enabled](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.SpotEventPluginSettings.html#enableresourcetracker):
    1.  **If enabled**, the existing `SpotEventPluginFleet`s would lose their connection to the RQ. There would be a race condition that depends on the Deadline Resource Tracker's configured [fleet fault tolerance and thresholds](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/resource-tracker-overview.html#fleet-fault-tolerance-and-termination-threshold). The Resource Tracker would start by rebooting unhealthy instances and eventually mark the fleet as unhealthy and would [need to be reset](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/resource-tracker-troubleshooting.html#unable-to-start-new-instances).
    2.  **If disabled**, the new Deadline Spot Event Plugin configuration would be applied to the existing spot fleet request at the next Deadline house cleaning event and Spot Workers would resume operation.

packages/aws-rfdk/docs/upgrade/upgrading-0.37.md
```diff
@@ -0,0 +1,41 @@
+# Upgrading to RFDK v0.37.x or Newer
```
jusiskin, line 1: Great job with this upgrade doc. Do you mind adding a link to this from `packages/aws-rfdk/docs/upgrade/index.md`?

packages/aws-rfdk/docs/upgrade/upgrading-0.37.md
```diff
@@ -0,0 +1,41 @@
...
+
+Starting in RFDK v0.37.0, the default for TLS between the render queue and its clients, which is configured using the `RenderQueueExternalTLSProps` interface that the `RenderQueue` construct takes as a part of its constructor props, is now set to be enabled.
+
+## Upgrading Farms Already Using TLS
+
+If you are already setting fields on the `RenderQueueExternalTLSProps` for the Render Queue, no action is required. Redeploying your render farm after upgrading your version of RFDK should have no effect.
+
+## Upgrading Farms Not Using TLS
```
jusiskin, line 9: While you have the options documented in this file, perhaps it would help make things clearer if we added a paragraph at this level (just below this heading) outlining the two upgrade paths with links to each sub-section:

1.  Migrating to TLS (recommended)
2.  Preserving plain HTTP

Do you think it would help?

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -1134,7 +1243,7 @@ describe('RenderQueue', () => {
                   },
                 ],
               },
-              `" --render-queue "http://renderqueue.${ZONE_NAME}:8080" \n` +
+              '" --render-queue "http://renderqueue.renderfarm.local:8080" \n' +
```
jusiskin: Should we have changed this? The previous code avoided duplication and more clearly articulated the expected semantics of the behavior.

Same feedback applies for line 1417 below

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -2097,6 +2088,50 @@ describe('RenderQueue', () => {
...
+        AliasTarget: objectLike({
+          HostedZoneId: {
+            'Fn::GetAtt': [
+              loadBalancerLogicalId,
+              'CanonicalHostedZoneID',
+            ],
+          },
+        }),
+      }));
+    });
```
jusiskin: nit: missing the `GIVEN` and `THEN` comment headers.

jusiskin (changes requested):
Just a few more suggestions based on the new commit.

kozlove-aws (dismissed):

packages/aws-rfdk/docs/upgrade/upgrading-0.37.md
jusiskin: Looks good! resolved

packages/aws-rfdk/docs/upgrade/upgrading-0.37.md
```diff
@@ -0,0 +1,55 @@
...
+
+Since the endpoint and port the listener on the load balancer uses will be changed, and the TLS will require any clients connecting to verify its certificate, any stacks that contain dependencies on the Render Queue will first need to be destroyed. If you are using a tiered architecture similar to what we recommend in our documentation, this would include any `WorkerInstanceFleet` constructs or `SpotEventPluginFleet` and `ConfigureSpotEventPlugin` constructs. If you are not using a tiered architecture, we still recommend destroying these constructs since we're changing the endpoint that they need to connect to, and that configuration of the endpoint happens in the initialization script for an instance.
+
+To perform the removal of these constructs:
+1. Suspend any jobs that are being run by workers that the constructs deployed.
+1. If any spot instances are running, they will need to be terminated since their lifecycle is controlled by Deadline's Spot Event Plugin and not your RFDK app. Trying to destroy/remove the `SpotEventPluginFleet` construct will fail if these hosts are left running because the spot instances use the security group that the construct creates.
```
jusiskin: > If any spot instances are running

The `WorkerInstanceFleet` can deploy spot instances with the [`spotPrice` prop](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.WorkerInstanceFleet.html#spotprice). Perhaps we should qualify this with:

>  If you are using the `ConfigureSpotEventPlugin` and `SpotEventPluginFleet` constructs, then any spot fleets launched by the Spot Event Plugin will need to be terminated ....

packages/aws-rfdk/docs/upgrade/upgrading-0.37.md
```diff
@@ -0,0 +1,55 @@
...
+1. Next we have to destroy the constructs that are deploying workers, which could be done in a few ways:
+  1. If you can destroy the Stack that contains these constructs without destroying the rest of your app, then destroy it using the command `cdk destroy "ComputeTier"` (or whatever name you gave your stack).
+  1. If you cannot destroy a single stack or you are not using a tiered architecture, you can just comment out these constructs in your app, rebuild it, and then run `cdk deploy "*"` to perform the removal.
```
jusiskin: GitHub flavored markdown doesn't register the indentation as a sub-list here. Can we also make the conditions bold at the beginning of each bullet point to help visually guide the reader?

packages/aws-rfdk/docs/upgrade/upgrading-0.37.md
```diff
@@ -0,0 +1,55 @@
...
+1. Now that we won't cause any dependency issues, we can update the version of RFDK in our app's `package.json`, install it, and then run the deployment. If worker constructs were commented out in the last step, they can be added back in here and redeployed during the upgrade.
```
jusiskin: This is assuming TypeScript (`package.json`). It's probably okay to leave out the specifics of updating the TS/Python packages and just say something like "upgrade the RFDK version"

jusiskin (changes requested):
Just a few tiny suggestions left for the upgrade docs but otherwise this looks good to me.

packages/aws-rfdk/docs/upgrade/upgrading-0.37.md
jusiskin: Thanks for looking into this. Looks like it wasn't very straight-forward of an upgrade process, so it's helpful that you've tested and have a documented process for doing this cleanly. 👏

jusiskin (approved):
Great work @horsmand. Thanks for making RFDK more secure by default and providing docs to guide customers through this breaking change.

This looks good to me

kozlove-aws (approved):

PR #513, opened 2021-07-27
feat(deadline): change SEP construct to use launch templates instead of launch specifications

jericht:
### Notes
Refactors the code around the Spot Event Plugin construct to use Launch Templates instead of Launch Specifications.

### Testing
- Updated unit tests
- Deployed the SEP example app and verified SEP works as intended by:
    - Submitting jobs in Deadline to the configured SEP group
    - Verifying a new Spot Fleet Request was spun up using the Launch Template created by the SEP construct
- Verified the SFRs created by SEP use the latest version of the Launch Template, without requiring a redeploy, by first rendering a job with SEP, terminating the old SEP instance, updating the Launch Template in the AWS console, then submitting a new job and ensured the Worker spun up used the new launch template version.

----


gdday (changes requested):
How feasible would it be to prevent tying the Spot Event Plugin to the Launch Template version that was created and, instead, using the latest version of the Launch Template?

That would let users edit the Launch Template in the AWS Console rather than making their changes in their CDK application.

jericht:
@gdday So Spot does support using a special value `$Latest` for Launch Templates that would achieve this, however it looks like it's unfortunately [not supported by CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-fleetlaunchtemplatespecification.html):
> The version number of the launch template. You must specify a version number. AWS CloudFormation does not support specifying $Latest or $Default for the template version number.

Maybe there's a way to do this with a custom resource that would update that value via calling the appropriate AWS APIs though, I'll look into seeing what that would look like

gdday:
> Maybe there's a way to do this with a custom resource that would update that value via calling the appropriate AWS APIs though, I'll look into seeing what that would look like

Right, I think you could do it via the [describe-launch-templates API](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-launch-templates.html) in Lambda. I don't know how much additional work that would be, but I do think that would be useful for customers.

examples/deadline/All-In-AWS-Infrastructure-SEP/python/package/config.py
```diff
@@ -19,5 +19,11 @@ def __init__(self):
         #  should match the one used for staging the render queue and usage based licensing recipes.
         self.deadline_client_linux_ami_map: Mapping[str, str] = {'us-west-2': 'ami-0c8431fc72742c110'}
 
+        # Whether the DeadlineResourceTracker stack and supporting resources already exist in the account/region you are deploying to.
+        #
+        # If this is false, resources required by the Deadline Resource Tracker will be deployed into your account.
+        # If this is true, these resources will be skipped.
+        self.deadline_resource_tracker_exists: bool = False
```
jusiskin: I'm wondering if it would help to clarify things a bit more here. Specifically:

1.  the use of "stack" here:

    >   Whether the DeadlineResourceTracker **stack** and supporting resources already exist ...
2.  the property name (`deadline_resource_tracker_exists`)

Right now, this property controls whether the `DeadlineResourceTrackerAccessRole` IAM role gets deployed as part of the CDK example app. Deadline requires that users create this role up-front or use the AWS Portal Server installer to create it ([ref](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/resource-tracker-overview.html#resource-tracker-identity-and-access-management-iam-role)):

>   The Deadline Resource Tracker requires an IAM Role named “DeadlineResourceTrackerAccessRole” in order to access AWS resources in your account.
> * For AWS Portal, this role is created automatically as part of the AWS Portal Server installer’s IAM resource creation step.
> * For the Spot Event Plugin, you must create the role manually as described below.

Do you think it would be clearer if we renamed the prop to `create_resource_tracker_role` and make this a little more explicit? Something like:

>   Whether the DeadlineResourceTrackerAccessRole IAM role required by Deadline's Resource Tracker should be created in this CDK app.
>
> If you have previously used this same AWS account with either Deadline's AWS Portal feature or Spot Event Plugin and had used the Deadline Resource Tracker, then you likely have this IAM role in your account already unless you have removed it.
>
>  Note: Deadline's Resource Tracker only supports being used by a single Deadline Repository per AWS account.

What do you think? Same feedback applies to the TypeScript example

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
```diff
@@ -522,51 +521,52 @@ export class ConfigureSpotEventPlugin extends Construct {
    * Each congiguration is a mapping between one Deadline Group and one Spot Fleet Request Configuration.
```
gdday: You didn't change this but... congiguration?

gdday (dismissed):
Thank you for adding support for $Latest. This will be very helpful for customers who want to quickly modify the Launch Template without having to redeploy.

This looks good to me, I don't see any issues!

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
```diff
@@ -522,51 +521,52 @@ export class ConfigureSpotEventPlugin extends Construct {
...
     });
 
     const spotFleetRequestProps: SpotFleetRequestProps = {
       AllocationStrategy: fleet.allocationStrategy,
       IamFleetRole: fleet.fleetRole.roleArn,
-      LaunchSpecifications: launchSpecifications,
+      LaunchTemplateConfigs: launchTemplates.map(launchTemplate => {
+        return {
+          LaunchTemplateSpecification: {
+            Version: '$Latest',
```
jusiskin: nit: we can use `LaunchTemplateSpecialVersions.LATEST_VERSION` which is the named constant for this value.

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
```diff
@@ -522,51 +521,52 @@ export class ConfigureSpotEventPlugin extends Construct {
...
+      if (fleet.tags.hasTags()) {
+        const tags = fleet.tags.renderTags() as {[key: string]: string}[];
+        tags.forEach(tag => Tags.of(launchTemplate).add(tag.Key, tag.Value));
+      }
+
+      // Add tags that are added via the Tags API which uses Aspects and is a layer above the TagManager API
+      const tagAspects = Aspects.of(fleet).aspects.filter(aspect => aspect instanceof Tag) as Tag[];
+      tagAspects.forEach(tagAspect => Tags.of(launchTemplate).add(tagAspect.key, tagAspect.value));
+
+      return launchTemplate;
```
jusiskin: I wonder if it make sense to create the launch template in the `SpotEventPluginFleet` construct and reference it here instead. What do you think?

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
```diff
@@ -522,51 +521,52 @@ export class ConfigureSpotEventPlugin extends Construct {
...
+      // Add tags that are added via the TagManager API
+      /* istanbul ignore if */
+      if (fleet.tags.hasTags()) {
+        const tags = fleet.tags.renderTags() as {[key: string]: string}[];
+        tags.forEach(tag => Tags.of(launchTemplate).add(tag.Key, tag.Value));
+      }
+
+      // Add tags that are added via the Tags API which uses Aspects and is a layer above the TagManager API
+      const tagAspects = Aspects.of(fleet).aspects.filter(aspect => aspect instanceof Tag) as Tag[];
+      tagAspects.forEach(tagAspect => Tags.of(launchTemplate).add(tagAspect.key, tagAspect.value));
```
jusiskin: Not sure where to add this comment, but when using launch templates with the Deadline's Spot Event Plugin and the Deadline Resource Tracker is enabled, we need to be sure to add the `DeadlineTrackedAWSResource` tag with a value of `"SpotEventPlugin"` to the launch template. When the Resource Tracker is disabled, we need to add a tag with the key `DeadlineResourceTracker` with a value of `"SpotEventPlugin"`. This is what the Spot Event Plugin does for launch configurations, but it cannot do this for launch templates, since tag overrides are not possible at the present time.

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
```diff
@@ -522,51 +521,52 @@ export class ConfigureSpotEventPlugin extends Construct {
...
       IamFleetRole: fleet.fleetRole.roleArn,
-      LaunchSpecifications: launchSpecifications,
+      LaunchTemplateConfigs: launchTemplates.map(launchTemplate => {
+        return {
+          LaunchTemplateSpecification: {
+            Version: '$Latest',
+            LaunchTemplateId: launchTemplate.launchTemplateId,
+            LaunchTemplateName: launchTemplate.launchTemplateName,
+          },
+          Overrides: [{ SubnetId: subnetId }],
```
jusiskin: right now, `subnetId` is a comma-separated list of subnet IDs. From what I understand, this is not valid in `LaunchTemplateConfigs[*].Overrides.SubnetId` ([ref docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateOverrides.html)).

I believe we must map each launch template into a list of `LaunchTemplateConfigs` - each with their own subnet override.

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
```diff
@@ -522,51 +521,52 @@ export class ConfigureSpotEventPlugin extends Construct {
...
-        // Need to convert from IResolvable to bypass TypeScript
-        SecurityGroups: (securityGroupsToken  as unknown) as SpotFleetSecurityGroupId[],
-        SubnetId: subnetId,
-        // Need to convert from IResolvable to bypass TypeScript
-        TagSpecifications: (instanceTagsToken as unknown) as SpotFleetTagSpecification[],
-        UserData: userDataToken,
-        InstanceType: instanceType.toString(),
-      };
-      launchSpecifications.push(launchSpecification);
+    const launchTemplates: LaunchTemplate[] = fleet.instanceTypes.map((instanceType, idx) => {
```
jusiskin: Rather than create a launch template resource for every instance type, we can create a single one. Then when we are specifying the `LaunchTemplateConfigs` for the Spot Event Plugin configuration, we can create multiple configs pointing to the same launch template but with different instance type overrides.

This helps reduce the number of launch template resources that we'd deploy and reduce the number of CFN resources that contribute to the per-stack/template limit. What do you think?

packages/aws-rfdk/lib/deadline/test/configure-spot-event-plugin.test.ts
```diff
@@ -432,226 +428,94 @@ describe('ConfigureSpotEventPlugin', () => {
...
-            LaunchSpecifications: arrayWith(objectLike({
-              BlockDeviceMappings: arrayWith(objectLike({
-                DeviceName: deviceName,
-                NoDevice: '',
-              })),
-            })),
-          }),
+      cdkExpect(stack).to(haveResourceLike('AWS::EC2::LaunchTemplate', {
+        LaunchTemplateData: objectLike({
+          SecurityGroupIds: fleet2.securityGroups.map(sg => stack.resolve(sg.securityGroupId)),
```
jusiskin: nit: can we go with the slightly more explicit test:

```suggestion
          SecurityGroupIds: securityGroups.map(sg => stack.resolve(sg.securityGroupId)),
```

The current code does not necessarily assert that `fleet2.securityGroups` matches what was passed in via the props.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/conversion.test.ts
```diff
@@ -440,92 +337,320 @@ describe('isValidTagSpecification', () => {
...
+      Version: '1',
+      // @ts-ignore
+      LaunchTemplateId: 123,
+    };
+
+    // WHEN
+    expect(() => validateLaunchTemplateSpecification(spec, propertyName))
+
+      // THEN
+      .toThrow();
```
jusiskin: should we assert anything about the error message? same feedback applies for the tests below

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/conversion.ts
```diff
@@ -122,79 +101,35 @@ export function convertTagSpecifications(tagSpecifications: SpotFleetTagSpecific
...
-      SecurityGroups,
-      SubnetId: validateStringOptional(launchSpecification.SubnetId, `${propertyName}.SubnetId`),
-      TagSpecifications,
-      UserData: validateString(launchSpecification.UserData, `${propertyName}.UserData`),
-      InstanceType: validateString(launchSpecification.InstanceType, `${propertyName}.InstanceType`),
-    };
-    convertedLaunchSpecifications.push(convertedLaunchSpecification);
+    validateProperty(input => Array.isArray(input), ltc.Overrides, `${propertyName}.Overrides`);
+    ltc.Overrides.forEach(override => validateLaunchTemplateOverrides(override, `${propertyName}.Overrides`));
   });
```
jusiskin: Can we include the array index in the error message? I think we can get it by accepting an additional argument in the callback for `launchTemplateConfigs.forEach`:

```ts
launchTemplateConfigs.forEach((ltc, index) => {
   // ...
   validateProperty(..., ..., `${propertyName}[${index}].LaunchTemplateSpecification`);
});
```

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -260,15 +132,17 @@ describe('SEPConfiguratorResource', () => {
       jest.clearAllMocks();
     });
 
-    test('with no configs', async () => {
+    test('saves server data with no configuration', async () => {
```
jusiskin: Is this test name accurate? It appears like we are asserting:

```ts
expect(mockSaveServerData.mock.calls.length).toBe(0);
```

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -278,116 +152,31 @@ describe('SEPConfiguratorResource', () => {
...
-      };
-
-      // WHEN
-      await handler.doCreate('physicalId', noEbs);
-      const calledWithString = mockSaveServerData.mock.calls[0][0];
-      const calledWithObject = JSON.parse(calledWithString);
-
-      // THEN
-      expect(calledWithObject).toEqual(convertedNoEbs);
+      expect(JSON.stringify(calledWithObject)).toEqual(JSON.stringify(validSepConfiguration.spotFleetRequestConfigurations));
```
jusiskin: can this be simplified to:

```ts
expect(mockSaveServerData).toBeCalledWith(JSON.stringify(validSepConfiguration.spotFleetRequestConfigurations));
```

jusiskin (changes requested):
This is great work, @jericht. The three primary concerns are:

1.  Move the launch template to be created in the `SpotEvenPluginFleet` if possible
2.  Make sure the required tags are added to launch templates for the Deadline Resource Tracker to function properly
3.  Map each `SpotEventPluginFleet` subnet to an individual `LaunchTemplateConfig` instead of passing an invalid comma-separated list of subnet IDs

Otherwise, there are a few minor semantic/polish suggestions.

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
jericht: Nice catch, thanks! Fixed

packages/aws-rfdk/lib/deadline/test/configure-spot-event-plugin.test.ts
jericht: Good catch, updated

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
jericht: Sure, updated

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
jericht: Yeah, I agree, that seems more organized. Updated

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
jericht: Yeah, that does sound cleaner. Updated

examples/deadline/All-In-AWS-Infrastructure-SEP/python/package/config.py
jericht: Yeah I was struggling to find a good explanation for this flag since it's pretty specific to Deadline, but I think your description does a good job of explaining it. Updated

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/conversion.ts
jericht: Sure, added it in for this as well as the `Overrides`

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/conversion.test.ts
jericht: Sure, added in expected error messages

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
jericht: Ah, yeah that test name is wrong. Thanks for catching it, updated

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
jericht: Yep, updated

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
jericht: Right, I missed that. Updated.

packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
jericht: Updated

jericht:
@jusiskin @gdday I've retested this end-to-end and verified SEP is still working (as well as the Resource Tracker tags being applied to the spot instances). The CI build is failing because something is up with the CI runners that is causing an expected crash. I'll look into getting those fixed, but in the meantime this is ready for re-review

examples/deadline/All-In-AWS-Infrastructure-SEP/python/package/app.py
```diff
@@ -45,6 +45,7 @@ def main():
     sep_props = sep_stack.SEPStackProps(
         docker_recipes_stage_path=os.path.join(os.path.dirname(os.path.realpath(__file__)), os.pardir, 'stage'),
         worker_machine_image=MachineImage.generic_linux(config.deadline_client_linux_ami_map),
+        deadline_resource_tracker_exists=config.create_resource_tracker_role,
```
jusiskin: this logic doesn't seem intuitive...

shouldn't it be:

```suggestion
        deadline_resource_tracker_exists=(not config.create_resource_tracker_role),
```

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
```diff
@@ -214,6 +220,13 @@ export interface SpotEventPluginFleetProps {
    * @default: Not used.
    */
   readonly userDataProvider?: IInstanceUserDataProvider;
+
+  /**
+   * Whether the instances in the Spot Fleet should be tracked by Deadline Resource Tracker.
+   *
+   * @default true
+   */
+  readonly trackInstancesWithResourceTracker?: boolean;
```
jusiskin, line 233: I like option of disabling the resource tracker on a per-fleet basis.

One potential point of confusion that users might encounter is finding this API but not realizing how it interacts with [`SpotEventPluginSettings.enableResourceTracker`](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.SpotEventPluginSettings.html#enableresourcetracker). We might want to document this in both places - here and also in the API docs for `SpotEventPluginSettings.enableResourceTracker`:

https://github.com/aws/aws-rfdk/blob/68050c16bfaa9e273436d41a0ae60da506e20750/packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts#L158-L164


what do you think?

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/conversion.test.ts
```diff
@@ -440,92 +337,333 @@ describe('isValidTagSpecification', () => {
...
+    };
+
+    // WHEN
+    expect(() => validateLaunchTemplateConfigs([config], propertyName))
+
+      // THEN
+      .toThrowError(`${propertyName}[0].LaunchTemplateSpecification type is not valid.`);
+  });
+
+  test('throws when LaunchTemplateSpecification is invalid', () => {
```
jusiskin:
```suggestion
  test('throws when Version is invalid', () => {
```

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -458,63 +247,66 @@ describe('SEPConfiguratorResource', () => {
...
-      const mockSaveServerData = jest.fn( (a) => returnFalse(a) );
+      const mockSaveServerData = jest.fn( (_a) => Promise.resolve(false) );
       mockSpotEventPluginClient.saveServerData = mockSaveServerData;
 
       // WHEN
-      const promise = handler.doCreate('physicalId', noPluginConfigs);
+      const promise = handler.doCreate('physicalId', {
+        connection: validSepConfiguration.connection,
+        spotFleetRequestConfigurations: validSepConfiguration.spotFleetRequestConfigurations,
+      });
```
jusiskin: I might be missing something, but can we simplify this to:

```suggestion
      const promise = handler.doCreate('physicalId', validSepConfiguration);
```

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/test/handler.test.ts
```diff
@@ -524,11 +316,14 @@ describe('SEPConfiguratorResource', () => {
...
-      const mockConfigureSpotEventPlugin = jest.fn( (a) => returnFalse(a) );
+      const mockConfigureSpotEventPlugin = jest.fn( (_a) => Promise.resolve(false) );
       mockSpotEventPluginClient.configureSpotEventPlugin = mockConfigureSpotEventPlugin;
 
       // WHEN
-      const promise = handler.doCreate('physicalId', noFleetRequestConfigs);
+      const promise = handler.doCreate('physicalId', {
+        connection: validSepConfiguration.connection,
+        spotPluginConfigurations: validSepConfiguration.spotPluginConfigurations,
+      });
```
jusiskin: one more:

```suggestion
      const promise = handler.doCreate('physicalId', validSepConfiguration);
```

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
```diff
@@ -418,6 +437,16 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
...
 
+  /**
+   * The Launch Template for this Spot Fleet. This launch template does not specify an instance type or subnet.
+   */
+  public readonly launchTemplate: LaunchTemplate;
+
+  /**
+   * @internal
+   */
+  public readonly _launchTemplateConfigs: any[];
```
jusiskin: Can we type this?

```suggestion
  public readonly _launchTemplateConfigs: LaunchTemplateConfig[];
```

jusiskin (changes requested):
Just a few minor suggestions here, but this is shaping up really nicely

packages/aws-rfdk/lib/deadline/test/configure-spot-event-plugin.test.ts
```diff
@@ -434,226 +412,64 @@ describe('ConfigureSpotEventPlugin', () => {
       })));
     });
 
-    test('fleet with block devices', () => {
+    test('adds multiple fleet security groups to launch template', () => {
```
jusiskin: Should we move this test into `packages/aws-rfdk/lib/deadline/test/spot-event-plugin-fleet.test.ts`? Same question for the test below that tests the `AWS::EC2::LaunchTemplate` resource.

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
jericht: @jusiskin Sure, updated.

@horsmand I prefixed this with an underscore to show that it's not meant to be a public API and is an implementation detail. What do you think?

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
jericht: Good idea. Added a little blurb to the docstring for each prop.

examples/deadline/All-In-AWS-Infrastructure-SEP/python/package/app.py
jericht: Ah thanks for catching this. I forgot to update the named argument after changing the prop name. Updated

packages/aws-rfdk/lib/deadline/test/configure-spot-event-plugin.test.ts
jericht: Good catch, yeah we can now that we've moved the logic over. Updated

examples/deadline/All-In-AWS-Infrastructure-SEP/python/package/app.py
horsmand: I don't see `deadline_resource_tracker_exists` in the `SEPStackProps` at all. Should this be named `create_resource_tracker_role`?

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
horsmand: Why does this start with an underscore?

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
```diff
@@ -465,6 +494,7 @@ export class SpotEventPluginFleet extends Construct implements ISpotEventPluginF
     this.osType = imageConfig.osType;
     this.userData = props.userData ?? imageConfig.userData;
     this.imageId = imageConfig.imageId;
```
horsmand: Is this still used anywhere?

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/conversion.ts
```diff
@@ -27,7 +25,9 @@ export function convertSpotFleetRequestConfiguration(spotFleetRequestConfigs: Sp
     const convertedSpotFleetRequestProps: SpotFleetRequestProps = {
       AllocationStrategy: validateString(sfrConfigs.AllocationStrategy, `${group_name}.AllocationStrategy`),
       IamFleetRole: validateString(sfrConfigs.IamFleetRole, `${group_name}.IamFleetRole`),
-      LaunchSpecifications: convertLaunchSpecifications(sfrConfigs.LaunchSpecifications, `${group_name}.LaunchSpecifications`),
+      // Empty array needed for compatibility with SEP since it expects an array for the LaunchSpecifications property
+      LaunchSpecifications: [],
+      LaunchTemplateConfigs: sfrConfigs.LaunchTemplateConfigs ? validateLaunchTemplateConfigs(sfrConfigs.LaunchTemplateConfigs, `${group_name}.LaunchTemplateConfigs`) : undefined,
```
horsmand, line 30: Is it a valid use case if LaunchTemplateConfigs is empty?

horsmand (changes requested):

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
jericht: Good catch, it's not anymore. Removed

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/conversion.ts
jericht: Good call out. We need either `LaunchSpecifications` or `LaunchTemplateConfigs` to be specified for the `request-spot-fleet` API. This is being enforced in our `SpotEventPluginFleet` construct, since it generates these configs based on the instance types and subnets provided, which we ensure we have at least one of each over [here](https://github.com/aws/aws-rfdk/blob/94988f2d1337bb7507ef9ea58695a27f72ab14f1/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts#L526-L537). So, right now users won't run into an issue where they have an empty configs list here. Since this file is focused on converting and not so much on validation, I'm thinking that we can just leave this as is. How does that sound?

jericht:
@horsmand @jusiskin I also changed the initial commit message to have a BREAKING CHANGE section and added some upgrade documentation since old SFRs will be orphaned if users try to redeploy while they have active SFRs that don't use launch templates.

I also did a final E2E test (deploy SEP example and render a cmd line job with SEP), so if everything looks good we can merge this in.

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
jusiskin: Looks great! resolved

packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts
jusiskin: Hence the `@internal` tag above. Looks like [JSII might filter out such fields](https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/):

>   ... they do not apply to any internal declarations such as `private` members and declarations annotated with the `/** @internal */` tag.

jusiskin: Experimented with our documentation generation and it omits this b/c it is marked `@internal`.  I'm okay with keeping the `public` access modifier along with `@internal` and the signaling you've added to communicate that this is an internal API.

Resolving.

packages/aws-rfdk/lib/lambdas/nodejs/configure-spot-event-plugin/conversion.ts
horsmand: Sounds good to me

horsmand (dismissed):

horsmand:
Is there any easy way to preview the diagram?

jericht:
> Is there any easy way to preview the diagram?

@horsmand Yep, you can click the rich diff button in the top right of the file panel for the svg files

horsmand (approved):

jusiskin (approved):
Looks good to me! This will be a really great enhancement for RFDK users. Thanks for your hard work, @jericht

PR #514, opened 2021-07-29
feat(deadline): add Secret Management support for Repository

kozlove-aws:
### Problem
We need to have possibility to enable Secret Management in repository construct, also Secret Management should be enabled by default.

### Testing
 - Added unit tests
 - Deployed example and validated that Secret Management is enabled.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/repository.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/repository.ts
@@ -286,9 +286,9 @@ export interface RepositorySecurityGroupsOptions {
 export interface SecretsManagementProps {
   /**
    * Whether or not to enable the Secrets Management feature.
-   * @default true
    */
-  readonly enabled?: boolean;
+  readonly enabled: boolean;
+
   /**
    * A Secret containing the username and password to use for the admin role.
    * The contents of this secret must be a JSON document with the keys "username" and "password". ex:
@@ -296,7 +296,9 @@ export interface SecretsManagementProps {
    *         "username": <admin user name>,
    *         "password": <admin user password>,
    *     }
-   * Password should contain at least one lowercase letter, one uppercase letter, one symbol and one number.
+   * Password should be at least 8 characters long and contain at least one lowercase letter, one uppercase letter, one symbol and one number.
+   * In the case when the password does not meet the requirements, the repository construct will fail to deploy.
+   * It is highly recommended that you leave this parameter undefined to enable the automatic generation of a strong password.
    *
    * @default: A random username and password will be generated in a Secret with ID `SMAdminUser` and will need to be retrieved from AWS Secrets Manager if it is needed
    */
@@ -427,7 +429,7 @@ export interface RepositoryProps {
    * https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html
    * @default: Secrets Management will be enabled and a username and password will be automatically generated if none are supplied.
    */
-  readonly secretsManagementSettings?: SecretsManagementProps
+  readonly secretsManagementSettings?: SecretsManagementProps;
 }
 
 /**
@@ -519,6 +521,11 @@ export class Repository extends Construct implements IRepository {
   private static REPOSITORY_OWNER = { uid: 1000, gid: 1000 };
 
   /**
+   * Default username for auto generated admin credentials in Secret Manager.
+   */
+  private static DEFAULT_SECRETS_MANAGEMENT_USERNAME: string = 'admin';
+
+  /**
    * @inheritdoc
    */
   public readonly rootPrefix: string;
@@ -552,7 +559,7 @@ export class Repository extends Construct implements IRepository {
   /**
    * Deadline Secrets Management settings.
    */
-  public readonly secretsManagementSettings: SecretsManagementProps
+  public readonly secretsManagementSettings: SecretsManagementProps;
 
   constructor(scope: Construct, id: string, props: RepositoryProps) {
     super(scope, id);
@@ -576,15 +583,15 @@ export class Repository extends Construct implements IRepository {
       enabled: props.secretsManagementSettings?.enabled ?? true,
       credentials: props.secretsManagementSettings?.credentials ??
         ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(this, 'SMAdminUser', {
-          description: 'Admin credentials for Secret Management',
+          description: 'Admin credentials for Deadline Secrets Management',
           generateSecretString: {
-            excludeCharacters: '\"$&\'()-/<>[\\]\`{|}',
+            excludeCharacters: '\"$&\'()/<>[\\]\`{|}',
             includeSpace: false,
             passwordLength: 24,
             requireEachIncludedType: true,
 
             generateStringKey: 'password',
-            secretStringTemplate: JSON.stringify({ username: 'admin' }),
+            secretStringTemplate: JSON.stringify({ username: Repository.DEFAULT_SECRETS_MANAGEMENT_USERNAME }),
           },
         }) : undefined),
     };
@@ -999,8 +1006,8 @@ export class Repository extends Construct implements IRepository {
 
     if (this.secretsManagementSettings.enabled) {
       installerArgs.push('-r', Stack.of(this.secretsManagementSettings.credentials ?? this).region);
-      this.secretsManagementSettings.credentials?.grantRead(installerGroup);
-      installerArgs.push('-c', this.secretsManagementSettings.credentials?.secretArn ?? '');
+      this.secretsManagementSettings.credentials!.grantRead(installerGroup);
+      installerArgs.push('-c', this.secretsManagementSettings.credentials!.secretArn ?? '');
     }
 
     if (settings) {
--- a/packages/aws-rfdk/lib/deadline/lib/repository.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/repository.ts
@@ -46,6 +46,10 @@ import {
   Asset,
 } from '@aws-cdk/aws-s3-assets';
 import {
+  ISecret,
+  Secret,
+} from '@aws-cdk/aws-secretsmanager';
+import {
   Annotations,
   Construct,
   Duration,
@@ -272,6 +276,34 @@ export interface RepositorySecurityGroupsOptions {
 }
 
 /**
+ * Settings used by Deadline Secrets Management, a feature introduced in Deadline 10.1.10 for securely managing storage
+ * and access of Secrets for your render farm.
+ * More details at:
+ * https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html
+ * Using Secrets Management requires TLS to be enabled between the RenderQueue and its clients. If this feature is enabled, the
+ * `externalTLS` on the `RenderQueueTrafficEncryptionProps` interface on the RenderQueue cannot be disabled.
+ */
+export interface SecretsManagementProps {
+  /**
+   * Whether or not to enable the Secrets Management feature.
+   * @default true
+   */
+  readonly enabled?: boolean;
+  /**
+   * A Secret containing the username and password to use for the admin role.
+   * The contents of this secret must be a JSON document with the keys "username" and "password". ex:
+   *     {
+   *         "username": <admin user name>,
+   *         "password": <admin user password>,
+   *     }
+   * Password should contain at least one lowercase letter, one uppercase letter, one symbol and one number.
+   *
+   * @default: A random username and password will be generated in a Secret with ID `SMAdminUser` and will need to be retrieved from AWS Secrets Manager if it is needed
+   */
+  readonly credentials?: ISecret;
+}
+
+/**
  * Properties for the Deadline repository
  */
 export interface RepositoryProps {
@@ -387,6 +419,15 @@ export interface RepositoryProps {
    * @default Repository settings are not imported.
    */
   readonly repositorySettings?: Asset;
+
+  /**
+   * Define the settings used by Deadline Secrets Management, a feature introduced in Deadline 10.1.10 for securely managing storage
+   * and access of Secrets for your render farm.
+   * More details at:
+   * https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html
+   * @default: Secrets Management will be enabled and a username and password will be automatically generated if none are supplied.
+   */
+  readonly secretsManagementSettings?: SecretsManagementProps
 }
 
 /**
@@ -508,6 +549,11 @@ export class Repository extends Construct implements IRepository {
    */
   private readonly installerGroup: AutoScalingGroup;
 
+  /**
+   * Deadline Secrets Management settings.
+   */
+  public readonly secretsManagementSettings: SecretsManagementProps
+
   constructor(scope: Construct, id: string, props: RepositoryProps) {
     super(scope, id);
 
@@ -525,6 +571,23 @@ export class Repository extends Construct implements IRepository {
     }
 
     this.version = props.version;
+
+    this.secretsManagementSettings = {
+      enabled: props.secretsManagementSettings?.enabled ?? true,
+      credentials: props.secretsManagementSettings?.credentials ??
+        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(this, 'SMAdminUser', {
+          description: 'Admin credentials for Secret Management',
+          generateSecretString: {
+            excludeCharacters: '\"$&\'()-/<>[\\]\`{|}',
+            includeSpace: false,
+            passwordLength: 24,
+            requireEachIncludedType: true,
+
+            generateStringKey: 'password',
+            secretStringTemplate: JSON.stringify({ username: 'admin' }),
+          },
+        }) : undefined),
+    };
 
     this.fileSystem = props.fileSystem ?? (() => {
       const fs = new EfsFileSystem(this, 'FileSystem', {
@@ -933,6 +996,12 @@ export class Repository extends Construct implements IRepository {
       '-p', `"${installPath}"`,
       '-v', version.linuxFullVersionString(),
     ];
+
+    if (this.secretsManagementSettings.enabled) {
+      installerArgs.push('-r', Stack.of(this.secretsManagementSettings.credentials ?? this).region);
+      this.secretsManagementSettings.credentials?.grantRead(installerGroup);
+      installerArgs.push('-c', this.secretsManagementSettings.credentials?.secretArn ?? '');
+    }
 
     if (settings) {
       const repositorySettingsFilePath = installerGroup.userData.addS3DownloadCommand({
--- a/packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
+++ b/packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
@@ -18,6 +18,8 @@ Required arguments:
 Optional arguments
   -s Deadline Repository settings file to import.
   -o The UID[:GID] that this script will chown the Repository files for. If GID is not specified, it defults to be the same as UID."
+  -c Secret management admin credentials ARN. If this parameter is specified, secrets management will be enabled.
+  -r Region where stacks are deployed. Required to get secret management credentials.
 
 while getopts "i:p:v:s:o:" opt; do
   case $opt in
@@ -30,6 +32,10 @@ while getopts "i:p:v:s:o:" opt; do
     s) DEADLINE_REPOSITORY_SETTINGS_FILE="$OPTARG"
     ;;
     o) DEADLINE_REPOSITORY_OWNER="$OPTARG"
+    ;;
+    c) SECRET_MANAGEMENT_ARN="$OPTARG"
+    ;;
+    r) AWS_REGION="$OPTARG"
     ;;
     /?)
       echo "$USAGE"
@@ -109,6 +115,39 @@ if [ ! -z "${DEADLINE_REPOSITORY_SETTINGS_FILE+x}" ]; then
   fi
 fi
 
+SECRET_MANAGEMENT_ARG=''
+if [ ! -z "${SECRET_MANAGEMENT_ARN+x}" ]; then
+  sudo yum install -y jq
+
+  SM_SECRET_VALUE=$(aws secretsmanager get-secret-value --secret-id=$SECRET_MANAGEMENT_ARN --region=$AWS_REGION)
+  SM_SECRET_STRING=$(jq -r '.SecretString' <<< "$SM_SECRET_VALUE")
+  SECRET_MANAGEMENT_USER=$(jq -r '.username' <<< "$SM_SECRET_STRING")
+  SECRET_MANAGEMENT_PASSWORD=$(jq -r '.password' <<< "$SM_SECRET_STRING")
+
+  len=$(echo ${#SECRET_MANAGEMENT_PASSWORD})
+  if test $len -ge 8 ; then
+      echo "$SECRET_MANAGEMENT_PASSWORD" | grep -q [0-9]
+      if test $? -eq 0 ; then
+          echo "$SECRET_MANAGEMENT_PASSWORD" | grep -q [A-Z]
+          if test $? -eq 0 ; then
+              echo "$SECRET_MANAGEMENT_PASSWORD" | grep -q [a-z]
+              if test $? -eq 0 ; then
+                  echo "$SECRET_MANAGEMENT_PASSWORD" | grep -q [~,.,:,@,!,\#,%,*,_,+,-,=,?]
+                  if test $? -eq 0 ; then
+                      SM_STRONG_PASSWORD='true'
+                  fi  
+              fi
+          fi
+      fi
+  fi
+  if [ -z "${SM_STRONG_PASSWORD+x}" ]; then
+    echo "ERROR: Admin password is too weak. It must be at least 8 characters long and contain at least one lowercase letter, one uppercase letter, one symbol and one digit."
+    exit 1
+  fi
+  echo "Secret management is enabled. Credentials are stored in secret: $SECRET_MANAGEMENT_ARN"
+  SECRET_MANAGEMENT_ARG="--installSecretsManagement true --secretsAdminName \"$SECRET_MANAGEMENT_USER\" --secretsAdminPassword \"$SECRET_MANAGEMENT_PASSWORD\""
+fi
+
 if [[ -n "${DEADLINE_REPOSITORY_OWNER+x}" ]]; then
   if [[ ! "$DEADLINE_REPOSITORY_OWNER" =~ ^[0-9]+(:[0-9]+)?$ ]]; then
     echo "ERROR: Deadline Repository owner is invalid: ${DEADLINE_REPOSITORY_OWNER}"
@@ -138,7 +177,7 @@ if [[ -n "${DEADLINE_REPOSITORY_OWNER+x}" ]]; then
   fi
 fi
 
-$REPO_INSTALLER --mode unattended --setpermissions false --prefix "$PREFIX" --installmongodb false --backuprepo false ${INSTALLER_DB_ARGS_STRING} $REPOSITORY_SETTINGS_ARG_STRING
+$REPO_INSTALLER --mode unattended --setpermissions false --prefix "$PREFIX" --installmongodb false --backuprepo false ${INSTALLER_DB_ARGS_STRING} $REPOSITORY_SETTINGS_ARG_STRING $SECRET_MANAGEMENT_ARG
 
 if [[ -n "${REPOSITORY_OWNER_UID+x}" ]]; then
   echo "Changing ownership of Deadline Repository files to UID=$REPOSITORY_OWNER_UID GID=$REPOSITORY_OWNER_GID"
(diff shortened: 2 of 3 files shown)


packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -8,15 +8,25 @@
...
 PREFIX=$2
 DEADLINE_REPOSITORY_VERSION=$3
-DEADLINE_REPOSITORY_SETTINGS_FILE=${4:-}
+SECRET_MANAGEMENT_ENABLED=${4:-"false"}
+if [ "$SECRET_MANAGEMENT_ENABLED" == "true" ]; then
+    AWS_REGION=${5}
+    SECRET_MANAGEMENT_ARN=${6}
+    DEADLINE_REPOSITORY_SETTINGS_FILE=${7:-}
+else
+    DEADLINE_REPOSITORY_SETTINGS_FILE=$SECRET_MANAGEMENT_ENABLED
```
horsmand: It looks like this will get set to `false` due to how we're setting the default in `SECRET_MANAGEMENT_ENABLED=${4:-"false"}`

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -271,6 +275,27 @@ export interface RepositorySecurityGroupsOptions {
...
+  /**
+   * Whether or not to enable the Secrets Management feature.
+   * @default true
+   */
+  readonly enabled?: boolean;
+  /**
+   * A Secret containing the username and password to use for the admin role
+   * @default: A random username and password will be generated in a Secret and will need to be retrieved from AWS Secrets Manager if it is needed
+   */
+  readonly credentials?: ISecret;
```
horsmand: We should expand this comment to include a few more things:
1. We should show that it expects a JSON object with `username` and `password` fields.
2. We should add the password requirements, I can't remember them specifically but I think it might require a number and symbol, maybe a capital letter too?
3. We should give better directions on how to find the auto-generated Secret containing the credentials. The simplest way to do that might be to just tell them what the name of it will be so they can look it up in the AWS Console. Another option could be pointing customers to [--outputs-file/outputsFile](https://github.com/aws/aws-cdk/tree/master/packages/aws-cdk#outputs) so they can get the ARN from the generated file.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -519,6 +558,23 @@ export class Repository extends Construct implements IRepository {
...
+          generateSecretString: {
+            excludeCharacters: '\"$&\'()-/<>[\\]\`{|}',
+            includeSpace: false,
+            passwordLength: 24,
+            requireEachIncludedType: true,
+
+            generateStringKey: 'password',
+            secretStringTemplate: JSON.stringify({ username: 'admin' }),
+          },
+        }) : undefined),
```
horsmand: I feel like auto-generating credentials for our customers might be a bit confusing. I think we should refactor this code so that if we do end up generating the credentials, we print a message containing the ARN of the Secret to the command line to increase visibility.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -501,6 +535,11 @@ export class Repository extends Construct implements IRepository {
    */
   private readonly installerGroup: AutoScalingGroup;
 
+  /**
+   * Deadline Secrets Management settings.
+   */
+  public readonly secretsManagementSettings: SecretsManagementProps
```
horsmand: I don't think there's any use case where Secrets Management is enabled but the credentials are unset, or the credentials are set but Secrets Management is disabled, so we can simplify this to be an optional field that holds the credentials and we can infer if SM is enabled based on if it's set.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -923,6 +979,13 @@ export class Repository extends Construct implements IRepository {
       version.linuxFullVersionString(),
     ];
 
+    if (this.secretsManagementSettings.enabled) {
+      installerArgs.push('true');
+      installerArgs.push(Stack.of(this).region);
```
horsmand: Will this work if we do `Stack.of(this.secretsManagementCredentials).region`? It would protect us if for some reason the Secret the customer provided was in a different region than this Stack.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -1224,3 +1225,45 @@ test('imports repository settings', () => {
...
+    vpc,
+    version,
+    secretsManagementSettings: {
+      enabled: false,
+    },
+  });
+
+  // THEN
+  expect(repository.secretsManagementSettings.credentials).toBeUndefined();
+});
```
horsmand: Missing blank line.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -1224,3 +1225,45 @@ test('imports repository settings', () => {
   const installerGroup = repository.node.tryFindChild('Installer') as AutoScalingGroup;
   expect(installerGroup.userData.render()).toContain(`aws s3 cp '${repositorySettings.s3ObjectUrl}'`);
 });
+
+test('secret manager enabled', () => {
```
horsmand, line 1265: We should have a test case where we don't set the `enabled` prop or the `credentials` prop and we confirm that Secrets Management is enabled and a credentials Secret gets generated.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -519,6 +558,23 @@ export class Repository extends Construct implements IRepository {
...
+        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(this, 'SMAdminUser', {
+          description: 'Admin credentials for Secret Management',
+          generateSecretString: {
+            excludeCharacters: '\"$&\'()-/<>[\\]\`{|}',
+            includeSpace: false,
+            passwordLength: 24,
+            requireEachIncludedType: true,
+
+            generateStringKey: 'password',
+            secretStringTemplate: JSON.stringify({ username: 'admin' }),
```
horsmand: We can pull `admin` out into a `DEFAULT_SECRETS_MANAGEMENT_USERNAME` constant.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -1224,3 +1225,45 @@ test('imports repository settings', () => {
...
+    vpc,
+    version,
+    secretsManagementSettings: {
+      enabled: true,
+      credentials: expectedCredentials,
+    },
+  });
+
+  // THEN
+  expect(repository.secretsManagementSettings.credentials).toBe(expectedCredentials);
```
horsmand, line 1290: We should have another expectation on the User Data to make sure it includes the line to retrieve the credentials. Something like:
```
const installerGroup = repository.node.tryFindChild('Installer') as AutoScalingGroup;
expect(installerGroup.userData.render())
  .toContain(`aws secretsmanager get-secret-value --secret-id=${repository.secretsManagementSettings.credentials.secretArn}'`);
```

packages/aws-rfdk/lib/deadline/lib/repository.ts
kozlove-aws: I can imagine when user want to set Secrets Management is enabled (just to make sure) but expect secret is created by RFDK. Anyway such structure looks clear for understanding.

packages/aws-rfdk/lib/deadline/lib/repository.ts
kozlove-aws: It is our normal practice in RFDK to set username for admin credentials `admin` by default. User can always setup own username and password in custom secret.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
kozlove-aws: I believe this script is running not in UserData and I do not know a way to check script that is running on the instance.

packages/aws-rfdk/lib/deadline/lib/repository.ts
horsmand: We should definitely have it on the SecretsManagementProps interface, but storing the whole interface on the class instance rather than just the credentials isn't necessary. You can leave the interface as-is but change this to just store the `credentials` optional.

If I want to tell if Secrets Management is enabled I could just check if `credentials` is set rather than first checking the value of `enabled` and then checking the value of `credentials`. Also, in this context, enabled isn't optional, it must be set, but on the interface it's optional, so it isn't really what we need here.

packages/aws-rfdk/lib/deadline/lib/repository.ts
horsmand: Yes that's fine to set it to `admin` by default, I just mean to pull it into a constant and refer to that here rather than hardcoding the value.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
horsmand: I'm pretty sure it does get added to the User Data of the repository installer instance. The code you added went into `installDeadlineRepository.sh`. You can see the asset get created as a `ScriptAsset` [here](https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/deadline/lib/repository.ts#L917) and then added to the User Data [here](https://github.com/aws/aws-rfdk/blob/mainline/packages/aws-rfdk/lib/deadline/lib/repository.ts#L951) using the [executeOn](https://github.com/aws/aws-rfdk/blob/984201ca356c83c8e476ec5028c252f772c442bd/packages/aws-rfdk/lib/core/lib/script-assets.ts#L178) method from out `ScriptAsset` class.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -271,6 +275,34 @@ export interface RepositorySecurityGroupsOptions {
...
+   */
+  readonly enabled?: boolean;
+  /**
+   * A Secret containing the username and password to use for the admin role.
+   * The contents of this secret must be a JSON document with the keys "username" and "password". ex:
+   *     {
+   *         "username": <admin user name>,
+   *         "password": <admin user password>,
+   *     }
+   * Password should contain at least one lowercase letter, one uppercase letter, one symbol and one number.
```
horsmand: I just realized we have no easy way to do validation on the password that the user supplies. This means that if they supply a bad password, we won't find out until the CDK deployment gets to the point that the repository installer is running. This isn't ideal, but I don't see a better way to do it.

We could have a Custom Resource that runs a Lambda to do validation on it, but that also would happen at deploy time. At least if we did that, it would be more obvious to the user why the deployment failed. As it stands now, if the deployment fails, they would have to go check the logs in CloudWatch.

Maybe we should add a note here to say something like:
If the password requirements aren't met, the repository installer will fail during the deployment. We recommend leaving this blank so that a random password will be generated that meets the requirements.

horsmand (changes requested):

packages/aws-rfdk/lib/deadline/lib/repository.ts
jericht: I agree with @horsmand here. Exposing this as `SecretsManagementProps` could be confusing to work with since there are some combination of values that won't make sense, for example:
```ts
{
  enabled: true,
  credentials: undefined
}
```
Secrets Management can't be enabled if the credentials don't exist in SM. I think just exposing the `ISecret` with the SM credentials as an optional prop makes more sense here.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -923,6 +986,13 @@ export class Repository extends Construct implements IRepository {
       version.linuxFullVersionString(),
     ];
 
+    if (this.secretsManagementSettings.enabled) {
+      installerArgs.push('true');
```
jericht: This will need to be updated after a rebase. The installer script has been changed to use named parameters instead of positional parameters

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -82,7 +92,18 @@ if [ ! -z "$DEADLINE_REPOSITORY_SETTINGS_FILE" ]; then
...
-$REPO_INSTALLER --mode unattended --setpermissions false --prefix "$PREFIX" --installmongodb false --backuprepo false ${INSTALLER_DB_ARGS_STRING} $REPOSITORY_SETTINGS_ARG_STRING
+SECRET_MANAGEMENT_ARG=''
+if [ "$SECRET_MANAGEMENT_ENABLED" == "true" ]; then
+  sudo yum install -y jq
+  echo "Secret management is enabled. Credentials are stored in secret: $SECRET_MANAGEMENT_ARN"
+  SM_SECRET_VALUE=$(aws secretsmanager get-secret-value --secret-id=$SECRET_MANAGEMENT_ARN --region=$AWS_REGION)
+  SM_SECRET_STRING=$(jq -r '.SecretString' <<< "$SM_SECRET_VALUE")
+  SECRET_MANAGEMENT_USER=$(jq -r '.username' <<< "$SM_SECRET_STRING")
+  SECRET_MANAGEMENT_PASSWORD=$(jq -r '.password' <<< "$SM_SECRET_STRING")
+  SECRET_MANAGEMENT_ARG="--installSecretsManagement true --secretsAdminName $SECRET_MANAGEMENT_USER --secretsAdminPassword $SECRET_MANAGEMENT_PASSWORD"
```
jericht:
```suggestion
  SECRET_MANAGEMENT_ARG="--installSecretsManagement true --secretsAdminName \"$SECRET_MANAGEMENT_USER\" --secretsAdminPassword \"$SECRET_MANAGEMENT_PASSWORD\""
```

I can't find any restrictions on not having whitespace in the name or password, so I'm assuming there could be whitespace there. We should quote them if that's the case

packages/aws-rfdk/lib/deadline/test/repository.test.ts
jericht: I think we might be a bit confused about what's going on here. AFAIK, the changes to use the AWS CLI to grab the secret from SM happens in the `installDeadlineRepository.sh` script. The UserData downloads, unzips, and runs that script with the parameters we provide, but it does so without knowing the contents of the script itself, so I don't think we can actually assert the AWS CLI call here.

However, I think we _can_ assert that the new SM-related parameters are being passed to the `installDeadlineRepository.sh` script, which would be (roughly) in the following format in UserData:
```
<path-to-installDeadlineRepository.sh> ... <sm-enabled> <region> <sm-admin-creds-secret-arn> ...
```

Note that the above will change once this PR is rebased since the repo installer script uses named parameters instead of positional parameters

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -1224,3 +1225,57 @@ test('imports repository settings', () => {
...
+    vpc,
+    version,
+  });
+
+  // THEN
+  expect(repository.secretsManagementSettings.enabled).toBeTruthy();
+  expect(repository.secretsManagementSettings.credentials).toBeDefined();
+});
+
+test('secret manager disabled', () => {
```
jericht:
```suggestion
test('credentials are undefined when secrets management is disabled', () => {
```

jericht (changes requested):

packages/aws-rfdk/lib/deadline/lib/repository.ts
horsmand: I just noticed that the password must also be at least 8 characters long. We should add that to the requirement.

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -82,7 +92,50 @@ if [ ! -z "$DEADLINE_REPOSITORY_SETTINGS_FILE" ]; then
...
+  if test $len -ge 8 ; then
+      echo "$SECRET_MANAGEMENT_PASSWORD" | grep -q [0-9]
+      if test $? -eq 0 ; then
+          echo "$SECRET_MANAGEMENT_PASSWORD" | grep -q [A-Z]
+          if test $? -eq 0 ; then
+              echo "$SECRET_MANAGEMENT_PASSWORD" | grep -q [a-z]
+              if test $? -eq 0 ; then
+                  echo "$SECRET_MANAGEMENT_PASSWORD" | grep -q [~,.,:,@,!,\#,%,*,_,+,-,=,?]
+                  if test $? -eq 0 ; then
+                      echo "Admin password is strong enaugh"
```
horsmand:
```suggestion
                      echo "Admin password is strong enough"
```

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -82,7 +92,50 @@ if [ ! -z "$DEADLINE_REPOSITORY_SETTINGS_FILE" ]; then
...
+                  else
+                      echo "Admin password is weak. It should include at least one  special symbols"
```
horsmand:
```suggestion
                      echo "ERROR: Admin password is too weak. It should include at least one special symbol"
```

horsmand: The `ERROR` should be on all the failure messages to make it clearer that this is why the script failed.

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -82,7 +92,50 @@ if [ ! -z "$DEADLINE_REPOSITORY_SETTINGS_FILE" ]; then
...
+      else
+        echo "Admin password is weak. It should include at least one number"
+        exit 1
+      fi
+  else
+      echo "Admin password is weak. It should be at least 8 characters long"
+      exit 1
+  fi
+  echo "Secret management is enabled. Credentials are stored in secret: $SECRET_MANAGEMENT_ARN"
+fi
```
horsmand: Can we collapse this all down into one test with one failure message? Something like:
```
ERROR: Admin password is too weak. It must be at least 8 characters long and contain at least one lowercase letter, one uppercase letter, one symbol and one digit.
```
 It's nice to be more specific, but if we're going to do it that way, we should run all the tests, collect all the failure messages, and then only fail after we've told them all the problems.
For instance, right now, if the user used the password `aaaa` they would only get the message `Admin password is weak. It should be at least 8 characters long` but really we'd want to make sure they know they also didn't include a capital letter, a symbol, and a digit.

horsmand (changes requested):

packages/aws-rfdk/lib/deadline/lib/repository.ts
horsmand: Actually it is a valid case for enabled to be true and the credentials to be undefined, we will generate them and save them in Secrets Manager in that case. The only case that wouldn't be defined is if enabled was `false` but credentials were supplied. I don't really see that case as being a problem, the expected behaviour there would be that SM is disabled and the credentials are ignored.

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -95,12 +95,44 @@ fi
...
-  echo "Secret management is enabled. Credentials are stored in secret: $SECRET_MANAGEMENT_ARN"
+  
   SM_SECRET_VALUE=$(aws secretsmanager get-secret-value --secret-id=$SECRET_MANAGEMENT_ARN --region=$AWS_REGION)
   SM_SECRET_STRING=$(jq -r '.SecretString' <<< "$SM_SECRET_VALUE")
   SECRET_MANAGEMENT_USER=$(jq -r '.username' <<< "$SM_SECRET_STRING")
   SECRET_MANAGEMENT_PASSWORD=$(jq -r '.password' <<< "$SM_SECRET_STRING")
-  SECRET_MANAGEMENT_ARG="--installSecretsManagement true --secretsAdminName $SECRET_MANAGEMENT_USER --secretsAdminPassword $SECRET_MANAGEMENT_PASSWORD"
+  SECRET_MANAGEMENT_ARG="--installSecretsManagement true --secretsAdminName \"$SECRET_MANAGEMENT_USER\" --secretsAdminPassword \"$SECRET_MANAGEMENT_PASSWORD\""
+  len=$(echo ${#SECRET_MANAGEMENT_PASSWORD})
+  if test $len -ge 8 ; then
```
jericht: We could change this to one if statement where all the conditions are compared with logical OR and then have a single error message stating all the password requirements when the password is not strong enough. I think this will make the code look a little cleaner, but it's not necessary though, just a suggestion.

jericht (changes requested):

packages/aws-rfdk/lib/deadline/lib/repository.ts
kozlove-aws: For my opinion it is much clear and understandable to have separate parameters to enable SM and credentials.
Also this interface can be extended by others parameters in the future.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -387,6 +421,15 @@ export interface RepositoryProps {
...
   readonly repositorySettings?: Asset;
+
+  /**
+   * Define the settings used by Deadline Secrets Management, a feature introduced in Deadline 10.1.10 for securely managing storage
+   * and access of Secrets for your render farm.
+   * More details at:
+   * https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html
+   * @default: Secrets Management will be enabled and a username and password will be automatically generated if none are supplied.
+   */
+  readonly secretsManagementSettings?: SecretsManagementProps
```
horsmand:
```suggestion
  readonly secretsManagementSettings?: SecretsManagementProps;
```
Does our linting during the build not catch that? Seems like it should

horsmand: Also missing on the class variable on line 555.

packages/aws-rfdk/lib/deadline/lib/repository.ts
horsmand: My personal preference would be two interfaces, one for the construct properties and one for the class variable, but they would be so close to the same thing that I feel it's a bit nit-picky and not worth the effort. I'd say just leave the interface and the class variable as-is.

For reference, I was thinking we would have these interfaces:
```
export interface SecretsManagementProps {
  readonly enabled?: boolean;
  readonly credentials?: ISecret;
}
export interface SecretsManagementSettings {
  readonly enabled: boolean;
  readonly credentials?: ISecret;
}
```
then their usage would be:
```
export interface RepositoryProps {
  readonly secretsManagementSettings?: SecretsManagementProps;
  ...
}

export class Repository extends Construct implements IRepository {
  public readonly secretsManagementSettings: SecretsManagementSettings;

  constructor(scope: Construct, id: string, props: RepositoryProps) {
      this.secretsManagementSettings = {
      enabled: props.secretsManagementSettings?.enabled ?? true,
      credentials: props.secretsManagementSettings?.credentials ??
        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(this, 'SMAdminUser', {
          description: 'Admin credentials for Secret Management',
          generateSecretString: {
            excludeCharacters: '\"$&\'()/<>[\\]\`{|}',
            includeSpace: false,
            passwordLength: 24,
            requireEachIncludedType: true,

            generateStringKey: 'password',
            secretStringTemplate: JSON.stringify({ username: Repository.DEFAULT_SECRETS_MANAGEMENT_USERNAME }),
          },
        }) : undefined),
    };
    ...
  }
```
As you can see, the only difference is that `enabled` isn't optional on the class variable. The logic in the `constructor` is the exact same as it currently is in this PR.

kozlove-aws: Hm. Actually I can't see any reason to make `enabled` property optional.
If someone decided to define `secretsManagementSettings` what the reason do not set `enabled`?
What do you say if I just make this property mandatory for both cases?

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
jericht: Thanks, this looks better. I think we can make it even simpler though:

```bash
if [[ ${#SECRET_MANAGEMENT_PASSWORD} -lt 8 ]] || ! echo $1 | grep [0-9] | grep [a-z] | grep [A-Z] | grep -q [^[:alnum:]]; then
    echo "ERROR: Admin password is too weak. It must be at least 8 characters long and contain at least one lowercase letter, one uppercase letter, one symbol and one digit."
    exit 1
fi
```

I think shrinking it down to a few lines like this will make it easier to read, what do you think?

Edit: This solution assumes there are no newlines in the password.. let me think of another way to do this

jericht: Thanks, I think we can make this shorter though to make the code even easier to read:

```bash
if !([[ ${#SECRET_MANAGEMENT_PASSWORD} -ge 8 ]] && 
  echo $SECRET_MANAGEMENT_PASSWORD | grep -q [0-9] &&
  echo $SECRET_MANAGEMENT_PASSWORD | grep -q [a-z] &&
  echo $SECRET_MANAGEMENT_PASSWORD | grep -q [A-Z] &&
  echo $SECRET_MANAGEMENT_PASSWORD | grep -q [^[:alnum:]])
then
    echo "ERROR: Admin password is too weak. It must be at least 8 characters long and contain at least one lowercase letter, one uppercase letter, one symbol and one digit."
    exit 1
fi
```

This should be good enough, it at least gets rid of the nested if statements while still working with newlines unlike my previous suggestion. What do you think?

kozlove-aws: I like it. looks much better.

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -18,6 +18,8 @@ Required arguments:
 Optional arguments
   -s Deadline Repository settings file to import.
   -o The UID[:GID] that this script will chown the Repository files for. If GID is not specified, it defults to be the same as UID."
+  -c Secret management admin credentials ARN. If this parameter is specified, secrets management will be enabled.
+  -r Region where stacks are deployed. Required to get secret management credentials.
 
 while getopts "i:p:v:s:o:" opt; do
```
jericht:
```suggestion
while getopts "i:p:v:s:o:c:r:" opt; do
```

I missed this last time. This needs to be updated for these args to be captured

jericht (changes requested):

jericht (approved):
Looks good, thanks!

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -271,6 +275,36 @@ export interface RepositorySecurityGroupsOptions {
...
+
+  /**
+   * A Secret containing the username and password to use for the admin role.
+   * The contents of this secret must be a JSON document with the keys "username" and "password". ex:
+   *     {
+   *         "username": <admin user name>,
+   *         "password": <admin user password>,
+   *     }
+   * Password should be at least 8 characters long and contain at least one lowercase letter, one uppercase letter, one symbol and one number.
+   * In case when password does not meet requirements secret management will be disabled.
```
horsmand:
```suggestion
   * In the case when the password does not meet the requirements, the repository construct will fail to deploy.
```

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -526,6 +579,23 @@ export class Repository extends Construct implements IRepository {
 
     this.version = props.version;
 
+    this.secretsManagementSettings = {
+      enabled: props.secretsManagementSettings?.enabled ?? true,
+      credentials: props.secretsManagementSettings?.credentials ??
+        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(this, 'SMAdminUser', {
+          description: 'Admin credentials for Secret Management',
```
horsmand:
```suggestion
          description: 'Admin credentials for Deadline Secrets Management',
```
We should be clear this is for Deadline. I could see it being confusing otherwise.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -934,6 +1004,12 @@ export class Repository extends Construct implements IRepository {
       '-v', version.linuxFullVersionString(),
     ];
 
+    if (this.secretsManagementSettings.enabled) {
+      installerArgs.push('-r', Stack.of(this.secretsManagementSettings.credentials ?? this).region);
+      this.secretsManagementSettings.credentials?.grantRead(installerGroup);
+      installerArgs.push('-c', this.secretsManagementSettings.credentials?.secretArn ?? '');
```
horsmand:
```suggestion
      this.secretsManagementSettings.credentials!.grantRead(installerGroup);
      installerArgs.push('-c', this.secretsManagementSettings.credentials!.secretArn ?? '');
```
At this point in the code we should be certain that the credentials are populated.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -1260,3 +1261,59 @@ test('IMountableLinuxFilesystem.usesUserPosixPermissions() = false does not chan
...
+    vpc,
+    version,
+    secretsManagementSettings: {
+      enabled: false,
+    },
+  });
+
+  // THEN
+  expect(repository.secretsManagementSettings.credentials).toBeUndefined();
+});
```
horsmand: Missing a blank line.

horsmand (changes requested):
Really small stuff that should be quick to fix and not require re-testing.

horsmand (approved):

PR #515, opened 2021-07-30
fix(examples): change EFS access point for repository to use root

jericht:
### Notes
The [PR](https://github.com/aws/aws-rfdk/pull/461) that added integration with FSx for Lustre also added a step in the `Repository` installer script to `chown` the repository files. This broke the basic example app because it uses an EFS access point with a POSIX user that isn't root, causing the `chown` operations to fail.

This PR addresses this by changing the EFS access point in the basic example app to give access as root, with some added comments about using a different POSIX user when mounting to anything other than the Repository.

### Testing
- Deployed the Typescript app and verified the Repository installer script succeeded
- Deployed the Python app and verified the Repository installer script succeeded

----


jericht:
Closing this as we are taking another approach.

PR #518, opened 2021-08-04
fix(deadline): change ownership of repository files when applicable

jusiskin:
BREAKING CHANGE: MountableEfs will not work with the Repository construct when
created with an imported EFS Access Point

## Problem

In #461, the `Repository` construct's user-data script was modified to run `chown` to change the ownership of the Deadline Repository files. When [`MountableEfs`](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.MountableEfs.html) is used with an EFS Access Point that fixes the POSIX user, the `chown` operation fails.

This is a breaking change for any users that have adopted what we demonstrate in the `All-In-AWS-Infrastructure-Basic` example app:

https://github.com/aws/aws-rfdk/blob/b83c0e0823ef94e9df44112e194bc1bb2bf9e25a/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/storage-tier.ts#L105-L139

## Solution

Added a new method to the `IMountableLinuxFilesystem` interface (TypeScript: `.usesUserPosixPermissions()`) Python: `.uses_user_posix_permissions()`) shared by `MountableEfs` / `MountableFsxLustre` / `MountableBlockVolume` to be able to query at synthesis-time whether the mounted file-system uses the POSIX user permission of the host system. In the case described in the problem, `MountableEfs` returns `false` when an EFS access point is provided with a fixed POSIX user. All other `Mountable*` classes return `true`.

---

**NOTE:** EFS Access Points do not require a POSIX user to be specified and there is no way to determine whether an imported EFS Access Point (e.g. [`AccessPoint.fromAccessPointId(...)`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-efs.AccessPoint.html#static-fromwbraccesswbrpointwbridscope-id-accesspointid)) has a POSIX user. For this reason, the imported access points will throw an exception when calling `.usesUserPosixPermissions()`. Otherwise, if we add the `chown` command with an access point, deployment will fail. If we omit the `chown` without the access point, the `RenderQueue` will not have file-system permissions to the Deadline Repository files.

This is why we are considering this a breaking change, since prior to the introduction of the `chown` operation, imported EFS Access Points were supported.

---

The `Repository` construct now uses this method to determine whether it needs to change the ownership of the Deadline Repository files after the install completes.

## Testing

1.  Deployed the `All-In-AWS-Infrastructure-Basic` example app in this repository in its current form (EFS with an access point fixing the POSIX user) and verified that:
    1.  the deployment (and repository installation) succeeds.
    2.  connected a Monitor to the farm, and submitted a ping job and confirmed the render completes and task logs can be viewed (verifies repository file-system permissions are correct for the `RenderQueue`)
    3.  from the bastion, inspected the mounted EFS file-system and confirmed the Deadline Repository files are owned by UID/GID of `10000:10000` as specified by the access point.
1.  Modified the `All-In-AWS-Infrastructure-Basic` example app to use an FSx Lustre file-system for the Deadline Repository and verified that:
    1.  the deployment (and repository installation) succeeds.
    2.  connected a Monitor to the farm, and submitted a ping job and confirmed the render completes and task logs can be viewed (verifies repository file-system permissions are correct for the `RenderQueue`)
    3.  from the bastion, inspected the mounted FSx file-system and confirmed the Deadline Repository files are owned by UID/GID of `1000:1000` as the `Repository` construct is expected to `chown` the files for proper POSIX access by the `RenderQueue`.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/test/repository.test.ts
+++ b/packages/aws-rfdk/lib/deadline/test/repository.test.ts
@@ -1225,11 +1225,15 @@ test('imports repository settings', () => {
   expect(installerGroup.userData.render()).toContain(`aws s3 cp '${repositorySettings.s3ObjectUrl}'`);
 });
 
-test('changes ownership of repository files', () => {
+test('IFileSystem.usesUserPosixPermissions() = true changes ownership of repository files', () => {
   // GIVEN
   const repo = new Repository(stack, 'Repository', {
     version,
     vpc,
+    fileSystem: {
+      mountToLinuxInstance: (_target, _mount) => {},
+      usesUserPosixPermissions: () => true,
+    },
   });
 
   // WHEN
@@ -1237,4 +1241,22 @@ test('changes ownership of repository files', () => {
 
   // THEN
   expect(script).toMatch('-o 1000:1000');
+});
+
+test('IFileSystem.usesUserPosixPermissions() = false does not change ownership of repository files', () => {
+  // GIVEN
+  const repo = new Repository(stack, 'Repository', {
+    version,
+    vpc,
+    fileSystem: {
+      mountToLinuxInstance: (_target, _mount) => {},
+      usesUserPosixPermissions: () => false,
+    },
+  });
+
+  // WHEN
+  const script = (repo.node.defaultChild as AutoScalingGroup).userData.render();
+
+  // THEN
+  expect(script).not.toMatch('-o 1000:1000');
 });
--- a/packages/aws-rfdk/lib/deadline/test/repository.test.ts
+++ b/packages/aws-rfdk/lib/deadline/test/repository.test.ts
@@ -1225,7 +1225,7 @@ test('imports repository settings', () => {
   expect(installerGroup.userData.render()).toContain(`aws s3 cp '${repositorySettings.s3ObjectUrl}'`);
 });
 
-test('IFileSystem.usesUserPosixPermissions() = true changes ownership of repository files', () => {
+test('IMountableLinuxFilesystem.usesUserPosixPermissions() = true changes ownership of repository files', () => {
   // GIVEN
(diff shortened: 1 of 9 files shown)


packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -1225,11 +1225,15 @@ test('imports repository settings', () => {
   expect(installerGroup.userData.render()).toContain(`aws s3 cp '${repositorySettings.s3ObjectUrl}'`);
 });
 
-test('changes ownership of repository files', () => {
+test('IFileSystem.usesUserPosixPermissions() = true changes ownership of repository files', () => {
```
jericht:
```suggestion
test('IMountableLinuxFilesystem.usesUserPosixPermissions() = true changes ownership of repository files', () => {
```

minor nit

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -1238,3 +1242,21 @@ test('changes ownership of repository files', () => {
   // THEN
   expect(script).toMatch('-o 1000:1000');
 });
+
+test('IFileSystem.usesUserPosixPermissions() = false does not change ownership of repository files', () => {
```
jericht:
```suggestion
test('IMountableLinuxFilesystem.usesUserPosixPermissions() = false does not change ownership of repository files', () => {
```

jericht (changes requested):
Looks good! Just a couple of really minor comments. Thanks for getting this fix up!

ddneilson (approved):

jericht (approved):

PR #528, opened 2021-08-09
feat(deadline): add Deadline Secrets Management integration in the Render Queue

jericht:
Depends on:
- https://github.com/aws/aws-rfdk/pull/514
- Relevant changes to the Deadline Docker recipes

### Notes
Adds Deadline Secrets Management configuration to the `RenderQueue` construct.

### Testing
- Deployed a modified version of the example app that uses the staging workflow with the modified Deadline Docker recipes and verified that:
  - The Render Queue ECS task successfully configured Secrets Management
  - When an ECS task restarts (e.g. due to Docker container exiting), it uses the same Secrets Management identity as the previous ECS task
- Deployed the original example app without Secrets Management enabled, then updated it by enabling Secrets Management on the Repository and RenderQueue, then verified that Secrets Management was configured successfully on the RCS by running Secrets Management commands such as `ListAllMachines`, `ListAllSecrets`, `PutSecret`, and `GetSecret`.

----


packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
```diff
@@ -377,6 +377,24 @@ export interface RenderQueueProps {
...
+ * {
+ *   username: string,
+ *   password: string,
+ * }
+ * ```
+ *
+ * @see https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html
+ * @default Deadline Secrets Management is not enabled
+ */
+  readonly secretsManagementCredentials?: ISecret;
```
horsmand: Let's just take this value off the repository instead of exposing it as a property on the interface.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -701,6 +728,27 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
 
     containerDefinition.addMountPoints(connection.readWriteMountPoint);
 
+    if (props.secretsManagementOptions !== undefined) {
```
horsmand, line 742: Just check `props.repository.secretsManagementSettings.enabled`.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -412,12 +413,31 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+      if (props.trafficEncryption?.internalProtocol !== ApplicationProtocol.HTTPS) {
+        errors.push('The internal protocol on the Render Queue is not HTTPS.');
+      }
+      if (externalProtocol !== ApplicationProtocol.HTTPS) {
+        errors.push('External TLS on the Render Queue is not enabled.');
+      }
+      if (errors.length > 0) {
+        throw new Error(`Secrets Management cannot be enabled on the Render Queue for the following reasons:\n${errors.join('\n')}`);
+      }
+    }
```
horsmand:
```suggestion
    if (props.repository.secretsManagementSettings.enabled) {
      const errors = [];
      if (props.trafficEncryption?.internalProtocol !== ApplicationProtocol.HTTPS) {
        errors.push('The internal protocol on the Render Queue is not HTTPS.');
      }
      if (externalProtocol !== ApplicationProtocol.HTTPS) {
        errors.push('External TLS on the Render Queue is not enabled.');
      }
      if (errors.length > 0) {
        throw new Error(`Secrets Management cannot be enabled on the Render Queue for the following reasons:\n${errors.join('\n')}`);
      }
    }
```

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
```diff
@@ -132,7 +132,7 @@ if [ ! -z "${SECRET_MANAGEMENT_ARN+x}" ]; then
     exit 1
   fi
   echo "Secret management is enabled. Credentials are stored in secret: $SECRET_MANAGEMENT_ARN"
-  SECRET_MANAGEMENT_ARG="--installSecretsManagement true --secretsAdminName \"$SECRET_MANAGEMENT_USER\" --secretsAdminPassword \"$SECRET_MANAGEMENT_PASSWORD\""
+  SECRET_MANAGEMENT_ARG="--installSecretsManagement true --secretsAdminName $SECRET_MANAGEMENT_USER --secretsAdminPassword $SECRET_MANAGEMENT_PASSWORD"
```
horsmand: Were the double quotes causing problems here? I'm worried that the password requires special characters in it and without surrounding it in quotes bash could end up interpreting them in unintended ways

horsmand (changes requested):
Looks good overall, just a few minor things that are hopefully quick fixes.

packages/aws-rfdk/lib/deadline/scripts/bash/installDeadlineRepository.sh
jericht: Yeah, turns out the escaped double quotes within a double-quoted bash string get treated literally when we use this variable in the call to the repository installer. I've updated this to use bash arrays which will give us the result we want (bash won't treat the quotes literally like before and any whitespace inside a parameter value will be respected)

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: Updated this code

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: This is actually a prop for this method, but I think I know what you mean. I've updated the calling code in the constructor to just use the Repository's SM prop

packages/aws-rfdk/lib/deadline/lib/render-queue-ref.ts
jericht: Sure. I had it this way to make enabling SM on the RQ an explicit step, but we can just update the docstrings so that it's clear that enabling SM on the Repository will implicitly enable it on the RQ. Updated.

I also updated our unit tests to disable SM wherever we used an RCS w/o TLS, so let's remember to add a "BREAKING CHANGE" section when we merge this feature branch in because farms that use the default `Repository` and a `RenderQueue` without TLS enabled won't `cdk synth` anymore.

horsmand (approved):

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -412,12 +413,31 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+      if (internalProtocol !== ApplicationProtocol.HTTPS) {
+        errors.push('The internal protocol on the Render Queue is not HTTPS.');
+      }
+      if (externalProtocol !== ApplicationProtocol.HTTPS) {
+        errors.push('External TLS on the Render Queue is not enabled.');
+      }
+      if (errors.length > 0) {
+        throw new Error(`Secrets Management cannot be enabled on the Render Queue for the following reasons:\n${errors.join('\n')}`);
+      }
+    }
```
jusiskin: Is there a minimum Deadline version who's Docker Image recipes will support this feature? If so, let's add a validation here for it.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -412,12 +413,31 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+        errors.push('The Repository does not have Secrets Management credentials');
+      }
+      if (internalProtocol !== ApplicationProtocol.HTTPS) {
+        errors.push('The internal protocol on the Render Queue is not HTTPS.');
+      }
+      if (externalProtocol !== ApplicationProtocol.HTTPS) {
+        errors.push('External TLS on the Render Queue is not enabled.');
+      }
+      if (errors.length > 0) {
+        throw new Error(`Secrets Management cannot be enabled on the Render Queue for the following reasons:\n${errors.join('\n')}`);
```
jusiskin: Does it make sense to mentioned that secrets management is enabled on the supplied repository? This is a new implicit default on the `Repository` construct so users upgrading might not understand this error message without such a prompt.

jusiskin (changes requested):
Really great job here. Just two small suggestions around the validation.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: Good idea, updated the message

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: Right, I missed that. Added

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -415,6 +421,9 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
 
     if (props.repository.secretsManagementSettings.enabled) {
       const errors = [];
+      if (props.version.isLessThan(RenderQueue.MINIMUM_SECRETS_MANAGEMENT_VERSION)) {
+        errors.push(`The supplied Deadline version is lower than the minimum required version: ${RenderQueue.MINIMUM_SECRETS_MANAGEMENT_VERSION.toString()}`);
```
jusiskin: nit: can we put the received version in the error message? We can use [`VersionQuery.versionString`](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.deadline.VersionQuery.html#versionstring) for that purpose.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -214,15 +215,21 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
   */
   private static readonly MINIMUM_LOAD_BALANCING_VERSION = new Version([10, 1, 10, 0]);
 
+  /**
+   * The minimum Deadline version required to enable Deadline Secrets Management on the Render Queue.
+   */
+  private static readonly MINIMUM_SECRETS_MANAGEMENT_VERSION = new Version([10, 1, 18, 0]);
```
horsmand: This is tricky to add at this point because there are features being added to Deadline to make Secrets Management easier to use that are still unreleased, and we don't yet support `10.1.18` in RFDK, so adding this here will render our feature branch unusable. 

Maybe let's set this to 10.1.15 for now (that release had a fix for when RCS is behind a load balancer that is absolutely required) and we can update it after Deadline releases the new features and before we merge this from the feature branch to mainline.

horsmand (changes requested):

jusiskin (changes requested):
One minor suggestion remaining.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: Sounds good, I've put a TODO in there for now so we don't forget to do that before we merge this into mainline

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: Good idea, updated

horsmand (approved):

jusiskin (approved):

PR #537, opened 2021-08-17
fix(deadline): fix issue in client TLS configuration for Deadline 10.1.18.5

jusiskin:
## Problem

The behavior of the `ChangeRepository[SkipValidation]` Deadline command in 10.1.18.5 has changed. It now clears the `ProxySSLCA` setting in `deadline.ini` before attempting to apply/validate the configuration. RFDK relied on the previous behavior where the `ProxySSLCA` was left unchanged.

## Solution

We cannot use `ChangeRepository` without supplying a "client certificate" since this will disable TLS. Instead, we have switched from using `ChangeRepository` to instead use `SetIniFileSetting ProxyRoot <RQ_ENDPOINT>` .  RFDK also relied on having `ChangeRepository` validate the connection and error if the connection settings were incorrect. We now change that validation to instead run a separate Deadline command to test the connection.

We made the same change to the Deadline configuration script that runs on the RFDK integ tests.

## Testing

**WIP...**

Proposed test plan:

1.  Deploy the `All-in-AWS-Basic` example app using staged recipes for Deadline `10.1.18.5` and Deadline 10.1.18.5 AWS Portal AMIs. Confirm that the deploy succeeds and Workers are connected properly to Deadline. Repeat for Linux/Windows workers and for pre-`10.1.18` Worker AMIs.
2.  Run the RFDK integ tests and confirm they pass for both pre `10.1.18` and `10.1.18.5`

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
+++ b/packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
@@ -12,7 +12,7 @@ import errno
 import os
 import re
 import subprocess
-from sys import platform
+from sys import platform, stdout
 
 import boto3
 
@@ -133,8 +133,8 @@ def configure_deadline( config ):
 
     # Ensure that the client is configured to connect to a Remote RCS.
     call_deadline_command(['SetIniFileSetting', 'ConnectionType', 'Remote'])
+    call_deadline_command(['SetIniFileSetting', 'ProxyRoot', config.render_queue.address])
 
-    repo_args = ['ChangeRepository','Proxy',config.render_queue.address]
     if config.render_queue.scheme == 'http':
         print( "Configuring Deadline to connect to the Render Queue (%s) using HTTP Traffic" % config.render_queue.address )
         #Ensure SSL is disabled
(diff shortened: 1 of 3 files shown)


packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
```diff
@@ -168,7 +168,15 @@ def configure_deadline( config ):
 
             call_deadline_command(['SetIniFileSetting', 'ProxySSLCA', cert_path])
             call_deadline_command(['SetIniFileSetting', 'ClientSSLAuthentication', 'NotRequired'])
-            repo_args.append(cert_path)
+            
+            # Validate Deadline connection
+            print("Testing Deadline connection...")
+            stdout.flush()
+            try:
+                call_deadline_command(['GetJobs'])
```
jusiskin: We may want to change this to a command that scales better. In the case of an upgrade, there may be many jobs

horsmand:
Closing this as we have this commit updated to fix the integ tests and in a new PR: https://github.com/aws/aws-rfdk/pull/543

PR #543, opened 2021-08-24
fix(deadline): fix issue in client TLS configuration for Deadline 10.1.18.5

horsmand:
I applied a fix to [this draft PR](https://github.com/aws/aws-rfdk/pull/537) to get the integ tests to pass, so I've created a new PR from my fork.

## Problem

The behavior of the `ChangeRepository[SkipValidation]` Deadline command in 10.1.18.5 has changed. It now clears the `ProxySSLCA` setting in `deadline.ini` before attempting to apply/validate the configuration. RFDK relied on the previous behavior where the `ProxySSLCA` was left unchanged.

## Solution

We cannot use `ChangeRepository` without supplying a "client certificate" since this will disable TLS. Instead, we have switched from using `ChangeRepository` to instead use `SetIniFileSetting ProxyRoot <RQ_ENDPOINT>` .  RFDK also relied on having `ChangeRepository` validate the connection and error if the connection settings were incorrect. We now change that validation to instead run a separate Deadline command to test the connection.

We made the same change to the Deadline configuration script that runs on the RFDK integ tests.

## Testing

The RFDK integ tests are now passing using the Deadline Client 10.1.18.5 to stage the Docker recipes and the `Deadline Worker Base Image Windows 10.1.18.5 2021-07-30T215517Z - ami-0d5da435a9d85896b` AMI for the worker fleets. 

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
+++ b/packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
@@ -12,7 +12,7 @@ import errno
 import os
 import re
 import subprocess
-from sys import platform
+from sys import platform, stdout
 
 import boto3
 
@@ -133,8 +133,8 @@ def configure_deadline( config ):
 
     # Ensure that the client is configured to connect to a Remote RCS.
     call_deadline_command(['SetIniFileSetting', 'ConnectionType', 'Remote'])
+    call_deadline_command(['SetIniFileSetting', 'ProxyRoot', config.render_queue.address])
 
-    repo_args = ['ChangeRepository','Proxy',config.render_queue.address]
     if config.render_queue.scheme == 'http':
         print( "Configuring Deadline to connect to the Render Queue (%s) using HTTP Traffic" % config.render_queue.address )
         #Ensure SSL is disabled
@@ -144,7 +144,7 @@ def configure_deadline( config ):
 
     else:
         print("Configuring Deadline to connect to the Render Queue using HTTPS Traffic")
-        call_deadline_command(['SetIniFileSetting','ProxyUseSSL','True'])
+        call_deadline_command(['SetIniFileSetting', 'ProxyUseSSL', 'True'])
 
         try:
             os.makedirs(CERT_DIR)
@@ -168,7 +168,13 @@ def configure_deadline( config ):
 
             call_deadline_command(['SetIniFileSetting', 'ProxySSLCA', cert_path])
             call_deadline_command(['SetIniFileSetting', 'ClientSSLAuthentication', 'NotRequired'])
-            repo_args.append(cert_path)
+
+            # Validate Deadline connection
+            print("Testing Deadline connection...")
+            stdout.flush()
+            call_deadline_command(['GetRepositoryVersion'])
+
+            print("Deadline connection configured correctly")
         else:
             """
             If we are configuring Deadline to connect using a client cert we need to:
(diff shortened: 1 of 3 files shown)


packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
```diff
@@ -182,17 +190,18 @@ def configure_deadline( config ):
...
             if config.client_tls_cert_passphrase:
                 passphrase = fetch_secret(config.client_tls_cert_passphrase)
                 repo_args.append(passphrase)
-
-    change_repo_results = call_deadline_command(repo_args)
-    if change_repo_results.startswith('Deadline configuration error:'):
-        print(change_repo_results)
-        raise Exception(change_repo_results)
+            change_repo_results = call_deadline_command(repo_args)
+            if change_repo_results.startswith('Deadline configuration error:'):
```
jericht: Are we certain that errors in connecting to the repository with always start with this string? Is it possible to capture the exit code of `ChangeRepository` and check that instead?

jericht (changes requested):

packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
horsmand: I guess this can be removed now. Error checking was added on line 220.

horsmand: I performed a test using Deadline 10.1.18.5 without the fix applied so I could be sure that I'd be failing to run the Deadline Command. The log shows that it threw the `ValueError` exception when checking `proc.returncode` and never made it to the line to check if the result started with `Deadline configuration error:`, so I went ahead and removed this check.

jericht: Great, thanks!

jericht (dismissed):

packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
```diff
@@ -168,7 +168,15 @@ def configure_deadline( config ):
...
             call_deadline_command(['SetIniFileSetting', 'ProxySSLCA', cert_path])
             call_deadline_command(['SetIniFileSetting', 'ClientSSLAuthentication', 'NotRequired'])
-            repo_args.append(cert_path)
+
+            # Validate Deadline connection
+            print("Testing Deadline connection...")
+            stdout.flush()
+            try:
+                call_deadline_command(['GetRepositoryVersion'])
+            except Exception as e:
```
ddneilson: We should rethrow here, no? Else, if there's a connection error the log will look like:
```
Deadline connection error: ...
Deadline connection configured correctly
```
and this function will exit normally. This results in the script returning a 'success' return code whereas previously this function would throw an exception on error and result in the script returning an 'error' return code.

packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
```diff
@@ -182,17 +190,16 @@ def configure_deadline( config ):
...
 
-            call_deadline_command(['SetIniFileSetting', 'ProxySSLCA', cert_path])
             call_deadline_command(['SetIniFileSetting', 'ClientSSLAuthentication', 'Required'])
+
+            repo_args = ['ChangeRepository', 'Proxy', config.render_queue.address]
             repo_args.append(cert_path)
             if config.client_tls_cert_passphrase:
                 passphrase = fetch_secret(config.client_tls_cert_passphrase)
                 repo_args.append(passphrase)
+            change_repo_results = call_deadline_command(repo_args)
```
ddneilson, line 199: Just making 100% sure... you've verified that ChangeRepository returns a non-zero return code from deadlinecommand when it fails to connect to the repo? (i.e. when the settings are incorrect)?

ddneilson (changes requested):

packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
horsmand: Since the `proc.returncode` check was only added to the `call_deadline_command` in this commit, and the `ChangeRepository` command is also moved to only be called in the case where a client cert is applied (which isn't a valid use case in RFDK right now), I don't see why this should be a concern.

That said, I did kind of verify this by using Deadline 10.1.18.5 without the fix applied so that the `ChangeRepository` command would fail and the `call_deadline_command` should throw the `ValueError` exception when checking `proc.returncode`.

packages/aws-rfdk/lib/deadline/scripts/python/client-rq-connection.py
horsmand: Yeah you're right. I tested this by setting the RCS endpoint to 'test' and then deployed with and without the `try/except`. It failed to deploy both times, but with the `try/except` in place, the logs did say that the connection was configured correctly when it wasn't.

I removed it.

ddneilson (approved):

jericht (approved):

PR #573, opened 2021-09-17
feat(deadline): validate minimum Deadline version for secrets management

kozlove-aws:
### Problem

Deadline fully support all changes in RFDK only from version 10.1.19.0.
We need to validate that Deadline Secret management is enabled only from this version

### Solution
 - Moved constant with minimum version for secret management in Version 
 - Updated min version to 10.1.19.0
 - Added version validation in Repository construct.
 
----


packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -584,6 +585,10 @@ export class Repository extends Construct implements IRepository {
 
     this.version = props.version;
 
+    if ((props.secretsManagementSettings?.enabled ?? true) && this.version.isLessThan(Version.MINIMUM_SECRETS_MANAGEMENT_VERSION)) {
+      throw new Error(`The supplied Deadline version (${props.version.versionString}) does not support Deadline Secrets Management in RFDK. Either upgrade Deadline to the minimum required version (${Version.MINIMUM_SECRETS_MANAGEMENT_VERSION.versionString}) or disable the feature in the Repository's construct properties.`);
+    }
```
horsmand: I know I said we should throw an error like this if the version doesn't meet the requirement, but we discussed it on Slack yesterday and we landed on disabling the feature instead. We shouldn't have to change anything in the RenderQueue, but let's adjust this to something like:
```suggestion
    const meetsMinSecretsVersion = !this.version.isLessThan(Version.MINIMUM_SECRETS_MANAGEMENT_VERSION);
    const secretsManagementIsEnabled = props.secretsManagementSettings?.enabled ?? meetsMinSecretsVersion;
    
    if (secretsManamentIsEnabled && !meetsMinSecretsVersion) {
      throw new Error(`The supplied Deadline version (${props.version.versionString}) does not support Deadline Secrets Management in RFDK. Either upgrade Deadline to the minimum required version (${Version.MINIMUM_SECRETS_MANAGEMENT_VERSION.versionString}) or disable the feature in the Repository's construct properties.`);
    }
```

horsmand (changes requested):

packages/aws-rfdk/lib/deadline/lib/repository.ts
kozlove-aws: Good point. Updated.

horsmand (dismissed):
I made a commit to remove one extra blank line because that was the only issue I saw left and wanted to save the extra iteration on the review just to remove it. Everything looks good to me now.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -858,7 +858,7 @@ test('repository instance is created with correct installer path version', () =>
 
   // THEN
   const script = (repo.node.defaultChild as AutoScalingGroup).userData;
-  expect(script.render()).toMatch(/10\.1\.9\.2/);
+  expect(script.render()).toMatch(version.versionString);
```
jusiskin: [`expect(...).toMatch(....)`](https://jestjs.io/docs/expect#tomatchregexp--string) will treat the passed in string as a regular expression. Because of this, the periods (`.`) in the version string will be treated as matching any character.

Instead, how about we use [`expect.stringContaining(...)`](https://jestjs.io/docs/expect#expectstringcontainingstring):

```suggestion
  expect(script.render()).toEqual(expect.stringContaining(version.versionString));
```

jusiskin (changes requested):
just one minor test semantic. Otherwise this looks great.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
jusiskin: Looks good now. resolved

jusiskin (dismissed):
This looks good to go. We'll need to wait until the minimum Deadline version is released before we can merge (with a possible rebase required).

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -17,6 +17,11 @@ export class Version implements IPatchVersion {
    */
   public static readonly MINIMUM_SUPPORTED_DEADLINE_VERSION = new Version([10, 1, 9, 2]);
 
+  /**
+   * The minimum Deadline version required to enable Deadline Secrets Management.
+   */
+  public static readonly MINIMUM_SECRETS_MANAGEMENT_VERSION = new Version([10, 1, 19, 0]);
```
horsmand, line 23:
```suggestion
  public static readonly MINIMUM_SECRETS_MANAGEMENT_VERSION = new Version([10, 1, 19, 4]);
```

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -118,7 +118,7 @@ beforeEach(() => {
     }
   }
 
-  version = new MockVersion([10,1,9,2]);
+  version = new MockVersion([10,1,19,0]);
```
horsmand:
```suggestion
  version = new MockVersion([10,1,19,4]);
```

horsmand (changes requested):
The release is out and the version number is `10.1.19.4` so let's just bump the min version to that.

packages/aws-rfdk/lib/deadline/lib/version.ts
jusiskin: Unfortunately, we are unable to compare a potentially un-resolved VersionQuery to a non-zero patch version. Looks like CI is failing with this error:

```
Cannot compare a VersionQuery to a fully-qualified version with a non-zero patch number
```

The best we can do is change the logic back to making sure we're greater than 10.1.19.0.

kozlove-aws: I believe it is not so important since version 10.1.19.0 should support secret management as well

horsmand (approved):

jusiskin (approved):

PR #576, opened 2021-09-20
feat(deadline): configure identity registration settings for deadline clients

jusiskin:
## Summary

This change makes RFDK automatically configure Deadline Secrets Management identity registration settings based on the RFDK constructs that are present in a CDK application.

The changes are broken out by construct below:

### `DeploymentInstance` (new)

This construct is an extraction/generalization of the CDK/CloudFormation pattern used by the `Repository` construct. It deploys an EC2 Auto-Scaling Group with a size of one. The Auto-Scaling Group is configured to require one CloudFormation success signal which is only sent after the user-data completes without error. The user data can optionally (by default enabled) self-terminate the instance by scaling down its own auto-scaling group to zero.

For now, this construct is being kept internal to RFDK and not exported. It was designed with the possibility of making it a public construct in the future.

### `SecretsManagementIdentityRegistration` (new)

The `SecretsManagementIdentityRegistration` construct is also introduced in this PR. It is responsible for configuring the user data (on a `DeploymentInstance`) to configure the identity registration settings. It adds IAM permissions and user-data to perform the following steps:

1.  download and install the Deadline Client
2.  configure Deadline to make a direct connection to the Deadline Repository
3.  run a python script that
    1.  fetches CIDR ranges for subnets involved
    2.  fetches existing identity registration settings
    3.  creates/updates/deletse identity registration settings so that they match what's desired

The main API of this construct is the `.addSubnetIdentityRegistrationSetting(...)` method which a caller can use to specify the subnets of the client, the desired secrets management [role](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html#assigned-roles) and [registration status](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html#registration-status).

### `RenderQueue` (modified)

This PR introduces a new `configureSecretsManagementAutoRegistration(...)` method to `IRenderQueue`. When called on the `RenderQueue`, it will lazily create `DeploymentInstance` and `SecretsManagementIdentityRegistration` constructs and delegate the work to the `SecretsManagementIdentityRegistration` construct. It also accepts a `dependent` prop which is used to enforce a CloudFormation dependency on a dependent resource of the caller's choosing.

The CloudWatch logs for the `DeploymentInstance` are sent to the `/renderfarm/ConfigureRepository` log group with the intention that the behavior of the deployment instance can be augmented in the future.

### `WorkerInstanceFleet` / `ConfigureSpotEventPlugin` / `UsageBasedLicensing` (modified)

The RFDK constructs that deploy Deadline Clients were modified to call the `configureSecretsManagementAutoRegistration(...)` method 

### `IVersion` / `VersionQuery` / `ThinkboxDockerRecipes.version` (modified)

The `IVersion` interface and classes that implement it were modified to return the path to the Deadline Client installer location. This was required in order for the `SecretsManagementIdentityRegistration` construct to be able to install the Deadline Client on instance deployed by the `DeploymentInstance`.

## Testing

Unit tests were updated to reflect all of the changes

Automatic identity registration of Deadline Clients was end-to-end tested by deploying the `All-In-AWS-Infrastructure-Basic` TypeScript example app with Thinkbox Usage-Based Licensing (UBL). Verified that:

*   Client identities were properly being assigned with the "Client" role and statuses by checking in a connected Deadline Monitor
*   Logs in the `renderfarm/ConfigureRepository` log group showed correct behavior
*   Clients were able to access Deadline-managed secrets. Thinkbox UBL requires the "Client" role to be assigned properly. Renders were successful.

----

### DIFF
--- a/packages/aws-rfdk/lib/core/lib/deployment-instance.ts
+++ b/packages/aws-rfdk/lib/core/lib/deployment-instance.ts
@@ -9,7 +9,6 @@ import {
   UpdatePolicy,
 } from '@aws-cdk/aws-autoscaling';
 import {
-  AmazonLinuxImage,
   Connections,
   IConnectable,
   IMachineImage,
@@ -17,6 +16,7 @@ import {
   InstanceSize,
   InstanceType,
   IVpc,
+  MachineImage,
   SubnetSelection,
   SubnetType,
 } from '@aws-cdk/aws-ec2';
@@ -79,7 +79,7 @@ export interface DeploymentInstanceProps {
   readonly logGroupName?: string;
 
   /**
-   * Properties for setting up the Deadline DeploymentInstance's LogGroup in CloudWatch
+   * Properties for setting up the DeploymentInstance's LogGroup in CloudWatch
    *
    * @default the LogGroup will be created with all properties' default values to the LogGroup: /renderfarm/<construct id>
    */
@@ -164,7 +164,7 @@ export class DeploymentInstance extends Construct implements IScriptHost, IConne
     this.asg = new AutoScalingGroup(this, 'ASG', {
       instanceType: props.instanceType ?? InstanceType.of(InstanceClass.T3, InstanceSize.SMALL),
       keyName: props.keyName,
-      machineImage: props.machineImage ?? new AmazonLinuxImage(),
+      machineImage: props.machineImage ?? MachineImage.latestAmazonLinux(),
       vpc: props.vpc,
       vpcSubnets: props.vpcSubnets ?? {
         subnetType: SubnetType.PRIVATE,
@@ -203,14 +203,23 @@ export class DeploymentInstance extends Construct implements IScriptHost, IConne
     }
   }
 
+  /**
+   * @inheritdoc
+   */
   public get osType() {
     return this.asg.osType;
   }
 
+  /**
+   * @inheritdoc
+   */
   public get userData() {
     return this.asg.userData;
   }
 
+  /**
+   * @inheritdoc
+   */
   public get grantPrincipal() {
     return this.asg.grantPrincipal;
   }
@@ -222,6 +231,8 @@ export class DeploymentInstance extends Construct implements IScriptHost, IConne
    *   - The cloud-init log
    *
    * @param asg The auto-scaling group
+   * @param groupName The name of the Log Group, or suffix of the Log Group if `logGroupProps.logGroupPrefix` is
+   *                  specified
    * @param logGroupProps The properties for LogGroupFactory to create or fetch the log group
    */
   private configureCloudWatchAgent(asg: AutoScalingGroup, groupName: string, logGroupProps?: LogGroupFactoryProps) {
--- a/packages/aws-rfdk/lib/core/lib/deployment-instance.ts
+++ b/packages/aws-rfdk/lib/core/lib/deployment-instance.ts
@@ -46,7 +46,7 @@ import { tagConstruct } from './runtime-info';
  */
 export interface DeploymentInstanceProps {
   /**
-   * The time CloudFormation should wait for the success signals before failing the create/update.
+   * The amount of time that CloudFormation should wait for the success signals before failing the create/update.
    *
    * @default 15 minutes
    */
@@ -75,7 +75,8 @@ export interface DeploymentInstanceProps {
 
   /**
    * Properties for setting up the Deadline DeploymentInstance's LogGroup in CloudWatch
-   * @default - LogGroup will be created with all properties' default values to the LogGroup: /renderfarm/<construct id>
+   *
+   * @default the LogGroup will be created with all properties' default values to the LogGroup: /renderfarm/<construct id>
    */
   readonly logGroupProps?: LogGroupFactoryProps;
 
--- a/packages/aws-rfdk/lib/core/lib/deployment-instance.ts
+++ b/packages/aws-rfdk/lib/core/lib/deployment-instance.ts
@@ -1,0 +1,287 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  AutoScalingGroup,
+  Signals,
+  UpdatePolicy,
+} from '@aws-cdk/aws-autoscaling';
+import {
+  AmazonLinuxImage,
+  Connections,
+  IConnectable,
+  IMachineImage,
+  InstanceClass,
+  InstanceSize,
+  InstanceType,
+  IVpc,
+  SubnetSelection,
+  SubnetType,
+} from '@aws-cdk/aws-ec2';
+import {
+  PolicyStatement,
+} from '@aws-cdk/aws-iam';
+import {
+  Construct,
+  Duration,
+  Names,
+  Stack,
+  Tags,
+} from '@aws-cdk/core';
+
+import {
+  CloudWatchConfigBuilder,
+  CloudWatchAgent,
+  IScriptHost,
+  LogGroupFactory,
+  LogGroupFactoryProps,
+} from '.';
+import { tagConstruct } from './runtime-info';
+
+
+/**
+ * Properties for constructing a `DeploymentInstance`
+ */
+export interface DeploymentInstanceProps {
+  /**
+   * The time CloudFormation should wait for the success signals before failing the create/update.
+   *
+   * @default 15 minutes
+   */
+  readonly executionTimeout?: Duration;
+
+  /**
+   * The instance type to deploy
+   *
+   * @default t3.small
+   */
+  readonly instanceType?: InstanceType;
+
+  /**
+   * An optional EC2 keypair name to associate with the instance
+   *
+   * @default no EC2 keypair is associated with the instance
+   */
+  readonly keyName?: string;
+
+  /**
+   * The log group name for streaming CloudWatch logs
+   *
+   * @default the construct ID is used
+   */
+  readonly logGroupName?: string;
+
+  /**
+   * Properties for setting up the Deadline DeploymentInstance's LogGroup in CloudWatch
+   * @default - LogGroup will be created with all properties' default values to the LogGroup: /renderfarm/<construct id>
+   */
+  readonly logGroupProps?: LogGroupFactoryProps;
+
+  /**
+   * The machine image to use.
+   *
+   * @default latest Amazon Linux 2 image
+   */
+  readonly machineImage?: IMachineImage;
+
+  /**
+   * Whether the instance should self-terminate after the deployment succeeds
+   *
+   * @default true
+   */
+  readonly selfTerminate?: boolean;
+
+  /**
+   * The VPC that the instance should be launched in.
+   */
+  readonly vpc: IVpc;
+
+  /**
+   * The subnets to deploy the instance to
+   *
+   * @default private subnets
+   */
+  readonly vpcSubnets?: SubnetSelection;
+}
+
+/**
+ * Deploys an instance that runs its user data on deployment, waits for that user data to succeed, and optionally
+ * terminates itself afterwards.
+ *
+ * Resources Deployed
+ * ------------------------
+ * - Auto Scaling Group (ASG) with max capacity of 1 instance.
+ * - IAM instance profile, IAM role, and IAM policy
+ * - An Amazon CloudWatch log group that contains the instance cloud-init logs
+ * - A Lambda Function to fetch and existing Log Group or create a new one
+ * - IAM role and policy for the Lambda Function
+ *
+ * Security Considerations
+ * ------------------------
+ * - The instances deployed by this construct download and run scripts from your CDK bootstrap bucket when that instance
+ *   is launched. You must limit write access to your CDK bootstrap bucket to prevent an attacker from modifying the actions
+ *   performed by these scripts. We strongly recommend that you either enable Amazon S3 server access logging on your CDK
+ *   bootstrap bucket, or enable AWS CloudTrail on your account to assist in post-incident analysis of compromised production
+ *   environments.
+ */
+export class DeploymentInstance extends Construct implements IScriptHost, IConnectable {
+  /**
+   * The tag key name used as an IAM condition to restrict autoscaling API grants
+   */
+  private static readonly ASG_TAG_KEY: string = 'resourceLogicalId';
+
+  /**
+   * How often the CloudWatch agent will flush its log files to CloudWatch
+   */
+  private static readonly CLOUDWATCH_LOG_FLUSH_INTERVAL: Duration = Duration.seconds(15);
+
+  /**
+   * The default timeout to wait for CloudFormation success signals before failing the resource create/update
+   */
+  private static readonly DEFAULT_EXECUTION_TIMEOUT = Duration.minutes(15);
+
+  /**
+   * Default prefix for a LogGroup if one isn't provided in the props.
+   */
+  private static readonly DEFAULT_LOG_GROUP_PREFIX: string = '/renderfarm/';
+
+  /**
+   * @inheritdoc
+   */
+  public readonly connections: Connections;
+
+  /**
+   * The auto-scaling group
+   */
+  protected readonly asg: AutoScalingGroup;
+
+  constructor(scope: Construct, id: string, props: DeploymentInstanceProps) {
+    super(scope, id);
+
+    this.asg = new AutoScalingGroup(this, 'ASG', {
+      instanceType: props.instanceType ?? InstanceType.of(InstanceClass.T3, InstanceSize.SMALL),
+      keyName: props.keyName,
+      machineImage: props.machineImage ?? new AmazonLinuxImage(),
+      vpc: props.vpc,
+      vpcSubnets: props.vpcSubnets ?? {
+        subnetType: SubnetType.PRIVATE,
+      },
+      minCapacity: 1,
+      maxCapacity: 1,
+      signals: Signals.waitForAll({
+        timeout: props.executionTimeout ?? DeploymentInstance.DEFAULT_EXECUTION_TIMEOUT,
+      }),
+      updatePolicy: UpdatePolicy.replacingUpdate(),
+    });
+    this.node.defaultChild = this.asg;
+
+    this.connections = this.asg.connections;
+
+    const logGroupName = props.logGroupName ?? id;
+    this.configureCloudWatchAgent(this.asg, logGroupName, props.logGroupProps);
+
+    if (props.selfTerminate ?? true) {
+      this.configureSelfTermination();
+    }
+    this.asg.userData.addSignalOnExitCommand(this.asg);
+
+    // Tag deployed resources with RFDK meta-data
+    tagConstruct(this);
+  }
+
+  /**
+   * Make the execution of the instance dependent upon another construct
+   *
+   * @param dependency The construct that should be dependended upon
+   */
+  public addExecutionDependency(dependency: any): void {
+    if (Construct.isConstruct(dependency)) {
+      this.asg.node.defaultChild!.node.addDependency(dependency);
+    }
+  }
+
+  public get osType() {
+    return this.asg.osType;
+  }
+
+  public get userData() {
+    return this.asg.userData;
+  }
+
+  public get grantPrincipal() {
+    return this.asg.grantPrincipal;
+  }
+
+  /**
+   * Adds UserData commands to configure the CloudWatch Agent running on the deployment instance.
+   *
+   * The commands configure the agent to stream the following logs to a new CloudWatch log group:
+   *   - The cloud-init log
+   *
+   * @param asg The auto-scaling group
+   * @param logGroupProps The properties for LogGroupFactory to create or fetch the log group
+   */
+  private configureCloudWatchAgent(asg: AutoScalingGroup, groupName: string, logGroupProps?: LogGroupFactoryProps) {
+    const prefix = logGroupProps?.logGroupPrefix ?? DeploymentInstance.DEFAULT_LOG_GROUP_PREFIX;
+    const defaultedLogGroupProps = {
+      ...logGroupProps,
+      logGroupPrefix: prefix,
+    };
+    const logGroup = LogGroupFactory.createOrFetch(this, 'DeploymentInstanceLogGroupWrapper', groupName, defaultedLogGroupProps);
+
+    logGroup.grantWrite(asg);
+
+    const cloudWatchConfigurationBuilder = new CloudWatchConfigBuilder(DeploymentInstance.CLOUDWATCH_LOG_FLUSH_INTERVAL);
+
+    cloudWatchConfigurationBuilder.addLogsCollectList(logGroup.logGroupName,
+      'cloud-init-output',
+      '/var/log/cloud-init-output.log');
+
+    new CloudWatchAgent(this, 'CloudWatchAgent', {
+      cloudWatchConfig: cloudWatchConfigurationBuilder.generateCloudWatchConfiguration(),
+      host: asg,
+    });
+  }
+
+  private configureSelfTermination() {
+    // Add a policy to the ASG that allows it to modify itself. We cannot add the ASG name in resources as it will cause
+    // cyclic dependency. Hence, using Condition Keys
+    const tagCondition: { [key: string]: any } = {};
+    tagCondition[`autoscaling:ResourceTag/${DeploymentInstance.ASG_TAG_KEY}`] = Names.uniqueId(this);
+
+    Tags.of(this.asg).add(DeploymentInstance.ASG_TAG_KEY, Names.uniqueId(this));
+
+    this.asg.addToRolePolicy(new PolicyStatement({
+      actions: [
+        'autoscaling:UpdateAutoScalingGroup',
+      ],
+      resources: ['*'],
+      conditions: {
+        StringEquals: tagCondition,
+      },
+    }));
+
+    // Following policy is required to read the aws tags within the instance
+    this.asg.addToRolePolicy(new PolicyStatement({
+      actions: [
+        'ec2:DescribeTags',
+      ],
+      resources: ['*'],
+    }));
+
+    // wait for the log flush interval to make sure that all the logs gets flushed.
+    // this wait can be avoided in future by using a life-cycle-hook on 'TERMINATING' state.
+    const terminationDelay = Math.ceil(DeploymentInstance.CLOUDWATCH_LOG_FLUSH_INTERVAL.toMinutes({integral: false}));
+    this.asg.userData.addOnExitCommands(`sleep ${terminationDelay}m`);
+
+    // fetching the instance id and ASG name and then setting its capacity to 0
+    this.asg.userData.addOnExitCommands(
+      'TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 30" 2> /dev/null)',
+      'INSTANCE="$(curl -s -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id  2> /dev/null)"',
+      'ASG="$(aws --region ' + Stack.of(this).region + ' ec2 describe-tags --filters "Name=resource-id,Values=${INSTANCE}" "Name=key,Values=aws:autoscaling:groupName" --query "Tags[0].Value" --output text)"',
+      'aws --region ' + Stack.of(this).region + ' autoscaling update-auto-scaling-group --auto-scaling-group-name ${ASG} --min-size 0 --max-size 0 --desired-capacity 0',
+    );
+  }
+}
--- a/packages/aws-rfdk/lib/core/lib/deployment-instance.ts
+++ b/packages/aws-rfdk/lib/core/lib/deployment-instance.ts
@@ -46,6 +46,11 @@ import { tagConstruct } from './runtime-info';
  */
 export interface DeploymentInstanceProps {
   /**
+   * The VPC that the instance should be launched in.
+   */
+  readonly vpc: IVpc;
+
+  /**
    * The amount of time that CloudFormation should wait for the success signals before failing the create/update.
    *
    * @default 15 minutes
@@ -93,11 +98,6 @@ export interface DeploymentInstanceProps {
    * @default true
    */
   readonly selfTerminate?: boolean;
-
-  /**
-   * The VPC that the instance should be launched in.
-   */
-  readonly vpc: IVpc;
 
   /**
    * The subnets to deploy the instance to
--- a/packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
+++ b/packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
@@ -15,9 +15,8 @@ import { CfnLaunchConfiguration } from '@aws-cdk/aws-autoscaling';
 import {
   AmazonLinuxImage,
   ExecuteFileOptions,
-  InstanceClass,
-  InstanceSize,
   InstanceType,
+  MachineImage,
   SecurityGroup,
   SubnetType,
   Vpc,
@@ -168,7 +167,7 @@ describe('DeploymentInstance', () => {
 
       test('uses latest Amazon Linux machine image', () => {
         // GIVEN
-        const amazonLinux = new AmazonLinuxImage();
+        const amazonLinux = MachineImage.latestAmazonLinux();
         const imageId: { Ref: string } = stack.resolve(amazonLinux.getImage(stack)).imageId;
 
         // THEN
@@ -512,6 +511,8 @@ describe('DeploymentInstance', () => {
       });
     });
 
+    // RFDK adds the resourceLogicalId tag to the Auto-Scaling Group in order to scope down the permissions of the
+    // IAM policy given to the EC2 instance profile so that only that ASG can be scaled by the instance.
     test('Tagging for self-termination', () => {
       // THEN
       const matcher = resourceTagMatcher('AWS::AutoScaling::AutoScalingGroup', 'resourceLogicalId', cdk.Names.uniqueId(target));
@@ -608,11 +609,6 @@ describe('DeploymentInstance', () => {
         '/renderfarm/foo',
       ],
       [
-        { logGroupName: 'bar' },
-        // expected final log group name
-        '/renderfarm/bar',
-      ],
-      [
         {
           logGroupPrefix: 'logGroupPrefix',
         },
@@ -643,31 +639,23 @@ describe('DeploymentInstance', () => {
   });
 
   // GIVEN
-  describe.each<[InstanceType | undefined, string]>([
-    [undefined, 'Undefined'],
-    [new InstanceType('c5.large'), 'C5Large'],
-    [new InstanceType('m5.2xlarge'), 'M52XLarge'],
-  ])('.instanceType = %s', (instanceType, stackNameSuffix) => {
+  test('uses specified instance type', () => {
     // GIVEN
-    const expectedInstanceType = instanceType ?? InstanceType.of(InstanceClass.T3, InstanceSize.SMALL);
-    beforeAll(() => {
-      // We want an isolated stack to ensure expectCDK is only searching resources
-      // synthesized by the specific DeploymentInstance stack
-      stack = new cdk.Stack(app, `InstanceTypeStack${stackNameSuffix}`);
+    const instanceType = new InstanceType('c5.large');
+    // We want an isolated stack to ensure expectCDK is only searching resources
+    // synthesized by the specific DeploymentInstance stack
+    stack = new cdk.Stack(app, 'InstanceTypeStack');
+
+    // WHEN
+    new DeploymentInstance(stack, DEFAULT_CONSTRUCT_ID, {
+      vpc,
+      instanceType,
     });
 
-    test(`AWS::AutoScaling::LaunchConfiguration.InstanceType = ${expectedInstanceType}`, () => {
-      // WHEN
-      new DeploymentInstance(stack, DEFAULT_CONSTRUCT_ID, {
-        vpc,
-        instanceType,
-      });
-
-      // THEN
-      expectCDK(stack).to(haveResourceLike('AWS::AutoScaling::LaunchConfiguration', {
-        InstanceType: expectedInstanceType.toString(),
-      }));
-    });
+    // THEN
+    expectCDK(stack).to(haveResourceLike('AWS::AutoScaling::LaunchConfiguration', {
+      InstanceType: instanceType.toString(),
+    }));
   });
 
   describe('.selfTermination = false', () => {
@@ -694,7 +682,7 @@ describe('DeploymentInstance', () => {
       ]));
     });
 
-    test('is not grante IAM permissions to scale the Auto-Scaling Group', () => {
+    test('is not granted IAM permissions to scale the Auto-Scaling Group', () => {
       // GIVEN
       const instanceRole = (
         target
@@ -710,6 +698,7 @@ describe('DeploymentInstance', () => {
             {
               Action: 'autoscaling:UpdateAutoScalingGroup',
               Condition: {
+                // This tag is added by RFDK to scope down the permissions of the policy for least-privilege
                 StringEquals: { 'autoscaling:ResourceTag/resourceLogicalId': cdk.Names.uniqueId(target) },
               },
               Effect: 'Allow',
@@ -740,16 +729,13 @@ describe('DeploymentInstance', () => {
   });
 
   // GIVEN
-  describe.each<[string]>([
-    ['PT10M'],
-  ])('.executionTimeout = %s', (executionTimeoutStr) => {
+  describe('.executionTimeout is specified', () => {
+    const  executionTimeout = cdk.Duration.minutes(30);
 
-    let executionTimeout: cdk.Duration;
     beforeAll(() => {
       // GIVEN
       // Use a clean stack to not pollute other stacks with resources
-      stack = new cdk.Stack(app, `ExecutionTimeout${executionTimeoutStr}`);
-      executionTimeout = cdk.Duration.parse(executionTimeoutStr);
+      stack = new cdk.Stack(app, 'ExecutionTimeout');
       const deploymentInstanceProps: DeploymentInstanceProps = {
         vpc,
         executionTimeout,
@@ -760,7 +746,7 @@ describe('DeploymentInstance', () => {
     });
 
     // THEN
-    test('AWS::AutoScaling::AutoScalingGroup creation policy signal timeout is set', () => {
+    test('AWS::AutoScaling::AutoScalingGroup creation policy signal timeout is set accordingly', () => {
       expectCDK(stack).to(haveResourceLike(
         'AWS::AutoScaling::AutoScalingGroup',
         {
--- a/packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
+++ b/packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
@@ -439,6 +439,7 @@ describe('DeploymentInstance', () => {
               {
                 Action: 'autoscaling:UpdateAutoScalingGroup',
                 Condition: {
+                  // This tag is added by RFDK to scope down the permissions of the policy for least-privilege
                   StringEquals: { 'autoscaling:ResourceTag/resourceLogicalId': cdk.Names.uniqueId(target) },
                 },
                 Effect: 'Allow',
--- a/packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
+++ b/packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
@@ -28,7 +28,6 @@ import { Bucket } from '@aws-cdk/aws-s3';
 import { Asset } from '@aws-cdk/aws-s3-assets';
 import { StringParameter } from '@aws-cdk/aws-ssm';
 import * as cdk from '@aws-cdk/core';
-import { Duration } from '@aws-cdk/core';
 
 import {
   DeploymentInstance,
@@ -60,23 +59,22 @@ describe('DeploymentInstance', () => {
   let stack: cdk.Stack;
   let vpc: Vpc;
   let target: DeploymentInstance;
-  let deploymentInstanceProps: DeploymentInstanceProps;
 
-  beforeEach(() => {
+  beforeAll(() => {
     // GIVEN
     app = new cdk.App();
     depStack = new cdk.Stack(app, 'DepStack');
-    stack = new cdk.Stack(app, 'Stack');
     vpc = new Vpc(depStack, 'VPC');
-    deploymentInstanceProps = {
-      vpc,
-    };
   });
 
   describe('defaults', () => {
-    beforeEach(() => {
+
+    beforeAll(() => {
       // GIVEN
-      target = new DeploymentInstance(stack, DEFAULT_CONSTRUCT_ID, deploymentInstanceProps);
+      stack = new cdk.Stack(app, 'DefaultsStack');
+      target = new DeploymentInstance(stack, DEFAULT_CONSTRUCT_ID, {
+        vpc,
+      });
     });
 
     describe('Auto-Scaling Group', () => {
@@ -186,62 +184,6 @@ describe('DeploymentInstance', () => {
         }));
       });
 
-      describe('User Data', () => {
-        beforeEach(() => {
-          // WHEN
-          target = new DeploymentInstance(stack, 'DeploymentInstanceNew', {
-            vpc,
-            // a hack to be able to spy on the user data's "addOnExitCommand" and "addExecuteFileCommand" methods. NOTE:
-            // this is a slight deviation from a true default test, because we're passing in the "machineImage" prop. We use
-            // open-box knowledge of the default machineImage and change nothing but spying on the method
-            machineImage: new AmazonLinuxWithUserDataSpy(),
-          });
-        });
-
-        test('configures self-termination', () =>{
-          // THEN
-          expect(target.userData.addOnExitCommands).toHaveBeenCalledWith(
-            'TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 30" 2> /dev/null)',
-            'INSTANCE="$(curl -s -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id  2> /dev/null)"',
-            `ASG="$(aws --region ${stack.region} ec2 describe-tags --filters "Name=resource-id,Values=\${INSTANCE}" "Name=key,Values=aws:autoscaling:groupName" --query "Tags[0].Value" --output text)"`,
-            `aws --region ${stack.region} autoscaling update-auto-scaling-group --auto-scaling-group-name \${ASG} --min-size 0 --max-size 0 --desired-capacity 0`,
-          );
-        });
-
-        test('configures CloudWatch Agent', () =>{
-          // GIVEN
-          const spy = target.userData.addExecuteFileCommand as jest.Mock<void, [ExecuteFileOptions]>;
-          const cloudWatchConfigSsmParam = (
-            target
-              .node.findChild('StringParameter')
-          ) as StringParameter;
-
-          // THEN
-
-          // Should have been called
-          expect(spy.mock.calls.length).toBeGreaterThanOrEqual(1);
-
-          // The first call...
-          const executeFileOptions = spy.mock.calls[0][0];
-
-          // Should have been called with arguments
-          const args = executeFileOptions.arguments;
-          expect(args).not.toBeUndefined();
-
-          const splitArgs = args!.split(' ');
-          // Should have three arguments
-          expect(splitArgs).toHaveLength(3);
-
-          // Specify the flag to install the CloudWatch Agent
-          expect(splitArgs[0]).toEqual('-i');
-          // Should pass the region
-          expect(stack.resolve(splitArgs[1])).toEqual(stack.resolve(stack.region));
-          // Should pass the SSM parameter containing the CloudWatch Agent configuration
-          expect(stack.resolve(splitArgs[2])).toEqual(stack.resolve(cloudWatchConfigSsmParam.parameterName));
-        });
-
-      });
-
       test('Uses created Security Group', () => {
         // GIVEN
         const securityGroup = (target
@@ -293,7 +235,7 @@ describe('DeploymentInstance', () => {
     describe('ASG IAM role', () => {
       let instanceRole: iam.CfnRole;
 
-      beforeEach(() => {
+      beforeAll(() => {
         // GIVEN
         instanceRole = (
           target
@@ -519,7 +461,7 @@ describe('DeploymentInstance', () => {
       });
     });
 
-    describe('CloudWatch Agent config SSM paramter', () => {
+    describe('CloudWatch Agent config SSM parameter', () => {
       test('configures log group', () => {
         // GIVEN
         const logGroup = target.node.findChild(`${DEFAULT_CONSTRUCT_ID}LogGroup`) as ILogGroup;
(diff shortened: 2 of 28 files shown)


packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
```diff
@@ -0,0 +1,634 @@
...
+
+        match = SECRET_ARN_RE.match(value)
+        if match:
+            named_groups = match.groupdict()
+            return AwsSecret( arn=value,region=named_groups["Region"] )
+
+        match = FILE_URI_RE.match( value )
+        if match:
+            named_groups = match.groupdict()
+            return FileSecret( arn=named_groups['FilePath'] )
```
jericht:
```suggestion
            return FileSecret( file_path=named_groups['FilePath'] )
```

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
```diff
@@ -0,0 +1,287 @@
...
+} from '.';
+import { tagConstruct } from './runtime-info';
+
+
+/**
+ * Properties for constructing a `DeploymentInstance`
+ */
+export interface DeploymentInstanceProps {
+  /**
+   * The time CloudFormation should wait for the success signals before failing the create/update.
```
horsmand:
```suggestion
   * The amount of time that CloudFormation should wait for the success signals before failing the create/update.
```

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
```diff
@@ -0,0 +1,287 @@
...
+  /**
+   * The log group name for streaming CloudWatch logs
+   *
+   * @default the construct ID is used
+   */
+  readonly logGroupName?: string;
+
+  /**
+   * Properties for setting up the Deadline DeploymentInstance's LogGroup in CloudWatch
+   * @default - LogGroup will be created with all properties' default values to the LogGroup: /renderfarm/<construct id>
```
horsmand: To keep a similar style to the rest of the props:

```suggestion
   * Properties for setting up the Deadline DeploymentInstance's LogGroup in CloudWatch
   *
   * @default the LogGroup will be created with all properties' default values to the LogGroup: /renderfarm/<construct id>
```

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
```diff
@@ -0,0 +1,287 @@
...
+   * Whether the instance should self-terminate after the deployment succeeds
+   *
+   * @default true
+   */
+  readonly selfTerminate?: boolean;
+
+  /**
+   * The VPC that the instance should be launched in.
+   */
+  readonly vpc: IVpc;
```
horsmand: I think for the most part we try to put the required fields at the top of the interface

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
```diff
@@ -0,0 +1,287 @@
...
+      signals: Signals.waitForAll({
+        timeout: props.executionTimeout ?? DeploymentInstance.DEFAULT_EXECUTION_TIMEOUT,
+      }),
+      updatePolicy: UpdatePolicy.replacingUpdate(),
+    });
+    this.node.defaultChild = this.asg;
+
+    this.connections = this.asg.connections;
+
+    const logGroupName = props.logGroupName ?? id;
```
horsmand, line 183: How obvious is it what the `id` belongs to if a user sees this log group in the CloudWatch console?

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -112,6 +123,19 @@ export interface IRenderQueue extends IConstruct, IConnectable {
...
+  /**
+   * Configure a rule to automatically register all Deadline Secrets Management identities connecting from a given
+   * subnet to a specified role and status.
+   *
+   * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html#identity-management-registration-settings-ref-label
+   * for details
+   *
+   * @param props Properties that specify thje configuration to be applied to the Deadline' Secrets Management identity
+   * registration settings. This specifies a VPC subnet and configures Deadline to automatically register identities of
+   * clients connecting from the subnet to a chosen Deadline Secrets Management role and status.
```
horsmand:
```suggestion
   * @param props Properties that specify the configuration to be applied to the Deadline' Secrets Management identity
   * registration settings. This specifies a VPC subnet and configures Deadline to automatically register identities of
   * clients connecting from the subnet to a chosen Deadline Secrets Management role and status.
```

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -894,4 +948,47 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+    } else {
+      throw new Error(`Unexpected type for ${deploymentInstanceNode.node.path} (${typeof(deploymentInstanceNode)})`);
+    }
+  }
+
+  /**
+   * The construct that manages Deadline Secrets Management identity registration settings
+   */
+  private get identityRegistrationSettings(): SecretsManagementIdentityRegistration {
+    const secretsManagementIdentityRegistration = this.node.tryFindChild('SecretsManagementIdentityRegistration');
```
horsmand: We could pull this string out like we did for the `deploymentInstance`.

packages/aws-rfdk/lib/deadline/lib/secrets-management-ref.ts
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * The Deadline Secrets Management registration status to be applied to the Deadline Client identities that connect
+   * from the specified VPC subnets.
+   *
+   * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html#registration-status
+   */
+  readonly registrationStatus: SecretsManagementRegistrationStatus;
+
+  /**
+   * The role to assign to be assigned to the Deadline Client identities that connect from the specified VPC subnets.
```
horsmand:
```suggestion
   * The role to be assigned to the Deadline Client identities that connect from the specified VPC subnets.
```

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
```diff
@@ -0,0 +1,262 @@
...
+   */
+  readonly repository: IRepository;
+
+  /**
+   * The version of the Deadline Client to use for performing the identity registration settings commands
+   */
+  readonly version: IVersion;
+}
+
+export class SecretsManagementIdentityRegistration extends Construct {
```
horsmand, line 102: This class is missing a docstring. I believe we need resources deployed and security considerations sections? Not sure if it warrants a diagram, probably not.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -894,4 +948,47 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+    if (!secretsManagementIdentityRegistration) {
+      return new SecretsManagementIdentityRegistration(
+        this, 'SecretsManagementIdentityRegistration', {
+          deploymentInstance: this.deploymentInstance,
+          repository: this.repository,
+          renderQueueSubnets: this.props.vpc.selectSubnets(
+            this.props.vpcSubnetsAlb ?? RenderQueue.DEFAULT_VPC_SUBNETS_ALB,
```
horsmand, line 988: Do you think using the default subnets justifies a warning message saying that their registration settings are likely too permissive?

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
```diff
@@ -0,0 +1,262 @@
...
+            // Admin credentials
+            '--credentials',
+            `"${this.adminCredentials.secretArn}"`,
+          ],
+          // Subnets of the load balancer
+          (
+            props.renderQueueSubnets
+              .subnetIds
+              .map(subnetID => `--connection-subnet "${subnetID}"`)
+          ),
```
horsmand, line 192: What are the extra parentheses for around this?

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
```diff
@@ -0,0 +1,262 @@
...
+          Array.from(this.subnetRegistrations.entries())
+            // Each setting becomes a comma (,) separated string of fields
+            //   <SUBNET_ID>,<ROLE>,<REGISTRATIONSTATUS>
```
horsmand:
```suggestion
            //   <SUBNET_ID>,<ROLE>,<REGISTRATION_STATUS>
```

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
```diff
@@ -0,0 +1,262 @@
...
+            .map(subnetRegistrationEntry => {
+              const [subnetID, registrationSettingEffect] = subnetRegistrationEntry;
+              return [
+                subnetID,
+                registrationSettingEffect.role.toString(),
+                (registrationSettingEffect.registrationStatus).toString(),
+              ].join(',');
+            })
+            // quote
+            .map(joinedSubnetArgValue => `--source-subnet "${joinedSubnetArgValue}"`),
```
horsmand: Not sure what this comment is for.

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
```diff
@@ -0,0 +1,262 @@
...
+        });
+      }
+    });
+    addSubnetProps.dependent.node.addDependency(this.deploymentInstance);
+  }
+
+  private runPythonScript(props: SecretsManagementIdentityRegistrationProps, localScriptFile: string) {
+    // The command-line arguments to be passed to the script that configures the Deadline identity registration
+    // settings
+    const scriptArgs = Lazy.list({
```
horsmand, line 176: I had made a comment on the render queue about a warning if they're using the default subnets for the ALB, but maybe we could have a warning here instead if there is overlap between the subnets used for `--connection-subnet` and those in `--source-subnet`.

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
```diff
@@ -0,0 +1,607 @@
...
+    source_subnets = config.source_subnet # type: List[SourceSubnet]
+
+    # Validate that source subnets are unique
+    observed_subnets = set()
+    for source_subnet in source_subnets:
+        subnet_id = source_subnet.subnet_id
+        if subnet_id not in observed_subnets:
+            observed_subnets.add(subnet_id)
+        else:
+            raise ValueError(f"Subnet \"{subnet_id}\" is not unique")
```
horsmand, line 215: This shouldn't be able to happen, but would it be better just to print a message and skip a duplicate?

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
```diff
@@ -0,0 +1,262 @@
...
+  private installDeadlineClient(props: SecretsManagementIdentityRegistrationProps) {
+    props.version.linuxInstallers.client.s3Bucket.grantRead(
+      props.deploymentInstance,
+      props.version.linuxInstallers.client.objectKey,
+    );
+    const clientInstallerPath = props.deploymentInstance.userData.addS3DownloadCommand({
+      bucket: props.version.linuxInstallers.client.s3Bucket,
+      bucketKey: props.version.linuxInstallers.client.objectKey,
+    });
+    props.deploymentInstance.userData.addCommands('set -x');
```
horsmand, line 268: Will this cause it to print passwords anywhere? I'm guessing no since the Deadline Commands get run from in the python script, but just want to confirm.

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
```diff
@@ -0,0 +1,607 @@
...
+
+        :return: the string path to the Deadline executable.
+        """
+
+        deadline_bin = os.environ.get('DEADLINE_PATH', '')
+
+        # On Linux, the Deadline Client installer creates a system-wide script to set the DEADLINE_PATH environment
+        # variable. Cloud-init does not load system environment variables. Cherry-pick the
+        # environment variable installed by the Deadline Client installer.
+        if not deadline_bin and os.path.exists(DL_ENV_SCRIPT_PATH):
```
horsmand: Maybe we should rename this variable to have `LINUX` in the name and extract out the OSX path into a variable as well.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -949,6 +954,12 @@ test('repository configure client instance', () => {
   // Make sure we call the configureRepositoryDirectConnect script with appropriate argument.
   const regex = new RegExp(escapeTokenRegex('\'/tmp/${Token[TOKEN.\\d+]}${Token[TOKEN.\\d+]}\' \\"/mnt/repository/DeadlineRepository\\"'));
   expect(userData).toMatch(regex);
+
+  // TODO: Add test to assert the IAM instance profile is given read access to the database credentials secret
```
horsmand: Is this something we need to do later?

packages/aws-rfdk/lib/deadline/test/secrets-management.test.ts
```diff
@@ -0,0 +1,747 @@
...
+                ],
+              )),
+            ],
+          ],
+        }];
+        expect(resolvedCalls).toContainEqual(expectedCall);
+      });
+
+      // eslint-disable-next-line jest/no-commented-out-tests
+      // test('normalizes subnet ID order', () => {
```
horsmand: How come this test is commented out? I can't understand why subnet ID order matters.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+
+describe('DeploymentInstance', () => {
+  let app: cdk.App;
+  let depStack: cdk.Stack;
+  let stack: cdk.Stack;
+  let vpc: Vpc;
+  let target: DeploymentInstance;
+  let deploymentInstanceProps: DeploymentInstanceProps;
+
+  beforeEach(() => {
```
horsmand: Running this setup method before each test and having tests only assert on one field seems like a heavyweight way to do testing. Is there any opportunity to switch some of these `beforeEach()` setup methods to `beforeAll()` methods?

horsmand (changes requested):
Overall this looks really good, none of my comments should require changing too much in the overall architecture of this. I think the deployment instance could be a really powerful tool.

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
jusiskin: I agree. I'm not a fan of how the log group name's suffix is derived from the construct ID. I'm following the pattern RFDK has followed so far. Here are a few example constructs that do the same:

*   `Repository`

    https://github.com/aws/aws-rfdk/blob/8845f4787cd7761f4b94213a7577ad6745d4246e/packages/aws-rfdk/lib/deadline/lib/repository.ts#L659
*   `RenderQueue`

    https://github.com/aws/aws-rfdk/blob/8845f4787cd7761f4b94213a7577ad6745d4246e/packages/aws-rfdk/lib/deadline/lib/render-queue.ts#L409-L412
*   `UsageBasedLicensing`

    https://github.com/aws/aws-rfdk/blob/8845f4787cd7761f4b94213a7577ad6745d4246e/packages/aws-rfdk/lib/deadline/lib/usage-based-licensing.ts#L553
*   `WorkerInstanceFleet`

    https://github.com/aws/aws-rfdk/blob/8845f4787cd7761f4b94213a7577ad6745d4246e/packages/aws-rfdk/lib/deadline/lib/worker-fleet.ts#L512-L521
*   `SpotEventPluginFleet`

    https://github.com/aws/aws-rfdk/blob/8845f4787cd7761f4b94213a7577ad6745d4246e/packages/aws-rfdk/lib/deadline/lib/spot-event-plugin-fleet.ts#L461-L474

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Good idea! I'll add this

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: While this class is exported by the module, it is not exported from the package so it is internal. That being said, a docstring is still a good idea, so I'll add one.

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: Interesting decision. I'll think about this and try to arrive at the best approach. I agree that we can and should emit warnings in certain situations.

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: This was done for horizontally aligning the commas in the outer list. What do you think? Does it help or hinder readability?

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: I've confirmed this does not by reviewing CloudWatch logs. Our repository direct connection script runs in a bash sub-process and that securely handles the password.

Also, the SM admin password used by `packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py` is put in the `DL_SM_PW` environment variable and we use the `--password env:DL_SM_PW` method to avoid passing the password with command-line arguments.

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
jusiskin: I'm not so sure... If two options for a subnet are passed with different effects, we aren't sure which one is actually desired/intended by the user. Since this is a user error and we can catch a mistake on their part that has impact on their security posture, I think it is safer to fail in such a scenario.

What do you think?

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+   * The log group name for streaming CloudWatch logs
+   *
+   * @default the construct ID is used
+   */
+  readonly logGroupName?: string;
+
+  /**
+   * Properties for setting up the Deadline DeploymentInstance's LogGroup in CloudWatch
```
jericht:
```suggestion
   * Properties for setting up the DeploymentInstance's LogGroup in CloudWatch
```

nit: looks like a leftover `Deadline` we can remove since this class is generalized

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
```diff
@@ -0,0 +1,288 @@
...
+   */
+  protected readonly asg: AutoScalingGroup;
+
+  constructor(scope: Construct, id: string, props: DeploymentInstanceProps) {
+    super(scope, id);
+
+    this.asg = new AutoScalingGroup(this, 'ASG', {
+      instanceType: props.instanceType ?? InstanceType.of(InstanceClass.T3, InstanceSize.SMALL),
+      keyName: props.keyName,
+      machineImage: props.machineImage ?? new AmazonLinuxImage(),
```
jericht:
```suggestion
      machineImage: props.machineImage ?? MachineImage.latestAmazonLinux(),
```

nit: I think calling `latestAmazonLinux()` here makes this is a bit easier to read since the default constructor for `AmazonLinuxImage`  [uses the latest one anyway](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.AmazonLinuxImage.html), which wasn't obvious to me at first glance. What do you think?

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
```diff
@@ -0,0 +1,288 @@
...
+   *
+   * @param dependency The construct that should be dependended upon
+   */
+  public addExecutionDependency(dependency: any): void {
+    if (Construct.isConstruct(dependency)) {
+      this.asg.node.defaultChild!.node.addDependency(dependency);
+    }
+  }
+
+  public get osType() {
```
jericht, line 209:
```suggestion
  /** 
   * @inheritdoc
   */
  public get osType() {
```

nit: This getter and the one below are inheriting their docstring right?

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+   * Adds UserData commands to configure the CloudWatch Agent running on the deployment instance.
+   *
+   * The commands configure the agent to stream the following logs to a new CloudWatch log group:
+   *   - The cloud-init log
+   *
+   * @param asg The auto-scaling group
+   * @param logGroupProps The properties for LogGroupFactory to create or fetch the log group
```
jericht, line 236: nit: missing `@param` tag for `groupName`. This is a private method though, so not necessary. We can even leave out this entire comment if we want.

packages/aws-rfdk/lib/deadline/lib/secrets-management-ref.ts
```diff
@@ -0,0 +1,106 @@
...
+  CLIENT = 'Client',
+};
+
+/**
+ * The different possible Deadline Secrets Management registration statuses that a Deadline Client's identity can be set
+ * to.
+ *
+ * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html#registration-status
+ */
+export enum SecretsManagementRegistrationStatus {
```
kozlove-aws, line 55: What the point to have these statuses? Do we expect users will use something except `REGISTERED`?

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
```diff
@@ -0,0 +1,607 @@
...
+        source_subnet.subnet_id for source_subnet in source_subnets
+    )
+    subnet_to_cidr = get_subnet_cidrs(config.region, all_subnets)
+    print('subnet_to_cidr = ' + json.dumps(subnet_to_cidr, indent=4))
+
+    dl_secrets = DeadlineSecretsCommandClient(credentials['username'], credentials['password'])
+
+    prior_rfdk_settings = get_rfdk_registration_settings(dl_secrets)
+
+    delete_removed_settings(dl_secrets, prior_rfdk_settings, connection_subnets, source_subnets)
```
kozlove-aws, line 590: Am I understand right that settings that created manually will be removed?

kozlove-aws (commented):
Great job. Just a couple moments that I want clarify.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+          expect(splitArgs[0]).toEqual('-i');
+          // Should pass the region
+          expect(stack.resolve(splitArgs[1])).toEqual(stack.resolve(stack.region));
+          // Should pass the SSM parameter containing the CloudWatch Agent configuration
+          expect(stack.resolve(splitArgs[2])).toEqual(stack.resolve(cloudWatchConfigSsmParam.parameterName));
+        });
+
+      });
+
+      test('Uses created Security Group', () => {
```
jericht, line 186: Is this test necessary? Based on our code in `DeploymentInstance`, it doesn't look like we're doing anything fancy with the security group, so it seems like this test is actually testing the CDK construct, not ours.

Same question for the `depends on policy` test below

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
jusiskin: The code uses a RFDK-specific convention for naming the identity registration settings. The settings have a name convention of:

```
RfdkSubnet|<CONNECTING_SUBNET_ID>|<SOURCE_SUBNET_ID>
```

This code will only remove settings that:

1.  match the above naming convention
2.  do not match the connecting/source subnets supplied in the command-line arguments 

So the only scenario where it would delete a manually created identity registration setting is if a user created one that matches the RFDK naming convention.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+              stack.resolve(policy.logicalId),
+            ),
+          },
+          ResourcePart.CompleteDefinition,
+        ));
+      });
+    });
+
+    describe('Security Group', () => {
+      test('creates Security Group in the desired VPC', () => {
```
jericht, line 225: Same question here, is this testing CDK logic? Maybe we can omit this test

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+        // GIVEN
+        instanceRole = (
+          target
+            .node.findChild('ASG')
+            .node.findChild('InstanceRole')
+            .node.defaultChild
+        ) as iam.CfnRole;
+      });
+
+      test('creates an instance profile', () => {
```
jericht, line 247: Same thing here, do we need this test and the `creates a role that can be assumed by EC2` test below?

packages/aws-rfdk/lib/deadline/lib/secrets-management-ref.ts
jusiskin: I made all possible identity registration statuses available that Deadline supports. RFDK users should not need to interact with this part of the API, since the constructs will automatically work together to configure the identity registration settings.

The only way a user could use this API is if they called `RenderQueue.configureSecretsManagementAutoRegistration(...)` directly. On that note, I'll document that fact on the method's TSDoc header.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+
+      test('can scale the Auto-Scaling Group', () => {
+        // THEN
+        expectCDK(stack).to(haveResourceLike('AWS::IAM::Policy', {
+          PolicyDocument: {
+            Statement: arrayWith(
+              {
+                Action: 'autoscaling:UpdateAutoScalingGroup',
+                Condition: {
+                  StringEquals: { 'autoscaling:ResourceTag/resourceLogicalId': cdk.Names.uniqueId(target) },
```
jericht, line 443: Do you think it would be helpful to leave a comment here saying that the key `autoscaling:ResourceTag/resourceLogicalId` is set by the `DeploymentInstance` class? Specifically the `resourceLogicalId` part. At first glance, it kind of looks like this could be a tag created by AWS or CDK and not our code.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+            ),
+          },
+          Roles: [
+            stack.resolve(instanceRole.ref),
+          ],
+        }));
+      });
+    });
+
+    describe('CloudWatch Agent config SSM paramter', () => {
```
jericht:
```suggestion
    describe('CloudWatch Agent config SSM parameter', () => {
```

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+        // defaults expected final log group name
+        `/renderfarm/${DEFAULT_CONSTRUCT_ID}`,
+      ],
+      [
+        { logGroupName: 'foo' },
+        // expected final log group name
+        '/renderfarm/foo',
+      ],
+      [
+        { logGroupName: 'bar' },
```
jericht: Can we remove this test case? Looks like this has the same coverage as the previous `foo` case.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+
+      // THEN
+      expectCDK(stack).to(haveResourceLike('Custom::LogRetention', {
+        LogGroupName: expectedLogGroupName,
+      }));
+    });
+  });
+
+  // GIVEN
+  describe.each<[InstanceType | undefined]>([
```
jericht: Does this need to be its own `describe`  block with multiple test cases? We have the default test `uses t3.small` which asserts the default instance type, and the last two test cases cover the same code paths. Can we simplify this into a single test that asserts that the ASG uses the provided instance type?

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+    });
+
+    test('does not add on-exit commands', () => {
+      // THEN
+      expect(target.userData.addOnExitCommands).not.toHaveBeenCalledWith(expect.arrayContaining([
+        expect.stringMatching(/\baws\s+.*\bautoscaling\s+update-auto-scaling-group/),
+      ]));
+    });
+
+    test('is not grante IAM permissions to scale the Auto-Scaling Group', () => {
```
jericht:
```suggestion
    test('is not granted IAM permissions to scale the Auto-Scaling Group', () => {
```

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
kozlove-aws: Right, I missed that we have strong name filter for these settings.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
```diff
@@ -0,0 +1,763 @@
...
+      // THEN
+      const matcher = resourceTagMatcher('AWS::AutoScaling::AutoScalingGroup', 'resourceLogicalId', cdk.Names.uniqueId(target));
+
+      // THEN
+      expectCDK(stack).notTo(haveResourceLike('AWS::AutoScaling::AutoScalingGroup', matcher));
+    });
+  });
+
+  // GIVEN
+  describe.each<[string]>([
```
jericht: Since we only have a single case here, do we need the `.each` from the `describe`? I don't think we will need to add more cases here in the future. Maybe we can use the value in the description text or rename it to `.executionTimeout is specified` or something

packages/aws-rfdk/lib/deadline/lib/secrets-management-ref.ts
kozlove-aws: I mean, since we do not need others statuses we can just use constant `REGISTERED` in python script and do not ask for status in `SubnetIdentityRegistrationSettingsProps` unless we want to register settings with different statuses.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -112,6 +123,19 @@ export interface IRenderQueue extends IConstruct, IConnectable {
...
   configureClientInstance(params: InstanceConnectOptions): void;
+
+  /**
+   * Configure a rule to automatically register all Deadline Secrets Management identities connecting from a given
+   * subnet to a specified role and status.
+   *
+   * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html#identity-management-registration-settings-ref-label
+   * for details
+   *
+   * @param props Properties that specify the configuration to be applied to the Deadline' Secrets Management identity
```
jericht:
```suggestion
   * @param props Properties that specify the configuration to be applied to the Deadline Secrets Management identity
```

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -894,4 +948,47 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+   * The instance that runs commands during the deployment.
+   */
+  private get deploymentInstance(): DeploymentInstance {
+    const CONFIGURE_REPOSITORY_CONSTRUCT_ID = 'ConfigureRepository';
+    const deploymentInstanceNode = this.node.tryFindChild(CONFIGURE_REPOSITORY_CONSTRUCT_ID);
+    if (deploymentInstanceNode === undefined) {
+      return new DeploymentInstance(this, CONFIGURE_REPOSITORY_CONSTRUCT_ID, {
+        vpc: this.props.vpc,
+        vpcSubnets: {
+          subnetType: SubnetType.PUBLIC,
```
jericht: Maybe a naive question, but why do we need to put this in a public subnet instead of a private one?

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -894,4 +948,47 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+        },
+      });
+    } else if (deploymentInstanceNode instanceof DeploymentInstance) {
+      return deploymentInstanceNode;
+    } else {
+      throw new Error(`Unexpected type for ${deploymentInstanceNode.node.path} (${typeof(deploymentInstanceNode)})`);
```
jericht: This error will most likely never happen, but would it help to put the expected type here?

```suggestion
      throw new Error(`Unexpected type for ${deploymentInstanceNode.node.path}. Expected ${DeploymentInstance.name}, but found ${typeof(deploymentInstanceNode)}.`);
```

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
```diff
@@ -894,4 +948,47 @@ export class RenderQueue extends RenderQueueBase implements IGrantable {
...
+          renderQueueSubnets: this.props.vpc.selectSubnets(
+            this.props.vpcSubnetsAlb ?? RenderQueue.DEFAULT_VPC_SUBNETS_ALB,
+          ),
+          version: this.props.version,
+        },
+      );
+    } else if (secretsManagementIdentityRegistration instanceof SecretsManagementIdentityRegistration) {
+      return secretsManagementIdentityRegistration;
+    } else {
+      throw new Error(`Unexpected type for ${secretsManagementIdentityRegistration.node.path} (${typeof(secretsManagementIdentityRegistration)})`);
```
jericht: Same thing here, should we add the expected type?
```suggestion
      throw new Error(`Unexpected type for ${secretsManagementIdentityRegistration.node.path}. Expected ${SecretsManagementIdentityRegistration.name}, but found ${typeof(secretsManagementIdentityRegistration)}.`);
```

packages/aws-rfdk/lib/deadline/lib/secrets-management-ref.ts
```diff
@@ -0,0 +1,106 @@
...
+   * secrets.
+   */
+  ADMINISTRATOR = 'Administrator',
+
+  /**
+   * The Server role is intended to be granted to your machine(s) running the
+   * [Remote Connection Server](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/remote-connection-server.html#remote-connection-server-ref-label)
+   * application. The Server role is granted to a registered machine by an administrator in the
+   * [Monitor UI](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html#identity-management-assigning-ref-label).
+   * In order to encrypt and decrypt secrets, the master key must be assigned to the Server by and Administrator user
```
jericht:
```suggestion
   * In order to encrypt and decrypt secrets, the master key must be assigned to the Server by an Administrator user
```

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
```diff
@@ -0,0 +1,262 @@
...
+  /**
+   * The Deadline Secrets Management registration status to be applied to the Deadline Client identities that connect
+   * from the specified VPC subnets.
+   *
+   * See https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/secrets-management/deadline-secrets-management.html#registration-status
+   */
+  readonly registrationStatus: SecretsManagementRegistrationStatus;
+
+  /**
+   * The role to assign to be assigned to the Deadline Client identities that connect from the specified VPC subnets.
```
jericht:
```suggestion
   * The role to be assigned to the Deadline Client identities that connect from the specified VPC subnets.
```

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
```diff
@@ -0,0 +1,262 @@
...
+  SecretsManagementRole,
+  SecretsManagementRegistrationStatus,
+  SubnetIdentityRegistrationSettingsProps,
+} from '.';
+import {
+  ScriptAsset,
+} from '../../core';
+import { DeploymentInstance } from '../../core/lib/deployment-instance';
+
+interface RegistrationSettingStack {
```
jericht: nit: missing docstring. This is not a necessary change since this interface isn't exported, but might be helpful to people that read this, especially since all the fields below mention subnets, so it'd be nice to have something here to glue them all together

packages/aws-rfdk/lib/deadline/lib/stage.ts
```diff
@@ -228,4 +229,39 @@ export class Stage {
...
+      const version = match!.groups!.version;
+      const recipes = Array.from(Object.values(this.manifest.recipes));
+      const aRecipeHasMatchingDlVersion = recipes.some((recipe) => {
+        return recipe.buildArgs?.DL_VERSION === version;
+      });
+      const installerPath = path.join(
+        this.dirPath,
+        'bin',
+        listing[0],
+      );;
```
jericht:
```suggestion
      );
```

packages/aws-rfdk/lib/deadline/test/usage-based-licensing.test.ts
```diff
@@ -103,29 +110,71 @@ describe('UsageBasedLicensing', () => {
...
+        vpcSubnets: { subnetType: SubnetType.PRIVATE },
+      };
+
+      // THEN
+      expect(renderQueue.configureSecretsManagementAutoRegistration).toHaveBeenCalledWith(expectedCall);
+    });
+
+    test.each<[SubnetSelection]>([
+      [{
+        subnetType: SubnetType.PUBLIC,
```
jericht, line 173: Do we need to use `test.each` here? Can we simplify this to just be a single test?

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
```diff
@@ -0,0 +1,607 @@
...
+    return bool(RFDK_IDENTITY_REGISTRATION_SETTING_NAME_RE.search(setting.settings_name))
+
+
+def get_rfdk_registration_settings(dl_secrets: DeadlineSecretsCommandClient) -> List[LoadBalancerIdentityRegistrationSetting]:
+    all_registration_settings = [
+        LoadBalancerIdentityRegistrationSetting.from_json(registration_setting)
+        for registration_setting in dl_secrets.run_json('GetLoadBalancerIdentityRegistrationSettings')
+    ]
+
+    print('all registration settings = ', end='')
```
jericht, line 385: Are these debug statements that we should remove? or do we want to keep this?

jericht (changes requested):
Looks great! Just have a few minor comments

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
horsmand: You have a good point that it's unclear which duplicate entry to keep. I was thinking that since we're only ever setting the registration settings to configure identities as `registered` with the `client` role, then the entries are identical, but it does have the capability to set entries to other values than `registered` and `client`. I think we leave it as-is.

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
horsmand: I'm not sure what you mean by aligning the commas in the outer list. It might align it with the array above it, but it doesn't look like it lines up with the `Array.from()` item below it. Also I'm not sure if there should be a dangling comma on the line with the `.map()`.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: Keep in mind that using `beforeAll()` to create partial CDK construct trees means that tests would share the construct tree. A few reasons below why this is problematic/challenging:

1.  some tests/fixtures mutate the construct tree, so when a test calls `expectCDK(...).to(...)`, then we might get results that are confounded by inter-test pollution
2.  we'll run into duplicate construct ID errors unless care is taken to ensure unique IDs between each test code path.
3.  anyone working on these tests in the future will need to be aware that the tests use a shared construct tree. The difference between `beforeEach` and `beforeAll` is subtle and someone adding a test might not realize this and waste time debugging their tests

I refactored the tests to use `beforeAll(...)` everywhere and then timed 20 exeuctions of each implementation with:

```sh
for x in $(seq 1 20); do
    yarn jest lib/core/test/deployment-instance.test.js  |& egrep "^Time:" | awk '{ print $2 }'
done
```

Here are the results on my dev machine:

| Iteration | `beforeAll(...)` (seconds) | `beforeEach(...)` (seconds) |
| --------- | ---------------- | ----------------- |
| 1 | 4.946 | 4.805 |
| 2 | 4.568 | 4.779 |
| 3 | 4.538 | 4.854 |
| 4 | 4.506 | 4.869 |
| 5 | 4.525 | 4.83 |
| 6 | 4.668 | 4.85 |
| 7 | 4.556 | 5.037 |
| 8 | 4.542 | 4.761 |
| 9 | 4.764 | 4.961 |
| 10 | 4.582 | 4.959 |
| 11 | 4.94 | 4.81 |
| 12 | 4.666 | 4.847 |
| 13 | 4.917 | 4.803 |
| 14 | 4.494 | 4.894 |
| 15 | 4.559 | 4.851 |
| 16 | 4.599 | 5.214 |
| 17 | 4.639 | 4.816 |
| 18 | 4.733 | 4.834 |
| 19 | 4.987 | 5.121 |
| 20 | 4.533 | 4.88 |

Granted, these times are likely governed by loading up the node, jest, and the dependent code, JIT compiling it all, etc.. 

The statistical distributions for each are:

| Implementation | mean | stddev |
| --- | --- | --- |
| `beforeAll` | 4.6631 | 0.1627701 |
| `beforeEach` | 4.88875 | 0.1167336 |

The difference between means is ~220ms. Compounded that can add up. 100 runs of that test would mean 22seconds of saved compute in the `beforeAll` case.

Do you think that this is worth the trade-offs mentioned above?

packages/aws-rfdk/lib/deadline/test/secrets-management.test.ts
jusiskin: Initially I'd thought the order needed to be normalized because I'd thought that changes to the CDK code structure that change the call order to `RenderQueue.configureSecretsManagementAutoRegistration(...)` would cause the user data to change and trigger the configuration to happen again unnecessarily.

This is an optimization and not strictly required for the feature to work correctly. It also turned out that ordering subnet IDs is not always possible when they are unresolved CDK tokens. For this reason, I removed that from the implementation.

I commented out the test and forgot to remove it completely before pushing it up. The next revision will have this test removed.

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
jusiskin: Done

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: agreed. I've made the change

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: Added a TSDoc string. Let me know how it looks.

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: The map doesn't require a comma dangle because it is the contents of bracket-ed expression.

I've made a pass at fixing the alignment. Let me know if it looks acceptable to you.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Applied this suggestion

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: Oops, looks like was not updated after a change during implementation. I've revised the comment. Let me know what you think now.

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: I added this check among some others. What do you think?

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
jusiskin: done

packages/aws-rfdk/lib/deadline/test/repository.test.ts
jusiskin: During implementation, I ran into an issue with being unable to connect to the database and thought that we'd missed this behavior. It actually turned out that the database credentials were stored only for the `ec2-user`. I've since added a code comment in the relevant spot, but since we were missing this test coverage, I've added it now.

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: I've refactored to use `beforeAll()` where possible

packages/aws-rfdk/lib/deadline/lib/secrets-management-ref.ts
jusiskin: Added a TSDoc string explaining that this API should not used by RFDK end-users unless you have a specific need. Let me know what you think.

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
jusiskin: The `FileSecret` was removed due to it being unused code.

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
jusiskin: good catch - fixed

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
jusiskin: sure! that reads better - applied now.

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
jusiskin: VSCode seems to think so when I hover over them, but I'm not sure whether our auto-generated docs will behave the same. I'll add it to be explicit.

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
jusiskin: I opted to add in the `@param`. Let me know what you think

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: If we remove the test, then we don't have anything asserting what security group the synthesized ASG uses.

I understand that we defer that logic to CDK, but shouldn't we still assert that we are using CDK's security group code? The other option for that would be mocking, and I opted for `cdk-assert`.

What do you think?

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: I added this now. Let me know what you think

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: fixed

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: Removed

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: yep good call. less code is better! done

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: Fixed - thanks for catching

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jusiskin: Good feedback. I've taken a pass at it. Let me know if it looks okay.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Good catch - applied

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Not naive at all - in fact this is a great catch! We need to lock down this instance since it has full access to:

*   Repository file-system
*   Repository database
*   Database credentials
*   Secrets management admin credentials secret in AWS Secrets Manager

This was changed to public for speeding up my dev/test cycle iterations and it looks like I forgot to set it back. I've changed it to use the same subnets passed into the render queue's props. I've added a test to make sure this doesn't regress.

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Applied

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jusiskin: Applied

packages/aws-rfdk/lib/deadline/lib/secrets-management-ref.ts
jusiskin: thanks - fixed

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: Fixed

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
jusiskin: This led me to realize that the `stack` field is unused. I renamed the struct `RegistrationSettingEffect` such it only contains the output fields of a Deadline Secrets Management identity registration setting (role and status).

Since the structure is independent of subnets, I'll keep the TSDoc string general. Added now so let me know what you think

packages/aws-rfdk/lib/deadline/lib/stage.ts
jusiskin: done

packages/aws-rfdk/lib/deadline/test/usage-based-licensing.test.ts
jusiskin: I've moved away from `.each` for this single test case.

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
jusiskin: These are pretty sparse, and I think they might be helpful for troubleshooting if any issues arise. What do you think though?

packages/aws-rfdk/lib/deadline/lib/secrets-management-ref.ts
kozlove-aws: Hm. Where is this new TSDoc string?

jusiskin: https://github.com/aws/aws-rfdk/pull/576/files#diff-7087a85eeaddfc36b08d14a6bb1536c2c5a6291bf28d1c9ec2a65e4597c6ec09R134-R136

kozlove-aws (dismissed):

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
horsmand: Looks good

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
horsmand: Yeah I like it better with the last item indented as well.

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
horsmand: I think it was straightforward enough not to need a comment, but it's fine as-is.

packages/aws-rfdk/lib/deadline/lib/secrets-management.ts
horsmand: Yep, you've covered everything that we recommend against. Looks good to me.

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
horsmand: I don't see any harm in them, this file is only ran once per deployment so the amount of logs is pretty small and they are helpful.

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
jericht: thanks!

packages/aws-rfdk/lib/core/lib/deployment-instance.ts
jericht: Looks good, thanks

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jericht: That's a good point. I wanted to make sure we weren't unnecessarily testing CDK code, but I guess this is just one of the nuances when working with CDK since we're asserting the generated CFN template and not directly asserting the logic in our code.

It doesn't hurt to have these extra tests in, so let's keep them. Resolving these comments

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jericht: I don't see any changes around here, did you forget to commit this change?

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jericht: awesome, thanks!

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jericht: Nevermind, I missed it elsewhere in the diff, resolving :)

jusiskin: Oops. I updated here:

https://github.com/aws/aws-rfdk/pull/576/files#diff-f29bdea96151f22e55d5bf9b6dacb16ac189706acb83b3607d8c4dbaea2c0024R701

but not at this line as well. I'll add that....

packages/aws-rfdk/lib/core/test/deployment-instance.test.ts
jericht: looks great, thanks!

packages/aws-rfdk/lib/deadline/lib/render-queue.ts
jericht: Awesome, thanks for clearing that up for me and adding the additional testing

packages/aws-rfdk/lib/deadline/scripts/python/configure_identity_registration_settings.py
jericht: Yeah, totally fine to leave them in if they are intended logging messages, just wanted to make sure they weren't left in from development. Resolving

packages/aws-rfdk/lib/deadline/test/secrets-management.test.ts
```diff
@@ -586,111 +587,30 @@ describe('SecretsManagementIdentityRegistration', () => {
...
-      //               }),
-      //             ],
-      //           ],
-      //         },
-      //       ],
-      //     ],
-      //   }];
-      //   expect(resolvedCalls).toContainEqual(expectedCall);
-      // });
+      test('warns about dedicated subnets when ALB and source subnets match', () => {
```
jericht:
```suggestion
      test('warns about dedicated subnets when render queue ALB and source subnets match', () => {
```

this is a really minor nit though, so I won't hold the PR back for it

packages/aws-rfdk/lib/deadline/test/usage-based-licensing.test.ts
```diff
@@ -121,6 +121,45 @@ describe('UsageBasedLicensing', () => {
     });
   }
 
+  describe('vpcSubnets', () => {
+    describe('specified', () => {
```
jericht: Do we need `describe` blocks for the `specified` and the other `not specified` blocks? Since we only have one test case in each block, maybe we can just have to tests in this `vpcSubnets` block. What do you think?

Another really minor comment though, not holding back the PR for it.

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
```diff
@@ -1943,3 +1944,71 @@ test('worker fleet does not signal when zero minCapacity', () => {
...
+        vpc,
+        version,
+      }),
+      version,
+    });
+    wfstack = new Stack(app, 'workerFleetStack');
+    props = {
+      renderQueue,
+      vpc,
+      workerMachineImage: new GenericWindowsImage({
```
jericht:
```suggestion
      workerMachineImage: new GenericWindowsImage(),
```

We can remove the empty props `{}` here

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
```diff
@@ -1943,3 +1944,71 @@ test('worker fleet does not signal when zero minCapacity', () => {
...
+      }),
+    };
+  });
+
+  describe('vpc subnets not specified', () => {
```
jericht: Same comment here, do we need these to be `describe` blocks or can they just be `test`?

jericht (dismissed):
LGTM!

I left a few minor comments, but it's mostly tiny formatting/organization in the unit tests, so not holding back the PR for it.

packages/aws-rfdk/lib/deadline/test/usage-based-licensing.test.ts
jericht: The diff is still showing this as using `test.each`, but this is a super minor comment, so resolving

packages/aws-rfdk/lib/deadline/test/configure-spot-event-plugin.test.ts
```diff
@@ -968,4 +969,100 @@ describe('ConfigureSpotEventPlugin', () => {
...
+          vpc,
+          version,
+        }),
+        version,
+      });
+
+      groupName = 'group_name1';
+    });
+
+    describe('a fleet without vpcSunbets', () => {
```
horsmand:
```suggestion
    describe('a fleet without vpcSubnets', () => {
```

horsmand (dismissed):
I'm in the same boat as Jericho. This looks good as-is. I'm going to give my approval but leave it for Josh to look over the comments and decide whether to fix them or merge.

packages/aws-rfdk/lib/deadline/test/secrets-management.test.ts
jusiskin: Fixed

packages/aws-rfdk/lib/deadline/test/usage-based-licensing.test.ts
jusiskin: I've flattened the test hierarchy. Let me know what you think

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
jusiskin: This is actually the AMI region map and its required. I've collapsed it onto a single line though which makes it looks a bit nicer

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
jusiskin: I've flattened this now

horsmand (approved):

packages/aws-rfdk/lib/deadline/test/usage-based-licensing.test.ts
jericht: Looks good!

packages/aws-rfdk/lib/deadline/test/worker-fleet.test.ts
jericht: Ah, I should've double checked the API. Thanks

jericht (approved):

PR #578, opened 2021-09-21
chore: Revert "Modified basic example to make a secrets management PoC"

horsmand:
## Summary

This reverts commit f99d4e3295135e8afd5cdacb67351a931cd7e5b8. The commit was for an RFDK app used for testing the security of the feature branch. It shouldn't have been added to the feature branch since it was never intended to be merged into mainline.

## Testing

The basic example app was deployed using a release candidate of Deadline with the changes required to support Secrets Management in RFDK and it was confirmed that Secrets Management was enabled and the RCS was configured as the server machine.

----


examples/deadline/All-In-AWS-Infrastructure-Basic/ts/bin/app.ts
```diff
@@ -131,19 +125,7 @@ new ComputeTier(app, 'ComputeTier', {
   vpc: network.vpc,
   renderQueue: service.renderQueue,
   workerMachineImage: MachineImage.genericLinux(config.deadlineClientLinuxAmiMap),
-  keyPairName: config.keyPairName,
+  keyPairName: config.keyPairName ? config.keyPairName : undefined,
```
kozlove-aws, line 128: what the point to change this line like this?

kozlove-aws (approved):
Validated that all changes were reverted. Looks good.

examples/deadline/All-In-AWS-Infrastructure-Basic/ts/bin/app.ts
horsmand: I think I probably just noticed that the ` ? config.keyPairName : undefined` part of the line doesn't actually do anything since the value is optional and will resolve to undefined anyways without the need for the ternary operator, so I deleted it. It should probably be removed in mainline, but it isn't a big deal.

jericht (approved):

PR #584, opened 2021-09-26
fix(deadline): deploy deadline secrets management admin credentials in database stack

kozlove-aws:
### Problem
In case when Secret with credentials for Secrets Management administrator is created in Repository constructor we have a situation when this Secret can be removed with Repository stack and user lose access to Secret Management.

### Solution
Secret that created in Repository construct now is created in database stack.

### Testing

 - Added unit tests
 - Deployed farm with last changes and enabled SM
 - Make sure Secret with credentials in Storage stack
 - Make sure that repository installed with SM enabled.
 - Delete Service tier and deploy it again.
 - Validate that secret was not changed and SM is enabled. 
 - Deployed farm with MongoDB and make sure that Secret management is enabled.
 - Try to deploy farm with imported MongoDB and received an error
 - Deployed farm with imported database and provided Secret ARN for credentials
(have found an error in repository installation log that admin user exists and cannot be added)
 - Validate that SM is enabled and admin credentials for SM in provided Secret is correct.
 - 
----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
@@ -15,6 +15,7 @@ import {
   ISecurityGroup,
   OperatingSystemType,
   Port,
+  Volume,
 } from '@aws-cdk/aws-ec2';
 import {
   IGrantable,
@@ -26,6 +27,7 @@ import {
 } from '@aws-cdk/aws-secretsmanager';
 import {
   Annotations,
+  Construct,
   IConstruct,
   Stack,
 } from '@aws-cdk/core';
@@ -38,7 +40,6 @@ import {
 import {
   IHost,
 } from './host-ref';
-
 
 /**
  * Options when constructing UserData for Linux
@@ -114,6 +115,11 @@ export abstract class DatabaseConnection {
   public abstract readonly containerEnvironment: { [name: string]: string };
 
   /**
+   * Holds the CDK Construct that contains the database, or undefined if the database was imported into this application.
+   */
+  public abstract readonly databaseConstruct?: Construct;
+
+  /**
    * Allow connections to the Database from the given connection peer
    */
   public abstract allowConnectionsFrom(other: IConnectable): void;
@@ -175,6 +181,11 @@ class DocDBDatabaseConnection extends DatabaseConnection {
    */
   public readonly containerEnvironment: { [name: string]: string };
 
+  /**
+   * @inheritdoc
+   */
+  public readonly databaseConstruct?: Construct;
+
   constructor(private readonly props: DocDBConnectionOptions) {
     super();
 
@@ -186,6 +197,10 @@ class DocDBDatabaseConnection extends DatabaseConnection {
       // The container must fetch the credentials from Secrets Manager
       DB_CREDENTIALS_URI: this.props.login.secretArn,
     };
+
+    if (props.database instanceof DatabaseCluster) {
+      this.databaseConstruct = props.database;
+    }
   }
 
   /**
@@ -335,7 +350,15 @@ class DocDBDatabaseConnection extends DatabaseConnection {
 class MongoDbInstanceDatabaseConnection extends DatabaseConnection {
   private static readonly DB_CERT_LOCATION: string = '/opt/Thinkbox/certs/mongo_client.pfx';
 
+  /**
+   * @inheritdoc
+   */
   public readonly containerEnvironment: { [name: string]: string };
+
+  /**
+   * @inheritdoc
+   */
+  public readonly databaseConstruct?: Construct;
 
   constructor(protected readonly props: MongoDbInstanceConnectionOptions) {
     super();
@@ -343,6 +366,12 @@ class MongoDbInstanceDatabaseConnection extends DatabaseConnection {
       DB_TLS_CLIENT_CERT_URI: props.clientCertificate.cert.secretArn,
       DB_TLS_CLIENT_CERT_PASSWORD_URI: props.clientCertificate.passphrase.secretArn,
     };
+
+    if (props.database instanceof MongoDbInstance) {
+      if (props.database.mongoDataVolume instanceof Volume) {
+        this.databaseConstruct = props.database.mongoDataVolume;
+      }
+    }
   }
 
   /**
--- a/packages/aws-rfdk/lib/deadline/lib/repository.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/repository.ts
@@ -308,6 +308,14 @@ export interface SecretsManagementProps {
    * @default: A random username and password will be generated in a Secret with ID `SMAdminUser` and will need to be retrieved from AWS Secrets Manager if it is needed
    */
   readonly credentials?: ISecret;
+
+  /**
+   * If Secret with admin credentials is not defined in property `credentials`, then this specifies the retention policy to
+   * use on the Secret with generated credentials. If Secret with credentials is provided, then this property is ignored.
+   *
+   * @default RemovalPolicy.RETAIN
+   */
+  readonly credentialsRemovalPolicy?: RemovalPolicy;
 }
 
 /**
@@ -461,6 +469,7 @@ export interface RepositoryProps {
  * - Instance Role and corresponding IAM Policy.
  * - An Amazon CloudWatch log group that contains the Deadline Repository installation logs.
  * - An RFDK PadEfsStorage - If no filesystem is provided.
+ * - An AWS Secrets Manager Secret - If no Secret with admin credentials for Deadline Secrets Management is provided.
  *
  * Security Considerations
  * ------------------------
@@ -581,23 +590,26 @@ export class Repository extends Construct implements IRepository {
     if (props.fileSystem instanceof MountableEfs && !props.fileSystem.accessPoint) {
       throw new Error('When using EFS with the Repository, you must provide an EFS Access Point');
     }
+    if ((props.secretsManagementSettings?.enabled ?? true) && props.database && !props.database.databaseConstruct) {
+      throw new Error('Admin credentials for Deadline Secrets Management cannot be generated when using an imported database. For setting up your own credentials, please refer to https://github.com/aws/aws-rfdk/tree/mainline/packages/aws-rfdk/lib/deadline#configuring-deadline-secrets-management-on-the-repository.');
+    }
 
     this.version = props.version;
 
     this.secretsManagementSettings = {
       enabled: props.secretsManagementSettings?.enabled ?? true,
       credentials: props.secretsManagementSettings?.credentials ??
-        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(this, 'SMAdminUser', {
+        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret( props.database?.databaseConstruct ? Stack.of(props.database?.databaseConstruct) : this, 'SMAdminUser', {
           description: 'Admin credentials for Deadline Secrets Management',
           generateSecretString: {
             excludeCharacters: '\"$&\'()/<>[\\]\`{|}',
             includeSpace: false,
             passwordLength: 24,
             requireEachIncludedType: true,
-
             generateStringKey: 'password',
             secretStringTemplate: JSON.stringify({ username: Repository.DEFAULT_SECRETS_MANAGEMENT_USERNAME }),
           },
+          removalPolicy: props.secretsManagementSettings?.credentialsRemovalPolicy ?? RemovalPolicy.RETAIN,
         }) : undefined),
     };
 
--- a/packages/aws-rfdk/lib/deadline/test/database-connection.test.ts
+++ b/packages/aws-rfdk/lib/deadline/test/database-connection.test.ts
@@ -24,6 +24,7 @@ import {
   MachineImage,
   SecurityGroup,
   SubnetType,
+  Volume,
   Vpc,
   WindowsVersion,
 } from '@aws-cdk/aws-ec2';
@@ -32,6 +33,7 @@ import {
   Role,
 } from '@aws-cdk/aws-iam';
 import {
+  IPrivateHostedZone,
   PrivateHostedZone,
 } from '@aws-cdk/aws-route53';
 import {
@@ -335,6 +337,13 @@ describe('DocumentDB', () => {
     ]));
   });
 
+  test('Document DB connection is pointed to correct construct', () => {
+    // GIVEN
+    const connection = DatabaseConnection.forDocDB({database, login: database.secret!});
+
+    // THEN
+    expect(connection.databaseConstruct).toEqual(database);
+  });
 });
 
 describe('DocumentDB Version Checks', () => {
@@ -390,10 +399,11 @@ describe('DocumentDB Version Checks', () => {
     });
 
     // WHEN
-    DatabaseConnection.forDocDB({database, login: secret});
+    const databaseConnection = DatabaseConnection.forDocDB({database, login: secret});
 
     // THEN
     expect(database.node.metadata.length).toBe(0);
+    expect(databaseConnection.databaseConstruct).toBeUndefined();
   });
 
   test('No engineVersion given', () => {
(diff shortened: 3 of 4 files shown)


packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -175,7 +181,12 @@ class DocDBDatabaseConnection extends DatabaseConnection {
    */
   public readonly containerEnvironment: { [name: string]: string };
 
-  constructor(private readonly props: DocDBConnectionOptions) {
+  /**
+   * @inheritdoc
+   */
+  public readonly database: Construct;
+
+  constructor(scope: Construct, private readonly props: DocDBConnectionOptions) {
```
jusiskin: This is a breaking API change. We should already have a reference to the stack of the database from `props.database`. Technically, it has a type of `IDatabase`, but can we use type-checking at run-time to determine whether this is the L2 construct vs the imported resource? So I'm thinking something like:

```ts
if (props.database instanceof DatabaseCluster) {
    // props.database is a L2 DatabaseCluster
    // ...
} else {
    // props.database is an imported DatabaseCluster
    // ...
}
```

what do you think?

jusiskin (changes requested):
I wonder if we can re-work this so that we don't have to change our `DatabaseConnection` API.

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
horsmand: The problem is that `props.database` is a `DatabaseConnection` and not a `DatabaseCluster`. There are 2 implementations of `DatabaseConnection` and neither expose the construct that represents the actual database.

1. `DocDBDatabaseConnection` stores its `props` as a private field and the `IDatabaseCluster` is the `database` field on that object.
2. `MongoDbInstanceDatabaseConnection` also stores its `props` as a private field, but the `database` field on that object is an `IMongoDb`.

What if we add the abstract method `isDatabaseImported` to `DatabaseConnection` that just returns true or false? I think the Mongo implementation will always return false, but the DocDB version would do the `instanceof` test to see if it's a class or interface.

horsmand: An alternative to my suggestion is the visitor pattern. It would be more extensible going forward, but it's a bit heavy-duty.

kozlove-aws: Thank you for advices. What about this version? Does it look better?

horsmand: You're not traversing down into the MongoDbInstanceConnection properly to determine if it's using an import or not, but otherwise this looks like it should work. However, I think the new public class variable to hold the database construct could be confusing to anybody using the DatabaseConnection class; it isn't obvious why it's there or why it might not be populated with anything. I guess if we documented it properly then it should be fine.


Determining if the MongoDB connection is an import might be something like this:
```
if (mongoDbConnection.database instanceof MongoDbInstance) {
  dbInstance = (MongoDbInstance) mongoDbConnection.database;
  if (dbInstance.mongoDataVolume instanceof Volume) {
    return dbInstance.mongoDataVolume;
  }
}
```

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -581,20 +581,22 @@ export class Repository extends Construct implements IRepository {
...
+      throw new Error('Admin credentials for Deadline Secrets Management cannot be generated when using imported database. Please provide secret with admin credentials or disable Secrets Management in repository properties.');
+    }
 
     this.version = props.version;
 
     this.secretsManagementSettings = {
       enabled: props.secretsManagementSettings?.enabled ?? true,
       credentials: props.secretsManagementSettings?.credentials ??
-        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(this, 'SMAdminUser', {
+        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret( props.database?.databaseConstruct ?? this, 'SMAdminUser', {
```
jusiskin: I'm not sure whether we want to put the secret in the database's scope or in the stack's scope. The alternative is:

```suggestion
        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(Stack.of(props.database?.databaseConstruct) ?? this, 'SMAdminUser', {
```

We need to commit to whatever choice  we make here, since changing scopes is a breaking change. The choice of scopes has two potential impacts that I'm aware of:

## Identifiers

In your current choice of scope, the secret's construct path derives from the construct path of the database. The logical IID for the secret becomes a compound value of the construct path ([ref](https://docs.aws.amazon.com/cdk/latest/guide/identifiers.html)). Depending on the implementation of the `AWS::SecretsManager::Secret` resource provider, it may use the logical ID to generate the secret name. This will affect how readable the secret name is in the AWS console and API responses. That being said, if there may be some value to knowing which database the secret belongs to.

## Aspects

[CDK's aspects](https://docs.aws.amazon.com/cdk/latest/guide/aspects.html) can be selectively applied to a node in the construct tree and its descendents. Users intending to walk the construct nodes that are children of the database will also visit the secret's construct nodes.  This may depend on the call order, but it is worth some consideration.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -581,20 +581,22 @@ export class Repository extends Construct implements IRepository {
     if (props.fileSystem instanceof MountableEfs && !props.fileSystem.accessPoint) {
```
jusiskin, line 590: I noticed that we also don't document the admin credentials secret being deployed in the "Resources Deployed" section of the `Repository` construct's TSDoc header. Do you mind adding that as part of this PR?

jusiskin (changes requested):
I just noticed that we missed documenting the secret as a resource deployed.

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -336,13 +351,20 @@ class MongoDbInstanceDatabaseConnection extends DatabaseConnection {
   private static readonly DB_CERT_LOCATION: string = '/opt/Thinkbox/certs/mongo_client.pfx';
 
   public readonly containerEnvironment: { [name: string]: string };
+  public readonly databaseConstruct?: Construct;
```
horsmand: Can you add the `@inheritdoc` on both of these variables?

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -113,6 +114,11 @@ export abstract class DatabaseConnection {
    */
   public abstract readonly containerEnvironment: { [name: string]: string };
 
+  /**
+   * Returns the environment variables for configuring containers to connect to the database
```
horsmand: Copy
```suggestion
   * Holds the CDK Construct that contains the database, or undefined if the database was imported into this application.
```

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -336,13 +351,20 @@ class MongoDbInstanceDatabaseConnection extends DatabaseConnection {
...
   constructor(protected readonly props: MongoDbInstanceConnectionOptions) {
     super();
     this.containerEnvironment = {
       DB_TLS_CLIENT_CERT_URI: props.clientCertificate.cert.secretArn,
       DB_TLS_CLIENT_CERT_PASSWORD_URI: props.clientCertificate.passphrase.secretArn,
     };
+
+    if (props.database instanceof MongoDbInstance) {
+      if (props.database.mongoDataVolume instanceof Volume) {
+        this.databaseConstruct = props.database.mongoDataVolume;
```
horsmand, line 372: How have we tested that the `mongoDataVolume` is the correct Construct to track and not the `server`? Have we tried deploying a farm with a MongoDB and Secrets Management enabled?

Also I don't see any unit tests for these changes similar to the updated and added DocDB tests.

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -461,6 +461,7 @@ export interface RepositoryProps {
  * - Instance Role and corresponding IAM Policy.
  * - An Amazon CloudWatch log group that contains the Deadline Repository installation logs.
  * - An RFDK PadEfsStorage - If no filesystem is provided.
+ * - An Amazon Secret - If no Secret with admin credetials for Secrets Management is provided.
```
horsmand:
```suggestion
 * - An AWS Secrets Manager Secret - If no Secret with admin credentials for Deadline Secrets Management is provided.
```

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -581,23 +582,26 @@ export class Repository extends Construct implements IRepository {
     if (props.fileSystem instanceof MountableEfs && !props.fileSystem.accessPoint) {
       throw new Error('When using EFS with the Repository, you must provide an EFS Access Point');
     }
+    if ((props.secretsManagementSettings?.enabled ?? true) && props.database && !props.database.databaseConstruct) {
+      throw new Error('Admin credentials for Deadline Secrets Management cannot be generated when using imported database. Please provide secret with admin credentials or disable Secrets Management in repository properties.');
```
horsmand:
```suggestion
      throw new Error('Admin credentials for Deadline Secrets Management cannot be generated when using an imported database. For setting up your own credentials, please refer to https://github.com/aws/aws-rfdk/tree/mainline/packages/aws-rfdk/lib/deadline#configuring-deadline-secrets-management-on-the-repository.');
```

packages/aws-rfdk/lib/deadline/lib/repository.ts
horsmand: I think putting it in the Stack's scope is probably the right place for it.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -1317,3 +1317,33 @@ test('credentials are undefined when secrets management is disabled', () => {
...
+    new Repository(stack, 'Repository', {
+      vpc,
+      version,
+      database: databaseConnection,
+    });
+  }
+
+  // THEN
+  expect(when).toThrow('Admin credentials for Deadline Secrets Management cannot be generated when using imported database. Please provide secret with admin credentials or disable Secrets Management in repository properties.');
+});
```
horsmand: Missing blank line

horsmand (changes requested):
This looks pretty good overall. The DocDB code seems good but I'm a bit concerned we haven't tested the MongoDB code to be sure that the same issue happens as with the DocDB database when we roll the repository back, and we're tying the admin credentials to the right construct

packages/aws-rfdk/lib/deadline/test/database-connection.test.ts
```diff
@@ -390,10 +390,11 @@ describe('DocumentDB Version Checks', () => {
     });
 
     // WHEN
-    DatabaseConnection.forDocDB({database, login: secret});
+    const databaseConnection = DatabaseConnection.forDocDB({database, login: secret});
 
     // THEN
     expect(database.node.metadata.length).toBe(0);
+    expect(databaseConnection.databaseConstruct).toBeUndefined();
```
jusiskin, line 406: Should we add a test for the opposite case to make sure `databaseConnection.databaseConstruct` is defined and points to the correct construct?

packages/aws-rfdk/lib/deadline/lib/repository.ts
```diff
@@ -581,23 +582,26 @@ export class Repository extends Construct implements IRepository {
...
           generateSecretString: {
             excludeCharacters: '\"$&\'()/<>[\\]\`{|}',
             includeSpace: false,
             passwordLength: 24,
             requireEachIncludedType: true,
-
             generateStringKey: 'password',
             secretStringTemplate: JSON.stringify({ username: Repository.DEFAULT_SECRETS_MANAGEMENT_USERNAME }),
           },
+          removalPolicy: RemovalPolicy.RETAIN,
```
jusiskin: I agree with having a default removal policy of retain, but I'm thinking it would be nice to expose this with a new `Repository` prop. Granted that users can pass in their own secrets with a destroy removal policy, this would be a nice-to-have for anyone developing/iterating on their CDK applications so they:

1.  don't have to concern themselves with the secret JSON format and password requirements
2.  aren't left with secret artifacts during development iterations

Another such example is the RFDK CI/CD automation which would currently leave secrets around in the accounts. I'm thinking the new prop could be something like:

```ts
new Repository(..., {
  secretsManagementSettings: {
    // "credentials" becomes mutually exclusive with removalPolicy below
    credentials: undefined,
    credentialsRemovalPolicy: RemovalPolicy.DESTROY,
  }
});
```

what do you think?

packages/aws-rfdk/lib/deadline/lib/repository.ts
jusiskin: > I think putting it in the Stack's scope is probably the right place for it.

I agree, although I think there is a logic error in my original suggestion. I think it should be:

```suggestion
        ((props.secretsManagementSettings?.enabled ?? true) ? new Secret(props.database?.databaseConstruct ? Stack.of(props.database.databaseConstruct) : this, 'SMAdminUser', {
```

packages/aws-rfdk/lib/deadline/lib/repository.ts
jusiskin: Looks like this is now added and the contents are being reviewed in another discussion. Resolving here.

jusiskin (changes requested):
Just a few suggestions here beyond what @horsmand already mentioned about missing tests.

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
jusiskin: Looks like we've pivoted to this, so I'll resolve this one.

packages/aws-rfdk/lib/deadline/lib/repository.ts
jusiskin: Looks good - thanks!

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
horsmand: This wasn't updated

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -109,10 +110,15 @@ export abstract class DatabaseConnection {
   }
 
   /**
-   * Returns the environment variables for configuring containers to connect to the database
+   * Holds the CDK Construct that contains the database, or undefined if the database was imported into this application.
```
horsmand: Looks like the wrong one was updated here.

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
horsmand: Can you comment here on the additional testing done to be sure we're tying the lifecycle of the admin credentials to the lifecycle of the mongo database?

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
kozlove-aws: Just put it in wrong line. Fixed.

packages/aws-rfdk/lib/deadline/lib/repository.ts
jusiskin: looks good!

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
jusiskin: According to our docs, [`MongoDbInstance.mongoDataVolume`](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.MongoDbInstance.html#mongodatavolume) is:

>   The EBS Volume on which we are storing the MongoDB database data.

We want to tie the admin secret life-cycle to the data, so this appears to be the correct construct.

jusiskin (approved):

horsmand (approved):

PR #593, opened 2021-10-04
chore(deadline): added tests for configure_identity_registration_settings.py

horsmand:
1. Wrote a bash script to execute the python unit tests and added a script entry in `package.json` called `testpy` to run the bash script. The tests are really lightweight, so I expanded the `test` target to run them along with the other CDK unit tests.
2. Wrote unit tests for `configure_identity_registration_settings.py`. I think coverage is pretty good but likely not 100%

These tests will run after the CDK unit tests are complete. The current output from their run looks like this:

```
aws-rfdk: Tests successful. Total time (4m12.2s) | /__w/aws-rfdk/aws-rfdk/node_modules/jest/bin/jest.js (4m12.1s)
aws-rfdk: $ ./test-python.sh
aws-rfdk: Running python tests
aws-rfdk: ............
aws-rfdk: ----------------------------------------------------------------------
aws-rfdk: Ran 12 tests in 0.009s
aws-rfdk: OK
```

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
+++ b/packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
@@ -1,0 +1,495 @@
+#!/usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Tests for configure_identity_registration_settings.py
+"""
+
+import sys
+import unittest
+from unittest.mock import MagicMock, Mock
+
+sys.modules['boto3'] = Mock()
+
+import configure_identity_registration_settings as subject
+
+RFDK_REG_SETTING_JSON_1 = {
+    "ConnectionIpFilterType":"IPv4Match",
+    "ConnectionIpFilterValue":"111.111.*.*", # 111.111.0.0/16
+    "SourceIpFilterType":"IPv4Match",
+    "SourceIpFilterValue":"123.123.123.*", # 123.123.123.0/24
+    "SettingsId":"settings1",
+    "SettingsName":"RfdkSubnet|subnet-aaaaaaa|subnet-1111111",
+    "IsEnabled":True,
+    "DefaultStatus":"Registered",
+    "DefaultRole":"Client"
+}
+RFDK_REG_SETTING_1 = subject.LoadBalancerIdentityRegistrationSetting(
+    connection_ip_filter_type=RFDK_REG_SETTING_JSON_1["ConnectionIpFilterType"],
+    connection_ip_filter_value=RFDK_REG_SETTING_JSON_1["ConnectionIpFilterValue"],
+    source_ip_filter_type=RFDK_REG_SETTING_JSON_1["SourceIpFilterType"],
+    source_ip_filter_value=RFDK_REG_SETTING_JSON_1["SourceIpFilterValue"],
+    settings_id=RFDK_REG_SETTING_JSON_1["SettingsId"],
+    settings_name=RFDK_REG_SETTING_JSON_1["SettingsName"],
+    is_enabled=RFDK_REG_SETTING_JSON_1["IsEnabled"],
+    default_status=RFDK_REG_SETTING_JSON_1["DefaultStatus"],
+    default_role=RFDK_REG_SETTING_JSON_1["DefaultRole"],
+)
+
+RFDK_REG_SETTING_JSON_2 = {
+    "ConnectionIpFilterType":"IPv4Match",
+    "ConnectionIpFilterValue":"255.*.*.*", # 255.0.0.0/8
+    "SourceIpFilterType":"IPv4Match",
+    "SourceIpFilterValue":"3.3.0-63.*", # 3.3.0.0/18
+    "SettingsId":"settings3",
+    "SettingsName":"RfdkSubnet|subnet-bbbbbbb|subnet-2222222",
+    "IsEnabled":True,
+    "DefaultStatus":"Registered",
+    "DefaultRole":"Client"
+}
+RFDK_REG_SETTING_2 = subject.LoadBalancerIdentityRegistrationSetting(
+    connection_ip_filter_type=RFDK_REG_SETTING_JSON_2["ConnectionIpFilterType"],
+    connection_ip_filter_value=RFDK_REG_SETTING_JSON_2["ConnectionIpFilterValue"],
+    source_ip_filter_type=RFDK_REG_SETTING_JSON_2["SourceIpFilterType"],
+    source_ip_filter_value=RFDK_REG_SETTING_JSON_2["SourceIpFilterValue"],
+    settings_id=RFDK_REG_SETTING_JSON_2["SettingsId"],
+    settings_name=RFDK_REG_SETTING_JSON_2["SettingsName"],
+    is_enabled=RFDK_REG_SETTING_JSON_2["IsEnabled"],
+    default_status=RFDK_REG_SETTING_JSON_2["DefaultStatus"],
+    default_role=RFDK_REG_SETTING_JSON_2["DefaultRole"],
+)
+
+NON_RFDK_REG_SETTING_JSON = {
+    "ConnectionIpFilterType":"IPv4Match",
+    "ConnectionIpFilterValue":"2.2.2.2",
+    "SourceIpFilterType":"IPv4Match",
+    "SourceIpFilterValue":"1.1.1.1",
+    "SettingsId":"settings2",
+    "SettingsName":"LBSetting2",
+    "IsEnabled":False,
+    "DefaultStatus":"Registered",
+    "DefaultRole":"Server"
+}
+
+
+class TestConfigureIdentityRegistrationSettingsConfigAndHelpers(unittest.TestCase):
+    def test_camel_to_snake_case(self):
+        test_cases = {
+            'CamelCase': 'camel_case',
+            'camelCamelCase': 'camel_camel_case',
+            'Camel2Camel2Case': 'camel_2_camel_2_case',
+            'GetHTTPResponseCode': 'get_http_response_code',
+            'get200HTTPResponseCode': 'get_200_http_response_code',
+            'getHTTP200ResponseCode': 'get_http_200_response_code',
+            'HTTPResponseCode': 'http_response_code',
+            'ResponseHTTP': 'response_http',
+            'ResponseHTTP2': 'response_http_2',
+            '10CoolDudes': '10_cool_dudes',
+            'aB': 'a_b'
+        }
+        for camel_case, snake_case in test_cases.items():
+            self.assertEqual(subject._camel_to_snake_case(camel_case), snake_case)
+
+
+    def test_parsing_basic_config(self):
+        config = subject.parse_args([
+            "--credentials", "arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c",
+            "--region", "test-region",
+            "--connection-subnet", "subnet-7e572",
+            "--source-subnet", "subnet-50c34,Client,Registered"
+        ])
+
+        self.assertEqual(config.credentials, subject.AwsSecret(arn="arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c", region="us-west-2"))
+        self.assertEqual(config.region, "test-region")
+        self.assertEqual(config.connection_subnet, ["subnet-7e572"])
+        self.assertListEqual( config.source_subnet, [ subject.SourceSubnet(subnet_id="subnet-50c34", role="Client", registration_status="Registered") ] )
+
+
+    def test_parsing_complex_config(self):
+        config = subject.parse_args([
+            "--credentials", "arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c",
+            "--region", "test-region",
+            "--connection-subnet", "subnet-7e573",
+            "--connection-subnet", "subnet-7e574",
+            "--source-subnet", "subnet-50c31,Client,Registered",
+            "--source-subnet", "subnet-50c32,Server,Pending"
+        ])
+
+        self.assertEqual(config.credentials, subject.AwsSecret(arn="arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c", region="us-west-2"))
+        self.assertEqual(config.region, "test-region")
+        self.assertListEqual(config.connection_subnet, [ "subnet-7e573", "subnet-7e574" ])
+        self.assertListEqual(
+            config.source_subnet,
+            [
+                subject.SourceSubnet(subnet_id="subnet-50c31", role="Client", registration_status="Registered"),
+                subject.SourceSubnet(subnet_id="subnet-50c32", role="Server", registration_status="Pending")
+            ]
+        )
+
+
+    def test_validate_config(self):
+        config = subject.parse_args([
+            "--credentials", "arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c",
+            "--region", "test-region",
+            "--connection-subnet", "subnet-7e573",
+            "--connection-subnet", "subnet-7e574",
+            "--source-subnet", "subnet-50c31,Client,Registered",
+            "--source-subnet", "subnet-50c32,Server,Pending"
+        ])
+
+        # Shouldn't throw
+        subject.validate_config(config)
+
+
+    def test_validate_config_no_connection(self):
+        config = subject.parse_args([
+            "--credentials", "arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c",
+            "--region", "test-region",
+            "--source-subnet", "subnet-50c31,Client,Registered"
+        ])
+
+        self.assertRaisesRegex(ValueError, "no --connection-subnet specified", lambda: subject.validate_config(config))
+
+
+    def test_validate_config_duplicate_subnet(self):
+        subnet_id = "subnet-8008135"
+        config = subject.parse_args([
+            "--credentials", "arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c",
+            "--region", "test-region",
+            "--connection-subnet", "subnet-7e573",
+            "--connection-subnet", "subnet-7e574",
+            "--source-subnet", f"{subnet_id},Client,Registered",
+            "--source-subnet", f"{subnet_id},Server,Pending"
+        ])
+
+        self.assertRaisesRegex(ValueError, f"Subnet \"{subnet_id}\" is not unique", lambda: subject.validate_config(config))
+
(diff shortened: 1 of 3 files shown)


packages/aws-rfdk/lib/deadline/scripts/python/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Created when running python tests
+__pycache__/
```
jusiskin: I moved this to `packages/aws-rfdk/.gitignore` in #576, so we can remove this file here.

packages/aws-rfdk/test-python.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -eou pipefail
```
jusiskin: nit: I had thought this needed to be:

```suggestion
set -euo pipefail
```

since the `pipefail` is an arg to the `-o` command-line option of `set`.

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,317 @@
+#!/usr/bin/env python3
```
jusiskin: nit: since we're introducing the test path convention here, what do you think about moving this file into a `test` sub-directory to keep them separate?

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,317 @@
...
+            subject.SourceSubnet(subnet_id="subnet-1111111", role="Client", registration_status="Registered"), # Used by REG_SETTING_RFDK_1
+            subject.SourceSubnet(subnet_id="subnet-1234567", role="Server", registration_status="Pending")
+        ]
+
+        subject.delete_removed_settings(self.dl_secrets, prior_rfdk_settings, connection_subnet_ids, source_subnets)
+
+        self.dl_secrets.run_sh.assert_called_once_with([
+            'DeleteLoadBalancerIdentityRegistrationSetting',
+            REG_SETTING_RFDK_2["SettingsId"]
+        ])
```
jusiskin: So we expect `delete_removed_settings` to remove `REG_SETTING_RFDK_2` here which uses both `subnet-bbbbbbb` (connection subnet) and `subnet-2222222` (source subnet).

`subject.delete_removed_settings` will remove this setting because `subnet-2222222` is not included in `source_subnets`. The test code/comments do not mention this subnet, so it is not quite obvious why the behavior is expected.

Perhaps we need a comment above the assertion that explains this, or is there a way we can make this test easier to follow?

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,317 @@
...
+    "ConnectionIpFilterType":0,
+    "ConnectionIpFilterValue":"111.111.*.*", # 111.111.0.0/16
+    "SourceIpFilterType":0,
+    "SourceIpFilterValue":"123.123.123.*", # 123.123.123.0/24
+    "SettingsId":"settings1",
+    "SettingsName":"RfdkSubnet|subnet-aaaaaaa|subnet-1111111",
+    "IsEnabled":True,
+    "DefaultStatus":2,
+    "DefaultRole":"Client"
+}
```
jusiskin: nit: could these be made into their named tuple literal equiavlents?

```suggestion
REG_SETTING_RFDK_1 = subject.LoadBalancerIdentityRegistrationSetting(
    connection_ip_filter_type=0,
    connection_ip_filter_value="111.111.*.*", # 111.111.0.0/16
    source_ip_filter_type=0,
    source_ip_filter_value="123.123.123.*", # 123.123.123.0/24
    settings_id="settings1",
    settings_name="RfdkSubnet|subnet-aaaaaaa|subnet-1111111",
    is_dnabled=True,
    default_status=2,
    default_role="Client"
)
```

this de-couples the tests from the `LoadBalancerIdentityRegistrationSetting.from_json(...)`. Same feedback applies to other places in this file

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,317 @@
...
+            "ConnectionIpFilterType":"IPv4Match",
+            "ConnectionIpFilterValue":"111.111.*.*",
+            "SourceIpFilterType":"IPv4Match",
+            "SourceIpFilterValue":"123.123.123.*",
+            "SettingsId":"settings1",
+            "SettingsName":"RfdkSubnet|subnet-aaaaaaa|subnet-1111111",
+            "IsEnabled":True,
+            "DefaultStatus":"Pending",
+            "DefaultRole":"Server"
+        })
```
jusiskin: What do you think about sourcing the values for some of these expected fields from literals to using attributes of `REG_SETTINGS_RFDK_1` where applicable? just to help depict what fields we expect to be updated vs what we expect to remain unchanged?

jusiskin (changes requested):
This looks great overall. I just have a few minor suggestions - mostly around improving readability of the tests.

kozlove-aws (dismissed):
LGTM! Good job.

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
horsmand: Updated

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
horsmand: Updated

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
horsmand: Updated to explain better.

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
horsmand: Moved.

packages/aws-rfdk/test-python.sh
horsmand: Oh yeah, my mistake.

packages/aws-rfdk/lib/deadline/scripts/python/.gitignore
horsmand: Removed.

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,375 @@
...
+from unittest.mock import MagicMock, Mock
+
+sys.modules['boto3'] = Mock()
+
+# Python 3's modules don't make it easy to import from a directory above. This solves that.
+import pathlib
+_parentdir = pathlib.Path(__file__).parent.parent.resolve()
+sys.path.insert(0, str(_parentdir))
+import configure_identity_registration_settings as subject
+sys.path.remove(str(_parentdir))
```
jusiskin: This can be simplified to:

```suggestion
import configure_identity_registration_settings as subject
```

but you need to make a few changes:

1.  Modify the `-s` to be up a directory in `packages/aws-rfdk/test-python.sh`:

    ```sh
    python3 -m unittest discover -s "$PWD/lib/deadline/scripts/python" -b
    ```
2.  Add an empty `packages/aws-rfdk/lib/deadline/scripts/python/test/__init__.py` file to mark the `test` directory as a python package

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,375 @@
...
+        self.dl_secrets.run_json = MagicMock(return_value=configured_reg_settings)
+
+        # WHEN
+        rfdk_managed_settings = subject.get_rfdk_registration_settings(self.dl_secrets)
+
+        # THEN
+        self.assertListEqual(
+            rfdk_managed_settings,
+            self.prior_rfdk_reg_settings
+        )
```
jusiskin: should we assert the arguments that `self.dl_secrets_run_json` was called with?

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def test_create(self):
```
jusiskin, line 233: while this covers one situation where we **want to** create a setting, should we have negative test for when we **should not** create a setting?

same applies for update and delete

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,375 @@
...
+            connection_ip_filter_type="IPv4Match", # This is equivalent to `0` that `RFDK_REG_SETTING_JSON_1` uses
+            connection_ip_filter_value=RFDK_REG_SETTING_JSON_1["ConnectionIpFilterValue"],
+            source_ip_filter_type="IPv4Match", # This is equivalent to `0` that `RFDK_REG_SETTING_JSON_1` uses
+            source_ip_filter_value=RFDK_REG_SETTING_JSON_1["SourceIpFilterValue"],
+            settings_id=RFDK_REG_SETTING_JSON_1["SettingsId"],
+            settings_name=RFDK_REG_SETTING_JSON_1["SettingsName"],
+            is_enabled=RFDK_REG_SETTING_JSON_1["IsEnabled"],
+            default_status="Pending",
+            default_role="Server"
+        )
```
jusiskin: I'm wondering if we can write this in a way that makes the intent a little more clear. While the above code is logically equivalent, how about this:

```suggestion
        expected = subject.LoadBalancerIdentityRegistrationSetting(
            # We expect RFDK_REG_SETTING_1 to be updated
            settings_id                 = RFDK_REG_SETTING_1.settings_id,
            # The values we expect to be applied
            connection_ip_filter_type   = "IPv4Match",
            connection_ip_filter_value  = subject.cidr_to_ipv4_match(subnet_to_cidr[connection_subnet_ids[0]]),
            source_ip_filter_type       = "IPv4Match",
            source_ip_filter_value      = subject.cidr_to_ipv4_match(subnet_to_cidr[source_subnets[0].subnet_id]),
            settings_name               = RFDK_REG_SETTING_1.settings_name,
            is_enabled                  = True,
            default_status              = source_subnets[0].registration_status,
            default_role                = source_subnets[0].role
        )
```

Do you think this makes it any more clear when reading the code that we expect the setting values to be updated using the values from `connection_subnet_ids`, `subnet_to_cidr`, and `source_subnets` rather than from the previous values?

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,375 @@
...
+            "subnet-4444444": "104.0.0.0/22", # 104.0.0-3.*
+        }
+
+        # WHEN
+        subject.create_and_update_settings(self.dl_secrets, [], connection_subnet_ids, source_subnets, subnet_to_cidr)
+
+        # THEN
+        # We can only assert on the values of the last call to a Mock, so instead we just make sure that the right number
+        # of identity registration settings are created
+        self.assertEqual(self.dl_secrets.run_str.call_count, 4)
```
jusiskin: We can use [`Mock.assert_any_call(...)`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_any_call) to search all calls to the mock.

```suggestion
        for connection_subnet_id in connection_subnet_ids:
            for source_subnet in source_subnets:
                expected = subject.LoadBalancerIdentityRegistrationSetting(
                    connection_ip_filter_type  = "IPv4Match",
                    connection_ip_filter_value = subject.cidr_to_ipv4_match(subnet_to_cidr[connection_subnet_id]),
                    source_ip_filter_type      = "IPv4Match",
                    source_ip_filter_value     = subject.cidr_to_ipv4_match(subnet_to_cidr[source_subnet.subnet_id]),
                    settings_id                = '', # Unused
                    settings_name              = f"RfdkSubnet|{connection_subnet_id}|{source_subnet.subnet_id}",
                    is_enabled                 = True,
                    default_status             = source_subnet.registration_status,
                    default_role               = source_subnet.role
                )
                self.dl_secrets.run_str.assert_any_call(
                    "CreateLoadBalancerIdentityRegistrationSetting",
                    expected.settings_name,
                    expected.connection_ip_filter_type,
                    expected.connection_ip_filter_value,
                    expected.source_ip_filter_type,
                    expected.source_ip_filter_value,
                    expected.default_role,
                    expected.default_status,
                    str(expected.is_enabled)
                )
```

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,375 @@
...
+            "ConnectionIpFilterType":"IPv4Match",
+            "ConnectionIpFilterValue":"111.111.*.*", # Similar to REG_SETTING_1
+            "SourceIpFilterType":"IPv4Match",
+            "SourceIpFilterValue":"3.3.0-63.*", # Similar to REG_SETTING_2
+            "SettingsId":"",
+            "SettingsName":"RfdkSubnet|subnet-aaaaaaa|subnet-2222222",
+            "IsEnabled":True,
+            "DefaultStatus":"Registered",
+            "DefaultRole":"Client"
+        })
```
jusiskin: How about:

```suggestion
        expected = subject.LoadBalancerIdentityRegistrationSetting(
            connection_ip_filter_type   = "IPv4Match",
            connection_ip_filter_value  = subject.cidr_to_ipv4_match(subnet_to_cidr[connection_subnet_ids[0]]),
            source_ip_filter_type       = "IPv4Match",
            source_ip_filter_value      = subject.cidr_to_ipv4_match(subnet_to_cidr[source_subnets[0].subnet_id]),
            settings_id                 = "", # Ignored
            settings_name               = f"RfdkSubnet|{connection_subnet_ids[0]}|{source_subnets[0].subnet_id}",
            is_enabled                  = True,
            default_status              = source_subnets[0].registration_status,
            default_role                = source_subnets[0].role
        )
```

we already have separate testing for cidr_to_ipv4_match so this should be okay and helps illustrate the expected behavior.

What do you think?

kozlove-aws (dismissed):

jusiskin (changes requested):
Looking good. I have a few readability suggestions, which I've provided code for, but there's still some missing coverage around negative tests that we may want to add.

packages/aws-rfdk/lib/deadline/scripts/python/test_configure_identity_registration_settings.py
jusiskin: I'd overlooked the fact that `get_rfdk_registration_settings` gets JSON-encoded settings from `deadlinecommand` and passes that to `LoadBalancerIdentityRegistrationSetting.from_json`. This makes testing tricky since we either need to mock `deadlinecommand` to return valid JSON, or mock `LoadBalancerIdentityRegistrationSetting.from_json` to have stateful returns.

Either is not ideal, but I think it suffices for now and if we ever need to, we can refactor that later. Resolving.

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
horsmand: Well that was some python wizardry I didn't know about. Thanks!

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
horsmand: Sure, updated

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
horsmand: I've added some tests for these cases.

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
horsmand: Yeah this makes sense to me. I was a bit wary about using the `cidr_to_ipv4_match` to do the conversion, but that method is tested elsewhere so it should be fine. Do you see any reason not to take the filter types from `RFDK_REG_SETTING_1` as well?

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
horsmand: Ah thanks, not sure how I missed that method.

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
horsmand: Updated

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
```diff
@@ -0,0 +1,495 @@
...
+        }
+
+        # WHEN
+        subject.create_and_update_settings(self.dl_secrets, self.prior_rfdk_reg_settings, connection_subnet_ids, source_subnets, subnet_to_cidr)
+
+        # THEN
+        self.dl_secrets.run_str.assert_not_called()
+
+
+    def test_create_and_update_no_connection_subnets(self):
```
jusiskin:
```suggestion
    def test_create_and_update_no_source_subnets(self):
```

jusiskin (changes requested):
Just one minor method name conflict. But otherwise this is great

packages/aws-rfdk/lib/deadline/scripts/python/test/test_configure_identity_registration_settings.py
horsmand: Whoops, surprised it let me create methods with the same name. Fixed.

jusiskin: Technically it's valid python, but it could be caught with a linter. That's out of scope for this work, so we can add it to the backlog

jusiskin (approved):
Great work here! This looks good to me

kozlove-aws (approved):

PR #596, opened 2021-10-07
feat(integ): add secrets management identity registration settings tests

jericht:
### Notes
- Adds integration tests for the Secrets Management identity registration settings feature implemented in https://github.com/aws/aws-rfdk/pull/576
    - The `deadline_05_secretsManagement` test suite now uses a default (dummy) UBL certificate secret (create if not found). This secret is meant to be created once per account + region these integ tests are run in and referenced in all future integ tests, since its content is static (basically empty).
- Adds a new developer mode to the integration test script so that the CF stacks will not be torn down, and making the bastion host accessible via SSM, making it easier to iterate on integ test development
- Reorganizes the `NetworkTier` to have dedicated subnet configurations for each Deadline component (RQ ALB, workers, license forwarder, etc.)
- Adds options to the `RenderStruct` construct to setup a `UsageBasedLicensing` construct. This is currently only used in the `deadline_05_secretsManagement` test suite.
- Adds a new `SepWorkerStruct` that configures the Spot Event Plugin via `ConfigureSpotEventPlugin` construct

~~**Note for reviewers:** This is based off of the changes in https://github.com/aws/aws-rfdk/pull/576. It will be easier to review by looking at the commits authored by me so only my changes are shown in the file diffs.~~

### Testing
Took the following steps to test:
- Staged Deadline 10.1.18 Docker recipes & updated them with the changes for the Secrets Management feature
- Changed the `SecretsManagementIdentityRegistration` construct to use a version of Deadline Client with the Secrets Management changes
- Let the integ test create a dummy secret for the UBL certificates, then ran the integration tests and verified they passed
- Ran all the integ test suites and verified they passed

----


integ/components/_infrastructure/lib/network-tier.ts
```diff
@@ -3,10 +3,55 @@
...
+  SubnetType,
+  Vpc,
+} from '@aws-cdk/aws-ec2';
+import {
+  Construct,
+  Stack,
+  StackProps,
+} from '@aws-cdk/core';
+
+export enum NetworkTierSubnets {
```
horsmand: Do we need this enum? I don't see it referenced anywhere other than in the type right below it and we could just list the subnets there.

integ/components/_infrastructure/lib/network-tier.ts
```diff
@@ -3,10 +3,55 @@
...
+  testRunner = 'testRunner',
+  renderQueueAlb = 'renderQueueAlb',
+  sepFleet = 'sepFleet',
+  ubl = 'ubl',
+  workerInstanceFleet = 'workerInstanceFleet',
+}
+
+export type NetworkTierSubnetConfiguration = {
```
horsmand: What was the reasoning to use a `type` over an `interface` here? I'm not sure how this plays with JSII. That probably doesn't matter since I'm quite sure we never transpile this integ package, but it's probably best if we follow the typescript restrictions everywhere in our repository for consistency. 

From https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/#interfaces:
> The jsii type model distinguishes two kinds of interfaces:
>
> * Behavioral interfaces, which can declare methods and properties
> * Structs, which are immutable pure data entities, and can consequently only declare readonly properties
>
> A name convention is used to distinguish between these two: behavioral interfaces must have a name that starts with a I prefix, while structs must not have such a prefix.

I don't see why we can't use the struct interface here.

integ/scripts/bash/rfdk-integ-e2e.sh
```diff
@@ -71,8 +99,12 @@ export PRETEST_FINISH_TIME=$SECONDS
 
 # Define cleanup function for deployment failure
 cleanup_on_failure () {
+  if [[ "$DEV_MODE" == "true" ]]; then
```
horsmand, line 100: I can't seem to comment on lines that aren't modified in this commit, but above this on line 88 there's a check that is broken. It currently has:
```
# If executing worker fleet tests, find Deadline AMIs based on supplied version
if [ ! "${SKIP_deadline_03_repository_TEST-}" = true ]; then
    source $BASH_SCRIPTS/fetch-worker-amis.sh
fi
```
I think we want to make sure we're not skipping `deadline_03_workerFleetHttp`, `deadline_03_workerFleetHttps`, and `deadline_05_secretsManagement`. I almost wonder if the check is worth doing since it's easy to get missed when we make modifications and fetching AMIs likely isn't too expensive.

horsmand:
This commit just got merged to the feature branch so we'll need to rebase on it and set the credentials removal policy to `DESTROY` https://github.com/aws/aws-rfdk/commit/d0241d6989b1af1fbced953ed91444946b2655f8#diff-299db76f4b1d275d1140f5b68b75b5954a2c54ef4f6770dfc410bb21668aea8eR318

integ/components/deadline/deadline_05_secretsManagement/bin/deadline_05_secretsManagement.ts
```diff
@@ -34,21 +43,77 @@ const storageStruct = new StorageStruct(componentTier, 'StorageStruct', {
...
+/**
+ * If the UBL_CERTIFICATE_BUNDLE_SECRET_ARN env var is specified, this function parses the UBL_LICENSE_MAP environment variable
+ * and returns an object containing the ARN of the UBL certificate secret and the UsageBasedLicenses that will be used.
+ * Otherwise, it returns undefined.
+ */
+function getUsageBasedLicensingProperties(): RenderStructUsageBasedLicensingProps | undefined {
+  const ublCertificateBundleSecretArn = process.env.UBL_CERTIFICATE_BUNDLE_SECRET_ARN;
+  if (!ublCertificateBundleSecretArn) {
+    throw new Error('UBL_CERTIFICATE_BUNDLE_SECRET_ARN is required but was not specified.');
+  }
```
horsmand: Should this return undefined rather than throw an error? I don't see how these tests work without having the UBL env vars defined, but the documentation says it should be possible.

integ/components/deadline/deadline_05_secretsManagement/lib/secretsManagement-testing-tier.ts
```diff
@@ -4,18 +4,30 @@
...
+import { NetworkTier } from '../../../_infrastructure/lib/network-tier';
 
 /**
  * Properties for SecretsManagementTestingTier
  */
 export interface SecretsManagementTestingTierProps extends TestingTierProps {
   readonly renderStruct: RenderStruct;
   readonly storageStruct: StorageStruct;
+  readonly workerStruct?: WorkerStruct;
+  readonly sepWorkerStruct?: SepWorkerStruct;
```
horsmand: These don't seem optional

integ/test-config.sh
```diff
@@ -27,6 +27,23 @@ export DEADLINE_STAGING_PATH
 #     }
 export RFDK_DOCKER_IMAGE_OVERRIDES
 
+export UBL_CERTIFICATE_BUNDLE_SECRET_ARN
```
horsmand, line 34: How is this intended to be used? Will UBL be used in our canaries or pipelines?

integ/components/deadline/deadline_05_secretsManagement/lib/secretsManagement-testing-tier.ts
```diff
@@ -48,6 +60,12 @@ export class SecretsManagementTestingTier extends TestingTier {
     this.configureBastionUserData({
       testingScriptPath: path.join(__dirname, '../scripts/bastion/testing'),
     });
+
+    props.workerStruct?.workerFleet.forEach(wf => this.testInstance.connections.allowTo(wf, Port.tcp(22)));
+    props.sepWorkerStruct?.fleets.forEach(f => this.testInstance.connections.allowTo(f, Port.tcp(22)));
```
horsmand: What is the test instance connecting to these fleets for?

integ/components/deadline/deadline_05_secretsManagement/lib/secretsManagement-testing-tier.ts
```diff
@@ -71,4 +89,34 @@ export class SecretsManagementTestingTier extends TestingTier {
       });
     }
   }
+
+  /**
+   * Generates CfnOutputs for the Render Queue ALB subnets and Deadline client components.
```
horsmand:
```suggestion
   * Generates CfnOutputs for the Render Queue ALB subnets and Deadline client subnets.
```

integ/components/deadline/deadline_05_secretsManagement/test/deadline_05_secretsManagement.test.ts
```diff
@@ -46,6 +100,30 @@ beforeAll(async () => {
   if (!smSecretArns[testSuiteId]) {
     errors.push(`A stack output for deadlineSecretsManagementCredentialsSM${testSuiteId} is required but was not found`);
   }
+  if (!renderQueueAlbSubnetIds[testSuiteId]) {
+    errors.push(`A stack output for renderQueueAlbSubnetIdsSM${testSuiteId} is required but was not found`);
+  }
+  if (!ublSubnetIds[testSuiteId]) {
```
horsmand: Isn't UBL optional?

horsmand (changes requested):
Overall this looks good, I really like that you fixed up the `--deploy-and-test-only` flag by adding the `DEV_MODE` env var. The only part I'm unsure about is the intention of adding UBL and whether it's actually optional given how it's currently written.

integ/components/_infrastructure/lib/network-tier.ts
jericht: We don't. I originally thought it would be needed but turned out it wasn't used, so I've removed it

integ/components/_infrastructure/lib/network-tier.ts
jericht: Changed to a struct interface

integ/test-config.sh
jericht: Sorry, forgot to put a comment here. Originally, I intended for this to be required when running the Secrets Management tests, but it makes more sense to make them optional overrides and have default behavior (e.g. upload dummy secret and use dummy license limits). I've added a comment here as well as implemented the default behavior.

integ/components/deadline/deadline_05_secretsManagement/bin/deadline_05_secretsManagement.ts
jericht: I changed this to create a default dummy secret for UBL when this in not defined

integ/components/deadline/deadline_05_secretsManagement/lib/secretsManagement-testing-tier.ts
jericht: These are actually not needed, removed

integ/components/deadline/deadline_05_secretsManagement/lib/secretsManagement-testing-tier.ts
jericht: It's not, I forgot to remove this. Removed

integ/components/deadline/deadline_05_secretsManagement/lib/secretsManagement-testing-tier.ts
jericht: Updated

integ/components/deadline/deadline_05_secretsManagement/test/deadline_05_secretsManagement.test.ts
jericht: I've changed the UBL stuff to always get setup. This way our SM integ tests always have the same coverage, regardless of whether UBL certs were provided or not.

integ/scripts/bash/rfdk-integ-e2e.sh
jericht: Yeah I agree, I've removed that if condition.

integ/components/deadline/deadline_05_secretsManagement/test/deadline_05_secretsManagement.test.ts
```diff
@@ -108,4 +186,146 @@ describe('Deadline Secrets Management tests', () => {
...
+      DocumentName: 'AWS-RunShellScript',
+      Comment: 'Execute GetLoadBalancerIdentityRegistrationSettings via test script SM-run-secrets-command.sh',
+      InstanceIds: [bastionId],
+      Parameters: {
+        commands: [
+          `sudo -u ec2-user ~ec2-user/testScripts/SM-run-secrets-command.sh '${AWS.config.region}' '${smSecretArns[testSuiteId]}' GetLoadBalancerIdentityRegistrationSettings`,
+        ],
+      },
+    };
+    const rqAlbSubnetIds = renderQueueAlbSubnetIds[testSuiteId];
```
horsmand: I'm a little confused by the use of arrays and this `testSuiteId` in this file. I don't see the value of `testSuiteId` changing anywhere.

integ/components/deadline/deadline_05_secretsManagement/bin/deadline_05_secretsManagement.ts
```diff
@@ -3,52 +3,197 @@
...
+ * @param dirPath The path to create the certificate bundle in.
+ * @returns The path to the certificate bundle file.
+ */
+function createDummyUblCertificates(dirPath: string): Promise<string> {
+  return new Promise(async (res, rej) => {
+    const outputPath = path.join(dirPath, 'certificates.zip');
+    const output = fs.createWriteStream(outputPath);
+    output.once('close', () => res(outputPath));
+
+    const archive = archiver('zip');
```
horsmand: It looks like we've added a new dependency just to create this empty zip file. Why not just keep an empty zip file in our repo that we can use? Binary files in a git repo are generally bad, but I feel like this might warrant an exception since the file size is negligible. What do you think?

horsmand:
I forgot to mention that my last review was based on this diff, which covers the 2 latest commits: https://github.com/aws/aws-rfdk/pull/596/files/4bb8df5b36d32f0c435466b64cf21231950086d9..0a58b66e61cdca280e35e2d907b099ce7f485956

integ/components/deadline/deadline_05_secretsManagement/test/deadline_05_secretsManagement.test.ts
jericht: Good point. Unlike the other component tests, this test only has one test suite, I've removed this.

integ/components/deadline/deadline_05_secretsManagement/bin/deadline_05_secretsManagement.ts
jericht: Good point! I think having the static empty file is much cleaner, updated.

integ/components/deadline/deadline_05_secretsManagement/bin/deadline_05_secretsManagement.ts
```diff
@@ -3,52 +3,169 @@
...
+
+    let parsedUblLicenseMap: {[license: string]: number};
+    try {
+      parsedUblLicenseMap = JSON.parse(ublLicenseMap);
+    } catch (e) {
+      throw new Error(`Failed to parse UBL_LICENSE_MAP: ${e}`);
+    }
+
+    const licenses = [...Object.keys(parsedUblLicenseMap).map(l => {
+      const limit = parsedUblLicenseMap[l] > 0 ? parsedUblLicenseMap[l] : UsageBasedLicense.UNLIMITED;
```
kozlove-aws: if we expect here zero or negative values could be worth to validate if value more then `UsageBasedLicense.UNLIMITED`

integ/components/deadline/deadline_05_secretsManagement/bin/deadline_05_secretsManagement.ts
```diff
@@ -3,52 +3,169 @@
...
+        };
+      } else {
+        // Secret exists but is marked for deletion, so we need to restore the secret then update its value
+        putSecret = async data => {
+          await secrets.restoreSecret({ SecretId: secretName }).promise();
+          const updateSecretResponse = await secrets.updateSecret({ SecretId: secretName, SecretBinary: data }).promise();
+          return updateSecretResponse.ARN!;
+        };
+      }
+    } catch (e) {
```
kozlove-aws, line 142: How can we be sure that the only possible error is when the secret is not found? 
Could we check error type here?

kozlove-aws (commented):
Looks good, just a couple tiny things.

integ/lib/storage-struct.ts
```diff
@@ -79,12 +79,31 @@ export class StorageStruct extends Construct {
     // Get farm VPC from lookup
     const vpc = Vpc.fromLookup(this, 'Vpc', { tags: { StackName: infrastructureStackName }}) as Vpc;
 
+    // Create EFS filesystem
+    const deadlineEfs = new FileSystem(this, 'FileSystem', {
```
horsmand, line 83: Let's add a comment about why we need to create this all the time now but not the DocDB

horsmand (changes requested):
We're very close! Just one comment for clarification.

integ/components/deadline/deadline_05_secretsManagement/bin/deadline_05_secretsManagement.ts
horsmand: I know I saw it explained somewhere else in this PR, but maybe we can just add a comment here to say that anything that isn't a positive integer will be treated as unlimited.

jericht: Good call out! I think we can leave this as is though. Technically, there is no upper limit on the license count limit (see our [license-forwarder recipe documentation for UBL_LIMITS](https://gallery.ecr.aws/aws-thinkbox/deadline/license-forwarder)), and if there is a max value in Deadline, I think we can let Deadline enforce that. How does that sound?

integ/components/deadline/deadline_05_secretsManagement/bin/deadline_05_secretsManagement.ts
jericht: Good idea, added a check to make sure we get the `ResourceNotFoundException` error code

integ/lib/storage-struct.ts
jericht: Sure, added a little comment here as well as over the relevant `vpcSubnets` field

jericht: Also just to make sure we're on the same page here, we can't make the DocDB all the time since the `StorageStruct` switches the type of database it uses based on the props it's given (see the `if` statements below this code), but it does not switch the filesystem type, which is why I have refactored it out.

Previous behavior of filesystem creation w.r.t the `databaseType` prop was:
- `DatabaseType.DocDB` => Create EFS and pass it into the `Repository`
- `DatabaseType.MongoDB` => Let `Repository` create the EFS
- `undefined` => Also let `Repository` create the EFS

I found it strange that we only created the EFS in the DocDB case since those two things aren't really tied to each other for our integ tests. Since I needed something other than the default EFS created by our `Repository`, I decided to just refactor the EFS creation outside of the database creation branches.

jericht:
@kozlove-aws @horsmand This is ready for re-review. Please don't merge if both of you approve, I'm going to do a final run of the integ test and merge it myself once it passes.

horsmand (commented):
Ready to approve after we confirm the latest run of the tests pass.

kozlove-aws (approved):

jericht:
@horsmand The tests just passed, this is good to go

horsmand (approved):

PR #598, opened 2021-10-07
feat(examples): use dedicated subnets in All-In-AWS-Infrastructure-Basic example

jusiskin:
## Summary

In #576, RFDK was modified to automatically configure Deadline Secrets Management identity registration settings based on the subnets used by RFDK Deadline constructs. This sets up the rules with least-privilege such that the farm will always function, but in order to further scope-down the identity registration settings rules, RFDK users should deploy RFDK Deadline constructs into their own dedicated subnets.

This PR modifies the `All-In-AWS-Infrastructure-Basic` example CDK applications to use dedicated subnets as we now recommend in our documentation.

Updated the Deadline `README.md` to point to the example for a reference implementation.

## Testing

Built and deployed Python and TypeScript example apps and verified that:

*   Python
    *  All instances are deployed in the proper subnets
    *  The Render Queue ALB is deployed in its own dedicated subnets
    *  RFDK creates Deadline Secrets Management identity registration settings for only the subnets involved
    *  End-to-end UBL render using SM succeeds
*   TypeScript
    *  All instances are deployed in the proper subnets
    *  The Render Queue ALB is deployed in its own dedicated subnets
    *  RFDK creates Deadline Secrets Management identity registration settings for only the subnets involved
    *  End-to-end UBL render using SM succeeds

----

### DIFF
--- a/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/service-tier.ts
+++ b/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/service-tier.ts
@@ -7,7 +7,6 @@ import {
   BastionHostLinux,
   BlockDeviceVolume,
   IVpc,
-  SubnetType,
 } from '@aws-cdk/aws-ec2';
 import {
   ApplicationProtocol,
@@ -34,6 +33,8 @@ import {
   Secret,
 } from '@aws-cdk/aws-secretsmanager';
 import { SessionManagerHelper } from 'aws-rfdk/lib/core';
+
+import { Subnets } from './subnets';
 
 /**
  * Properties for {@link ServiceTier}.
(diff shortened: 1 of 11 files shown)


examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/service-tier.ts
```diff
@@ -191,6 +195,9 @@ export class ServiceTier extends cdk.Stack {
 
     this.renderQueue = new RenderQueue(this, 'RenderQueue', {
```
horsmand, line 196: This is missing setting the ALB subnets.

horsmand (changes requested):
Did a quick pass. Looks good!

examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/service-tier.ts
jusiskin: Good catch. Fixed this now

jericht (dismissed):
Looks good!

horsmand (approved):

jericht (approved):

PR #621, opened 2021-10-25
fix(integ): increase cidr range for render queue alb subnet

horsmand:
We were using a CIDR mask of 27 for the Render Queue ALB's subnet, which allowed for `2^(32-27)-2 = 30` IP addresses to be added to it. This was enough when we were running the tests serially, but when we run them in parallel we sometimes see errors like this:
```
CREATE_FAILED        | AWS::ElasticLoadBalancingV2::LoadBalancer   | RenderStructWF2/RenderQueue/LB (RenderStructWF2RenderQueueLBC2243A01) Not enough IP space available in subnet-08885c5805228af7a. ELB requires at least 8 free IP addresses in each subnet. (Service: AmazonElasticLoadBalancing; Status Code: 400; Error Code: InvalidSubnet; Request ID: 76f87391-94f2-4297-92d7-70811fe8f136; Proxy: null)
```
The [ALB documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#subnets-load-balancer) suggests to use at least a `/27` bitmask, and there should be at least 8 free IP addresses. Depending on the traffic profile, the ALB could use up to 100 addresses.

To fix this, we are increasing the size of the available address space to `2^(32-25)-2 = 126`, which is `126/7 = 18` IP addresses per render queue. This is lower than the recommendation of 30, but well above the minimum of 8.

## Testing 
I ran a parallel run of the integ tests in my own account. 
The repository tests failed with:
```
CREATE_FAILED        | Custom::LogRetention                        | AttachEniToInstance83a5dca5db544aa485d28d419cdf85ce/LogRetention (AttachEniToInstance83a5dca5db544aa485d28d419cdf85ceL
ogRetention77A16AA1) Received response status [FAILED] from custom resource. Message returned: A conflicting operation is currently in progress against this resource. Please try again. (RequestId: 39df591e-a993-4
a48-a4f8-5ed5a3395d7f)
```

The HTTP worker tests failed with:
```
CREATE_FAILED        | Custom::LogRetention                        | WorkerStructWF1/Worker1/Worker1/Worker1LogGroupWrapper (WorkerStructWF1Worker1Worker1LogGroupWrapperBC004C92) Received response status [FAILED] from custom resource. Message returned: A conflicting operation is currently in progress against this resource. Please try again. (RequestId: f573a7b6-52ed-479a-acec-0cc6af65a25e)
```

----

### DIFF
--- a/integ/components/_infrastructure/lib/network-tier.ts
+++ b/integ/components/_infrastructure/lib/network-tier.ts
@@ -32,7 +32,17 @@ export class NetworkTier extends Stack {
     renderQueueAlb: {
       name: 'RenderQueueAlbSubnets',
       subnetType: SubnetType.PRIVATE,
-      cidrMask: 27,
+      // Current RenderQueueStructs:
+      //   deadline_02_renderQueue: 2
+      //   deadline_03_workerFleetHttp: 2
+      //   deadline_04_workerFleetHttps: 2
+      //   deadline_05_secretsManagement: 1
+      // 7 total
+      // If we choose a CIDR mask of 25, we get 2^(32-25)-2 = 126 IP addresses
+      // 126/7 = 18 IP addresses per RenderQueue
+      // Recommended addresses is 30 per subnet with a minimum of 8, so /25 gives us a little room before we hit the minimum. Refer to:
+      // https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#subnets-load-balancer
+      cidrMask: 25, // 2^(32-25)-2 = 126 IP addresses
     },
     sepFleet: {
       name: 'SepFleetSubnets',


integ/components/_infrastructure/lib/network-tier.ts
```diff
@@ -32,7 +32,16 @@ export class NetworkTier extends Stack {
...
+      // Current RenderQueueStructs:
+      //   deadline_02_renderQueue: 2
+      //   deadline_03_workerFleetHttp: 2
+      //   deadline_04_workerFleetHttps: 2
+      //   deadline_05_secretsManagement: 1
+      // 7 total
+      // 126/7 = 18 IP addresses per RenderQueue
+      // Recommended addresses is 30 per subnet with a minimum of 8, refer to:
+      // https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#subnets-load-balancer
+      cidrMask: 25, // 2^(32-25)-2 = 126 IP addresses
```
kozlove-aws, line 45: From comment is not clear why we can`t use cidr mask 26 with 62 addresses.

horsmand: That makes `2^(32-26)-2 = 62` IP addresses and `62/7 = 8.86` which is close to the minimum. Adding one more render queue to the subnet would break it. I think using 25 leaves a little space if we update our tests in the future.

kozlove-aws: right, I just mean after all this calculations it is not clear why we choose 25.

horsmand: Updated to make it a little clearer, let me know what you think

kozlove-aws (approved):

jusiskin (approved):

PR #631, opened 2021-11-08
docs(deadline): fix VersionQuery docs

jusiskin:
## Problem

The docs in the Deadline module's `README.md` is stale and references some removed APIs (`VersionQuery.exact(...)` and `VersionQuery.exactString(...)`) that have since been removed (#175).

## Solution

Remove references to the old API and replace with documentation for the new equivalent usages.

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/README.md
+++ b/packages/aws-rfdk/lib/deadline/README.md
@@ -196,7 +196,9 @@ The `RenderQueue` construct sets up the RCS and configures it to communicate wit
 The following example outlines how to construct a `RenderQueue`:
 
 ```ts
-const version = VersionQuery.exactString(stack, 'Version', '1.2.3.4');
+const version = new VersionQuery(stack, 'Version', {
+  version: '1.2.3.4',
+});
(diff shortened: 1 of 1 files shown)


packages/aws-rfdk/lib/deadline/README.md
```diff
@@ -680,11 +684,8 @@ The `VersionQuery` construct encapsulates a version of Deadline and the location
 
 You can specify a Deadline version as follows:
 ```ts
-const version = VersionQuery.exact(stack, 'ExactVersion', {
-  majorVersion: '1',
-  minorVersion: '2',
-  releaseVersion: '3',
-  patchVersion: '4'
+const version = new VersionQuery(stack, 'ExactVersion', {
```
jericht:
```suggestion
const version = new VersionQuery(stack, 'Version', {
```

super minor, but I wonder if we should change the construct ID to just `Version` since we've removed the `exact` method. What do you think?

jericht (approved):

horsmand (approved):

PR #634, opened 2021-11-17
feat(deadline): add X509 certificate expiration alarms

kozlove-aws:
### Problem

Certificates created in RFDK has expiration date but does not have procedure for rotation.
By default certificates will expire after 3 years, but there is still possibility that it expire and we need to have mechanism to warn customer that certificate have to be rotated.

### Solution

Was added a Lambda function for each certificate that checks secret with certificate every day and fill Metric with data how many days left before certificate will expire.
Alarm will be raised when certificate has only 15 days left to be valid.
Customer can provide an email and message about this alarm will be send in this case.

### Testing

Deployed example and validated:
- Lambda is created for each certificate
- Lambda is running every day
- Metric for secrets are created and values are filled
- Alarm is raised when value in metric became less than 15 days
- SNS topic is created and email about alarm was sent.
 
----


packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +270,91 @@ abstract class X509CertificateBase extends Construct {
...
+      lambdaCode: props.lambdaCode,
+      tagCondition,
+    });
+    this.addExpDateValidation({
+      uniqueValue,
+      alarmEmailAddress: props.alarmEmailAddress,
+    });
+  }
+
+  private addLabmdaMonitorFunction(props: LambdaMonitorFunctionProps): LambdaFunction{
```
horsmand:
```suggestion
  private addLambdaMonitorFunction(props: LambdaMonitorFunctionProps): LambdaFunction{
```

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +270,91 @@ abstract class X509CertificateBase extends Construct {
...
+    const lambdaMonitor = new LambdaFunction(this, 'MonitoringExpDate', {
+      description: `Used by a X509Certificate ${Names.uniqueId(this)} to monitor certificate expiration date.`,
+      code: props.lambdaCode,
+      environment: {
+        UNIQUEID: props.uniqueValue,
+      },
+      runtime: Runtime.NODEJS_12_X,
+      layers: props.lambdaLayers,
+      handler: 'cert-rotation-monitor.handler',
```
horsmand: It seems a bit odd to have the code the monitor takes as a variable but then have the handler it runs as a hardcoded string. Let's follow the pattern that the class hierarchy is already using and pass this handler in as a `lambdaMonitorHandler` field from the concrete classes to the base class.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +270,91 @@ abstract class X509CertificateBase extends Construct {
...
+  }
+
+  private addLabmdaMonitorFunction(props: LambdaMonitorFunctionProps): LambdaFunction{
+    const lambdaMonitor = new LambdaFunction(this, 'MonitoringExpDate', {
+      description: `Used by a X509Certificate ${Names.uniqueId(this)} to monitor certificate expiration date.`,
+      code: props.lambdaCode,
+      environment: {
+        UNIQUEID: props.uniqueValue,
+      },
+      runtime: Runtime.NODEJS_12_X,
```
horsmand: Node 12 is coming up on EOL early next year. Maybe we should take the opportunity to upgrade this one and the actual Lambda to `Runtime.NODEJS_14_X`.

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/certificate.ts
```diff
@@ -90,6 +90,40 @@ export class Certificate implements ICertificate {
     }
   }
 
+  public static async getExpDate(
```
horsmand, line 93: This should have unit testing

packages/aws-rfdk/lib/lambdas/nodejs/cert-rotation-monitor/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+function putValueToMetric(value: number){
+  const metric = {
+    MetricData: [
+      {
+        MetricName: 'DaysToExpiry',
+        Dimensions: [
+          {
+            Name: 'Certificate Metrics',
+            Value: process.env.UNIQUEID!,
```
horsmand: Can we pass `process.env.UNIQUEID` into this function like you did for `getSecretsByTag`?

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +270,91 @@ abstract class X509CertificateBase extends Construct {
...
+    return lambdaMonitor;
+  }
+
+  private addExpDateValidation(props: X509CertValidationProps){
+    const certMetric = new Metric({
+      metricName: 'DaysToExpiry',
+      namespace: 'AWS/RFDK',
+      dimensions: {
+        'Certificate Metrics': props.uniqueValue,
+      },
```
horsmand: Can we share `metricName`, `namespace`, and `dimensionName` between this construct and the Lambda code where it also appears? I'm thinking they could be static constants in the Lambda code that we reference here.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +270,91 @@ abstract class X509CertificateBase extends Construct {
...
+      tagCondition,
+    });
+    this.addExpDateValidation({
+      uniqueValue,
+      alarmEmailAddress: props.alarmEmailAddress,
+    });
+  }
+
+  private addLabmdaMonitorFunction(props: LambdaMonitorFunctionProps): LambdaFunction{
+    const lambdaMonitor = new LambdaFunction(this, 'MonitoringExpDate', {
```
horsmand, line 292: The cost, security considerations, and resources deployed sections for both the X509CertificatePem and X509CertificatePkcs12 should be updated to include information about the lambda and monitor.

We might need to update the diagram for this as well.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +270,91 @@ abstract class X509CertificateBase extends Construct {
...
+
+  private addExpDateValidation(props: X509CertValidationProps){
+    const certMetric = new Metric({
+      metricName: 'DaysToExpiry',
+      namespace: 'AWS/RFDK',
+      dimensions: {
+        'Certificate Metrics': props.uniqueValue,
+      },
+      // One 99-th percentile data point hour
+      period: Duration.days(1),
```
horsmand: The comment says "data point hour" but then the Duration is `Duration.days(1)`. Is one of those incorrect? It looks like the Lambda that creates the metric runs daily, so maybe we just need to update the comment to say one data point per day.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +270,91 @@ abstract class X509CertificateBase extends Construct {
...
+      statistic: 'p99',
```
horsmand: How come you went with a percentile here? Shouldn't we only be collecting one data point per day? Since the value of the metric is the number of days until expiry, maybe we should use `Minimum` here.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +270,91 @@ abstract class X509CertificateBase extends Construct {
...
+    if (props.alarmEmailAddress) {
+      const snsTopic = new Topic(this, 'CertificateExpiryTopic');
+      snsTopic.addSubscription(new EmailSubscription(props.alarmEmailAddress));
+      const alarmAction = new SnsAction(snsTopic);
+      alarm.addAlarmAction(alarmAction);
+    }
+
+    new Rule(this, 'ScheduleSecretMonitoring', {
+      description: `Used for monitoring X509Certificate ${Names.uniqueId(this)} expiration date`,
+      schedule: Schedule.rate(Duration.days(1)),
```
horsmand, line 361: Did we figure out what this would cost to run every day indefinitely? It might make sense to run it weekly and set the threshold to a month, or make it configurable based on the `validFor` value. The `validFor` field defaults to 3 years, so we'd be fine to run weekly and notify with a month left in that case, but maybe if they set it for less than a year we could run daily and notify with 15 days left.

packages/aws-rfdk/lib/lambdas/nodejs/cert-rotation-monitor/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Lambda for adding metric about Secret expiration date.
+ *
+ */
+export async function handler(): Promise<void> {
+  const secretsManager = new SecretsManager({ apiVersion: SECRETS_MANAGER_VERSION });
+  const secretsList = await getSecretsByTag(secretsManager, process.env.UNIQUEID!);
+  if (secretsList) {
+    for (const secret of secretsList) {
```
horsmand: Do we ever expect multiple certificates to be attached to a single tag? Looking at the code, it didn't look possible to me. If that's the case, maybe we should check if this list contains <= 1 items.

Thinking it through, if we switch the statistic from `p99` to `minimum` like I suggested in a previous comment, maybe we can leave this to iterate through the items. I think leaving it at `p99` will be a problem though, because if for some reason it did emit 2 metrics for 2 expiry dates, I think p99 would end up looking at the higher value.

horsmand (changes requested):

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
kozlove-aws: As I know Lambda price is extremely low. Running once  a day can guarantee that alarm will be raised in time even when customer replace this certificate manually.

packages/aws-rfdk/lib/lambdas/nodejs/cert-rotation-monitor/index.ts
kozlove-aws: It is possible if will Lambda generate more than one certificate but we do not expect it so we do not want to have any exceptions in this case. This code will work if we receive one ore more secrets with certificates to check.

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/test/certificate.test.ts
```diff
@@ -275,3 +275,22 @@ test('decrypt private key', async () => {
...
+  const validFor = 100;
+  const certificate = await Certificate.fromGenerated(name, passphrase, validFor);
+
+  // WHEN
+  const expDate = await Certificate.getExpDate(certificate.cert);
+
+  // THEN
+  expect(expDate).toBeDefined();
+  expect(Math.ceil((expDate!.getTime() - (new Date()).getTime())/(1000 * 3600 * 24))).toEqual(validFor);
+});
```
horsmand: Missing blank line

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/certificate.ts
```diff
@@ -90,6 +90,40 @@ export class Certificate implements ICertificate {
...
+    try {
+      const certificateFile: string = path.join(tmpDir, 'crt');
+      await writeAsciiFile(certificateFile, certificate);
+      const cmd =
+        'openssl x509 ' +
+        '-enddate -noout ' +
+        `-in ${certificateFile} ` +
+        '| cut -d= -f2';
+
+      console.debug(`Running: ${cmd}`);
```
horsmand, line 106: What's the default log level for a Lambda? Will these debug lines always appear or should we change them to `console.log`?

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -229,6 +273,92 @@ abstract class X509CertificateBase extends Construct {
...
+    }));
+    return lambdaMonitor;
+  }
+
+  private addExpDateValidation(props: X509CertValidationProps){
+    const certMetric = new Metric({
+      metricName: CertificateRotationMonitor.METRIC_NAME,
+      namespace: CertificateRotationMonitor.METRIC_NAMESPACE,
+      dimensions: {},
+      // Minimum data point daily
```
horsmand: This comment doesn't really add any value beyond what the code already says, maybe we could change it to:
```suggestion
      // We're looking at the minimum value reported by the Lambda in a day, which should
      // be the only value, since it runs once daily.
```

packages/aws-rfdk/lib/lambdas/nodejs/cert-rotation-monitor/test/handler.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    // WHEN
+    // eslint-disable-next-line dot-notation
+    certificateRotationMonitor['putValueToMetric'](1);
+
+    // THEN
+    expect(consoleLogMock.mock.calls.length).toBe(2);
+    expect(consoleLogMock.mock.calls[1][0]).toBe(`Failed to put value to CertDaysToExpiry metric: ${err.message}`);
+  });
+});
```
horsmand: Missing blank line

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
horsmand: Any reason not to upgrade the other one too?

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
horsmand: I don't see this info added.

packages/aws-rfdk/lib/lambdas/nodejs/cert-rotation-monitor/index.ts
horsmand: If we do leave the logic like this where we are looping over the list, but don't expect it to have more than 1 item, maybe we should have a comment saying that we don't expect multiple items, but the monitor should still work if there is more than one since we're alarming on the minimum value.

packages/aws-rfdk/lib/lambdas/nodejs/lib/x509-certs/certificate.ts
kozlove-aws: Added debug mode, switched off by default.

packages/aws-rfdk/lib/deadline/test/render-queue.test.ts
```diff
@@ -2494,9 +2494,9 @@ describe('RenderQueue', () => {
       resourceTypeCounts: {
         'AWS::ECS::Cluster': 1,
         'AWS::EC2::SecurityGroup': 2,
-        'AWS::IAM::Role': 10,
+        'AWS::IAM::Role': 14,
         'AWS::AutoScaling::AutoScalingGroup': 1,
-        'AWS::Lambda::Function': 6,
+        'AWS::Lambda::Function': 10,
```
jusiskin, line 2499: I'm wondering if we want to create a Lambda function per certificate. In this test, we create a `RenderQueue` without providing any certificates which appears to create 4 certificates and therefor 4 Lambda functions and 4 IAM roles. This will bloat a CFN template and detract from its template [resource/file-size limits](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html). Currently there is a limit of 500 resources per template and 51.2KB inline / 1MB S3 max CFN template size. I believe CDK uploads templates to the CDK bootstrap S3 bucket, so we are dealing with a 1MB template size limit.

I wonder if there's a way we could have a single Lambda function and CloudWatch Events rule per stack, aggregate the cert unique IDs and then have the lambda loop over them all.

As a CDK construct library, RFDK should be a good citizen and try to minimize what we contribute to our end-user's CFN template limits. What do you think?

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -230,6 +274,93 @@ abstract class X509CertificateBase extends Construct {
...
+    });
+
+    if (props.alarmEmailAddress) {
+      const snsTopic = new Topic(this, 'CertificateExpiryTopic');
+      snsTopic.addSubscription(new EmailSubscription(props.alarmEmailAddress));
+      const alarmAction = new SnsAction(snsTopic);
+      alarm.addAlarmAction(alarmAction);
+    }
+
+    new Rule(this, 'ScheduleSecretMonitoring', {
```
jusiskin, line 359: Can we change `ScheduleSecretMonitoring` &rarr; `ExpirationCheckRule`?

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -272,7 +404,9 @@ export class X509CertificatePem extends X509CertificateBase implements IX509Cert
     super(scope, id, {
       lambdaCode: Code.fromAsset(join(__dirname, '..', '..', 'lambdas', 'nodejs')),
       lambdaHandler: 'x509-certificate.generate',
+      lambdaMonitorHandler: 'cert-rotation-monitor.handler',
```
jusiskin, line 407: we can probably just hard-code this in `X509CertificateBase.addLambdaMonitorFunction`, because right now its duplicated here and in `X509CertificatePkcs12`.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -251,10 +382,11 @@ abstract class X509CertificateBase extends Construct {
  * - DynamoDB Table - Used for tracking resources created by the Custom Resource.
  * - Secrets - 4 in total, for the certificate, it's private key, the passphrase to the key, and the cert chain.
  * - Lambda Function, with role - Used to create/update/delete the Custom Resource
+ * - Lambda Function, Metric and Alarm for monitoring and alarm about expiration date.
```
jusiskin, line 385: This change also deploys:

* IAM role
* IAM policy
* CloudWatch log group
* CloudWatch events rule
* SNS topic
* SNS subscription

according to a quick `cdk diff` I ran.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -143,9 +167,27 @@ export interface IX509CertificatePem extends IConstruct {
...
+  readonly lambdaMonitorHandler: string;
+  readonly tagCondition?: {
+    [key: string]: any;
+  };
+}
+
+interface X509CertValidationProps {
+  readonly uniqueValue: string;
+  readonly alarmEmailAddress?: string;
+}
```
jusiskin, line 188: it might help for code readability to move these interfaces to be defined inline into the function definition where they're used:

```ts
private addLambdaMonitorFunction(props: {
    readonly uniqueValue: string;
    readonly lambdaLayers: ILayerVersion[];
    readonly lambdaCode: Code;
    readonly lambdaMonitorHandler: string;
    readonly tagCondition?: {
      [key: string]: any;
    };
  }): LambdaFunction{
```

this keeps the props definition closer to the code where it is used. What do you think?

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -230,6 +274,93 @@ abstract class X509CertificateBase extends Construct {
...
+      ],
+      resources: ['*'],
+      conditions: {
+        StringEquals: { 'cloudwatch:namespace': CertificateRotationMonitor.METRIC_NAMESPACE },
+      },
+    }));
+    return lambdaMonitor;
+  }
+
+  private addExpDateValidation(props: X509CertValidationProps){
```
jusiskin, line 331: nit: can we rename this to `addExpirationDateTracking`?

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -230,6 +274,93 @@ abstract class X509CertificateBase extends Construct {
...
+      metricName: CertificateRotationMonitor.METRIC_NAME,
+      namespace: CertificateRotationMonitor.METRIC_NAMESPACE,
+      dimensions: {},
+      // We're looking at the minimum value reported by the Lambda in a day, which should
+      // be the only value, since it runs once daily.
+      period: Duration.days(1),
+      statistic: 'Minimum',
+    });
+
+    certMetric.dimensions![CertificateRotationMonitor.METRIC_DIMENSION] = props.uniqueValue;
```
jusiskin, line 342: I think we can simplify this to:

```ts
    const certMetric = new Metric({
      metricName: CertificateRotationMonitor.METRIC_NAME,
      namespace: CertificateRotationMonitor.METRIC_NAMESPACE,
      dimensions: {
        [CertificateRotationMonitor.METRIC_DIMENSION]: props.uniqueValue,
      },
      // We're looking at the minimum value reported by the Lambda in a day, which should
      // be the only value, since it runs once daily.
      period: Duration.days(1),
      statistic: 'Minimum',
    });
```

packages/aws-rfdk/lib/core/test/x509-certificate.test.ts
```diff
@@ -18,12 +18,12 @@ import { Key } from '@aws-cdk/aws-kms';
 import { CfnSecret } from '@aws-cdk/aws-secretsmanager';
 import { Stack } from '@aws-cdk/core';
 
+import { CertificateRotationMonitor } from '../../lambdas/nodejs/cert-rotation-monitor/handler';
```
jusiskin, line 21: it looks like we're missing tests for some of the new resources deployed by this code:

*   `AWS::SNS::Topic`
*   `AWS::IAM::Role`
*   `AWS::Events::Rule`
*   `Custom::LogRetention`

packages/aws-rfdk/lib/lambdas/nodejs/cert-rotation-monitor/handler.ts
```diff
@@ -0,0 +1,103 @@
...
+    };
+    const cloudwatch = new CloudWatch();
+    cloudwatch.putMetricData(metric, (err) => {
+      if (err) {
+        console.log(`Failed to put value to CertDaysToExpiry metric: ${err.message}`);
+      } else {
+        console.log('Value was added to CertDaysToExpiry metric.');
+      }
+    });
+  }
```
jusiskin, line 69: This should be changed into an `async` function, since `cloudwatch.putMetricData` is asynchronous and we want it to complete before the Lambda function exits.

packages/aws-rfdk/lib/core/test/x509-certificate.test.ts
```diff
@@ -324,10 +325,87 @@ test('Generate cert, all options set', () => {
...
+    Dimensions: [
+      {
+        Name: CertificateRotationMonitor.METRIC_DIMENSION,
+        Value: 'b2b09a6086e87fe14005f4e0b800e4f0',
+      },
+    ],
+    MetricName: CertificateRotationMonitor.METRIC_NAME,
+    Namespace: CertificateRotationMonitor.METRIC_NAMESPACE,
+    Period: 86400,
+    TreatMissingData: 'notBreaching',
```
jusiskin, line 403: should we alarm on missing data? that means that our Lambda is not publishing metrics correctly, right?

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -230,6 +274,93 @@ abstract class X509CertificateBase extends Construct {
...
+    const alarm = certMetric.createAlarm(this, 'CertificateExpiryAlarm', {
+      evaluationPeriods: 1,
+      threshold: 15,
+      comparisonOperator: ComparisonOperator.LESS_THAN_THRESHOLD,
+      treatMissingData: TreatMissingData.NOT_BREACHING,
+      actionsEnabled: props.alarmEmailAddress ? true : false,
+    });
+
+    if (props.alarmEmailAddress) {
+      const snsTopic = new Topic(this, 'CertificateExpiryTopic');
```
jusiskin, line 353: We should encrypt this SNS topic by default. It would be nice to be able to supply a KMS key in the props so we don't always create a KMS key per-certificate.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -230,6 +274,93 @@ abstract class X509CertificateBase extends Construct {
...
+
+    certMetric.dimensions![CertificateRotationMonitor.METRIC_DIMENSION] = props.uniqueValue;
+
+    const alarm = certMetric.createAlarm(this, 'CertificateExpiryAlarm', {
+      evaluationPeriods: 1,
+      threshold: 15,
+      comparisonOperator: ComparisonOperator.LESS_THAN_THRESHOLD,
+      treatMissingData: TreatMissingData.NOT_BREACHING,
+      actionsEnabled: props.alarmEmailAddress ? true : false,
+    });
```
jusiskin, line 350: Can we add an `alarmDescription` here? I suspect this description will be included in the email notification. We should probably provide or link to the action that the recipient should take to rotate the certificates. See the following example in our example app for setting up similar metric/alarm/notification for EFS burst credits:

https://github.com/aws/aws-rfdk/blob/0933a9e5a569d0f03a266503bfdf59206a2ce41d/examples/deadline/All-In-AWS-Infrastructure-Basic/ts/lib/storage-tier.ts#L200-L285

packages/aws-rfdk/lib/lambdas/nodejs/cert-rotation-monitor/handler.ts
```diff
@@ -0,0 +1,103 @@
...
+      MetricData: [
+        {
+          MetricName: CertificateRotationMonitor.METRIC_NAME,
+          Dimensions: [
+            {
+              Name: CertificateRotationMonitor.METRIC_DIMENSION,
+              Value: this.uniqueID,
+            },
+          ],
+          Unit: 'None',
```
jusiskin, line 55: nit: AWS Certificate Manager uses the "Count" unit for their DaysToExpiry metric. Can we do the same?

packages/aws-rfdk/lib/lambdas/nodejs/cert-rotation-monitor/handler.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import { Secret } from '../lib/secrets-manager';
+import { Certificate } from '../lib/x509-certs';
+
+const SECRETS_MANAGER_VERSION = '2017-10-17';
+
+export class CertificateRotationMonitor {
+  public static METRIC_NAME = 'DaysToExpiry';
+  public static METRIC_NAMESPACE = 'AWS/RFDK';
+  public static METRIC_DIMENSION = 'Certificate Metrics';
```
jusiskin, line 19: Can we change this to "Certificate ARN"?

EDIT: Or perhaps "Certificate ID"? Is it possible to feed the cert ARN into this lambda instead? Because that might be easier for an end-user to reason about.

packages/aws-rfdk/lib/core/lib/x509-certificate.ts
```diff
@@ -230,6 +274,93 @@ abstract class X509CertificateBase extends Construct {
...
+      lambdaLayers: [openSslLayer],
+      lambdaCode: props.lambdaCode,
+      tagCondition,
+      lambdaMonitorHandler: props.lambdaMonitorHandler,
+    });
+    this.addExpDateValidation({
+      uniqueValue,
+      alarmEmailAddress: props.alarmEmailAddress,
+    });
+  }
```
jusiskin, line 289: I tested this with an [`X509CertificatePkcs12`](https://docs.aws.amazon.com/rfdk/api/latest/docs/aws-rfdk.X509CertificatePkcs12.html) which derives from this base class, and found that it does not emit any metric. I'm not certain if this is because:
*   PKCS#12 certs don't have expiration
*   The command you are using to extract the certificate expiry does not work with PKCS#12 certificates
*   Something else...

That being said, based on the construct's API, PKCS#12 are always converted from another `X509CertificatePem` which will have its own metric/alarm. Depending on how RFDK instructs rotating these certs, we can probably do with out metrics/alarms on `X509CertificatePkcs12`. Should we move this logic into `X509CertificatePem` instead?

jusiskin (changes requested):
Great solution to this problem. There are a few things we need to address before we can merge this.

PR #652, opened 2022-01-04
chore: bump cdk to 1.137.0

horsmand:
This took some work to get RFDK to work with the new version of CDK:
- Added a new abstract Resource class to the RFDK core package that
  implements IResource. This was required because a new
  applyRemovalPolicy method was added to the IResource interface and all
  of our constructs that were implementing it broke. Now these
  constructs extend the new Resource class which implements the
  applyRemovalPolicy method the same way that CDK has implemented it in
  their Resource class. I also tried swapping our constructs to extend
  the CDK Resource class, but this caused issues as CDK expects classes
  extending that class to be backed by L1 resources.
- Updated licenses and notices to use 2022
- Replaced `Table#serverSideEncyption: true` to `Table#encryption:
  TableEncryption.AWS_MANAGED`
- Replaced `Metric#dimensions` with `Metric#dimenionsMap`
- Replaced `Secret#secretArn` with either `Secret#secretCompleteArn` or
  `Secret#secretPartialArn`
- Removed `Key#trustAccountIdentities: true` for the feature flag
  `defaultKeyPolicies: true` and replaced all our cdk.json files with
  the current CDK defaults. Details on the feature flags are here:
  https://github.com/aws/aws-cdk/blob/cd51a5dae1780e34aecd90d85783fb6d3c239903/packages/@aws-cdk/cx-api/lib/features.ts
- Updated examples to use AMI for Deadline 10.1.20.2
- Set jest to use the `--silent` flag so it didn't write a bunch of output
  to the console.

----


examples/deadline/All-In-AWS-Infrastructure-Basic/python/cdk.json
```diff
@@ -1,3 +1,18 @@
 {
-  "app": "python -m package.app"
+  "app": "python -m package.app",
+  "context": {
```
ddneilson, line 3: Are customer's deployments going to break if they don't have these defined in their app's `cdk.json` file? i.e. Did CDK change default values?

packages/aws-rfdk/lib/core/lib/resource.ts
```diff
@@ -0,0 +1,53 @@
...
+   *
+   * The Removal Policy controls what happens to this resource when it stops
+   * being managed by CloudFormation, either because you've removed it from the
+   * CDK application or because you've made a change that requires the resource
+   * to be replaced.
+   *
+   * The resource can be deleted (`RemovalPolicy.DESTROY`), or left in your AWS
+   * account for data recovery and cleanup later (`RemovalPolicy.RETAIN`).
+   */
+  public applyRemovalPolicy(policy: RemovalPolicy) {
```
ddneilson: If all Resources in CDK are L1 constructs, then is this implementation correct for RFDK (where our constructs are L3, generally)? Specifically, our constructs can (and do) have more than one child; I wonder if it's actually sufficient to only look at the defaultChild...

package.json
```diff
@@ -26,87 +26,87 @@
...
     "constructs": "^3.3.69",
     "fs-extra": "^10.0.0",
-    "jest": "^26.6.3",
-    "jest-junit": "^11.1.0",
-    "jsii-diff": "^1.40.0",
-    "jsii-pacmak": "^1.40.0",
+    "jest": "^27.4.5",
+    "jest-junit": "^13.0.0",
+    "jsii-diff": "^1.49.0",
+    "jsii-pacmak": "^1.49.0",
```
jusiskin: There are a few conflicting JSII versions being used.

*   Here we use `^1.49.0`
*   In `examples/deadline/Local-Zone/python/setup.py`, we're using jsii `1.50.0`.
*   In `tools/cdk-build-tools/package.json`, we use jsii `^1.50.0`

These versions should probably be internally consistent and match. 

Historically, I think we've tried to match the JSII version used by our compatible CDK version. Looking at [CDK's `tools/@aws-cdk/cdk-build-tools/package.json` for `v1.137.0`](https://github.com/aws/aws-cdk/blob/bfbdf64696a05a9a1b9dcad30904b17ff1a83f8c/tools/%40aws-cdk/cdk-build-tools/package.json#L59-L61), they appear to use JSII `^1.47.0` which is equivalent to the latest `1.47.x`. That being said, I recall from offline discussions that 1.50.0 has some critical performance improvements and without them the build takes incredibly long.

Assuming everything works when using a JSII package, should we bump this version to `^1.50.0`?

jusiskin (changes requested):
Beyond echoing @ddneilson's feedback about the `Resource` inheritance, I noticed one version mis-match. Otherwise, it looks good.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/cdk.json
horsmand: Hmm, that's a good point. CDK Did change their default values. I have copied over the settings from the `cdk.json` that gets generated when using `cdk init`.

This shouldn't break any existing deployments that don't have these flags, but if they don't have the `defaultKeyPolicies: true` flag, the change to the aws-kms construct `Key` where I've removed the deprecated `trustAccountIdentities` field would affect the `Key`'s policy.

package.json
horsmand: Oh thanks, I don't know how I missed that. I had originally updated everything to use 1.49.0, but there was a bug in it that made build times extremely long, like 30+ minutes for just `packages/aws-rfdk`, so I had to bump again to 1.50.0. I'll update this to use `^1.50.0`.

packages/aws-rfdk/lib/core/lib/resource.ts
horsmand: After speaking with Josh, he raised the question of why some of our constructs even needed to implement `IResource`. It doesn't seem like they do. I've removed the `Resource` construct I wrote and removed the `implements IResource` from all our classes that were using it. I was able to build and synthesize all our examples, so it seems like we don't need to.

Let me know what you think, the only concern I have with the update I made is that I removed the `stack` and `env` properties along with the interface, since they were requirements of the `IResource` interface. They can be retrieved for any construct without needing to store them as construct properties, and it seems really unlikely to me that anybody would have been using them, but it's a breaking change to remove them.

horsmand: I've run the integration tests against the updated code and also packaged it and dropped it into a StudioBuilder app and made sure it could synth without any issues.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/cdk.json
ddneilson: That sounds like a breaking change to me, then. We didn't have that context value set in our examples before, so I wouldn't expect customers to have it set either. 

`trustAccountIdentities` is deprecated but not removed, right? I'd think that it would be a breaking change in a stable API for the CDK to remove it in the 1.xx line. What's the effect of us leaving in the usage of the deprecated field? Just a warning? If so, then I think that we leave setting that field in place... we can add these context values as well, but I think that we want to be more careful with breaking changes.

packages/aws-rfdk/lib/core/test/health-monitor.test.ts
```diff
@@ -135,7 +135,23 @@ describe('HealthMonitor', () => {
       KeyPolicy: {
         Statement: [
           {
-            Action: 'kms:*',
+            Action: [
```
ddneilson: It's curious to me that this changed. Implies that we're now going down the `addLegacyAdminPolicy()` branch of https://github.com/aws/aws-cdk/blob/6359c12e3242e23d9b3bf0a42cac7c361c8d4d8a/packages/%40aws-cdk/aws-kms/lib/key.ts#L637

Looks like it's safe to update the policy ( https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kms-key.html#cfn-kms-key-keypolicy ) -- doesn't cause a replacement.

I'd guess this is related to `@aws-cdk/aws-kms:defaultKeyPolicies` feature flag (what version of CDK was that introduced in?). Since the `trustAccountIdentities` flag was removed from this KMS key, this implies to me that the tests aren't being run with that feature flag set.

ddneilson (changes requested):
From the 14.x build log:
```
warning " > @aws-cdk/aws-certificatemanager@1.137.0" has unmet peer dependency "@aws-cdk/aws-acmpca@1.137.0".
warning " > aws-rfdk@0.39.0" has unmet peer dependency "@aws-cdk/aws-acmpca@1.137.0".
...
aws-rfdk: warning JSII6: A "peerDependency" on "constructs" at "^3.3.69" means you should take a "devDependency" on "constructs" at "3.3.69" (found "^3.3.69")"
```

Looks like we need to add another peer dep and fixup the dev dep for constructs.

packages/aws-rfdk/lib/core/lib/resource.ts
ddneilson: Sounds good to me. I don't recall why we made them `IResource` instead of `IConstruct` originally.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/cdk.json
horsmand: The messaging of the warning said it would be removed in the next major release, so maybe it won't actually be removed in the 1.x.x versions, but the warning will always be outputted. I'll add it back and I guess people will just have to deal with the warnings that come out of the more recent 1.x.x releases.

horsmand:
> From the 14.x build log:
> 
> ```
> warning " > @aws-cdk/aws-certificatemanager@1.137.0" has unmet peer dependency "@aws-cdk/aws-acmpca@1.137.0".
> warning " > aws-rfdk@0.39.0" has unmet peer dependency "@aws-cdk/aws-acmpca@1.137.0".
> ...
> aws-rfdk: warning JSII6: A "peerDependency" on "constructs" at "^3.3.69" means you should take a "devDependency" on "constructs" at "3.3.69" (found "^3.3.69")"
> ```
> 
> Looks like we need to add another peer dep and fixup the dev dep for constructs.

Thanks, I added acmpca to the package.json in aws-rfdk package, but missed adding it to the base package.json. Will add there too.

EDIT: As for the constructs message, Josh and I worked on trying to get that removed the last time we did a release and it isn't possible. It started showing up when we duplicated all the CDK dependencies into the devDependencies section of the aws-rfdk package.

horsmand:
Just for reference, this is the output of `cdk synth --all` in the basic example app:
```
[WARNING] @aws-cdk/aws-autoscaling.CommonAutoScalingGroupProps#updateType is deprecated.
  Use `updatePolicy` instead
  This API will be removed in the next major release.
[WARNING] @aws-cdk/aws-ecs.EcsOptimizedAmi is deprecated.
  see {@link EcsOptimizedImage#amazonLinux}, {@link EcsOptimizedImage#amazonLinux} and {@link EcsOptimizedImage#windows}
  This API will be removed in the next major release.
[WARNING] @aws-cdk/aws-autoscaling.CommonAutoScalingGroupProps#updateType is deprecated.
  Use `updatePolicy` instead
  This API will be removed in the next major release.
[WARNING] @aws-cdk/aws-ecs.EcsOptimizedAmi#getImage is deprecated.
  see {@link EcsOptimizedImage#amazonLinux}, {@link EcsOptimizedImage#amazonLinux} and {@link EcsOptimizedImage#windows}
  This API will be removed in the next major release.
[WARNING] @aws-cdk/aws-ecs.Cluster#addAutoScalingGroup is deprecated.
  Use {@link Cluster.addAsgCapacityProvider} instead.
  This API will be removed in the next major release.
{
  sourceArn: '${Token[TOKEN.1318]}',
  principal: <ref *1> ServicePrincipal {
    grantPrincipal: [Circular *1],
    principalAccount: undefined,
    assumeRoleAction: 'sts:AssumeRole',
    service: 'sns.amazonaws.com',
    opts: {}
  }
}
[WARNING] @aws-cdk/aws-kms.KeyProps#trustAccountIdentities is deprecated.
  redundant with the `@aws-cdk/aws-kms:defaultKeyPolicies` feature flag
  This API will be removed in the next major release.
{
  sourceArn: '${Token[TOKEN.1641]}',
  principal: <ref *1> ServicePrincipal {
    grantPrincipal: [Circular *1],
    principalAccount: undefined,
    assumeRoleAction: 'sts:AssumeRole',
    service: 'sns.amazonaws.com',
    opts: {}
  }
}
```

examples/deadline/All-In-AWS-Infrastructure-Basic/python/cdk.json
ddneilson: Yeah, I think that the warning about being removed in the next major release is fine. That's CDK 2.x. The update of RFDK to that version will be a bit more involved, so it makes sense to me to address this at that time.

packages/aws-rfdk/lib/core/test/health-monitor.test.ts
horsmand: I've changed this back to how it was originally, I believe you're right that it seems we aren't applying the feature flags to the tests but that seems like something we can deal with if/when we upgrade to CDK v2

package.json
jusiskin: Looks good now. Resolving

packages/aws-rfdk/lib/core/lib/resource.ts
jusiskin: Like you said, it is very unlikely that anyone used these properties, but do you think we should mark this as an official breaking change?

If so, then we could add upgrade documentation and before merging we should squash and edit the commit message to include a conventional commits breaking change. What do you think?

ddneilson (dismissed):

jusiskin (approved):

ddneilson (approved):

PR #682, opened 2022-03-31
chore(integ): fixes random failures in integration tests

ddneilson:
This PR fixes some random integration test failures that we have been seeing in daily canaries. 
1. Some deployments take more than 1 hr, randomly, and that results in expired AWS credentials being used partway through and a test failure.
2. We randomly have tests fail due to insufficient IP addresses available in a subnet.

To fix (1), this PR:
1. Changes the automated integ e2e test from explicitly calling sts assume-role to, instead, set up an assume-role credential provider as a profile and then run the tests with that profile.
2. The CDK works great with that change, but the AWS SDK for Javascript V2 does not support the flow. Fortunately, the AWS SDK for Javascript V3 does. So, I updated the JS in the integ tests to use V3 of the SDK.

To fix (2), this PR simply increases some CIDR ranges in the infrastructure component of the integration tests.

Testing:
1. Deployed a mirror of our integration testing canary. Ran the tests through it, and that showed that the CDK and new JS in integ tests could pick up the credentials.
2. The above had a test failure due to some bad translation, on my part, for the JS in some of the jest tests in integ. To fix those I deployed the integ tests to a personal account and ran the jest tests. They all pass now.

----


jusiskin (approved):
LGTM. Great work!

jericht (approved):
LGTM!

PR #685, opened 2022-04-01
fix(lambda-layers): add ap-southeast-3 to region deny list

ddneilson:
The `ap-southeast-3` region has been added to AWS as an opt-in region. (See "Available Regions" under https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html ).

This PR adds this region to the region deny list for building lambda layers.

----


yashda (approved):
Changes looks good. Thanks Daniel!

jericht (approved):

PR #738, opened 2022-07-27
feat(all): migrate to CDKv2

ddneilson:
* Migrates the RFDK library, examples, and integration tests to CDKv2.

BREAKING CHANGE: This change migrates the RFDK to be based on CDK v2.
All apps that use the RFDK will have to be migrated to CDKv2 as well.
To update your app, follow the CDK's migration guide at:
https://docs.aws.amazon.com/cdk/v2/guide/migrating-v2.html

---

Some notes:
1. Most of the changes are pretty superficial. Migrating application code from CDKv1 to CDKv2 really just entails changing imports, and updating any uses of construct properties that were deprecated in v1 and removed in v2. We've been pretty good about staying on top of updating uses of deprecated properties, so it was mostly import updates for us.
2. CDKv2 removed the `@aws-cdk/asserts` module that we were using for unit tests and replaced it with `aws-cdk-lib/assertions`. This was a major change. Syntax and functionality changed. The CDK team provides a migration guide (here: https://github.com/aws/aws-cdk/blob/v1-main/packages/@aws-cdk/assertions/MIGRATING.md ) that includes a migration script; the script got about 70% of the way, and the rest had to be done manually. The biggest change was that once you've asserted properties of a stack/app in a test then the state of the stack is baked-in at that time. So, you cannot, say, assert a property, then add something to the stack, and then assert a new property -- the second assertion will fail as it will be asserting against the stack pre-addition; this change required retooling some unit tests.
3. I cleaned up the integration tests a bit; mostly making sure that the test logs were more find-able, and fixing a resource leak (mongoDB test was leaking an EBS volume on every run).
4. The migration is to CDK v2.33.0 as that is the current version as of right now.

----


packages/aws-rfdk/lib/core/test/cloudwatch-agent.test.ts
```diff
@@ -60,32 +58,10 @@ describe('CloudWatchAgent', () => {
...
 
     // THEN
-    expectCDK(stack).to(haveResource('AWS::SSM::Parameter', {
+    Template.fromStack(stack).hasResourceProperties('AWS::SSM::Parameter', {
       Type: 'String',
       Value: cloudWatchConfig,
-    }));
-  });
-
-  test('creates an asset', () => {
```
ddneilson, line 69: Note: This test is not possible with the under-the-hood changes introduced in CDKv2. Fortunately, this is tested, in a way, by the tests that look at UserData; the CDK now generates the file sha256 before synthesis, so file hashes are directly embedded into UserData in both the rendered stack and when directly rendering the UserData string from the contruct.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/config.py
```diff
@@ -46,6 +46,7 @@ def __init__(self):
         self.deadline_client_linux_ami_map: Mapping[str, str] = {'us-west-2': 'ami-04ae356533dc07fb5'}
 
         # A secret (in binary form) in SecretsManager that stores the UBL certificates in a .zip file.
+        # This must be a complete ARN, including the "-<random text>" suffix.
```
kozlove-aws: including the random text a little bit confusing as for me.
```suggestion
        # This must be in format `arn:aws:secretsmanager:<Region>:<AccountId>:secret:<SecretName>-<6RandomCharacters>`
```

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/storage_tier.py
```diff
@@ -231,8 +231,7 @@ def add_low_efs_burst_credit_alarms(self, filesystem: FileSystem, email_address:
             'SNSEncryptionKey',
             description='Used to encrypt the SNS Topic for sending EFS Burst Credit alerts',
             enable_key_rotation=True,
-            removal_policy=RemovalPolicy.DESTROY,
-            trust_account_identities=True
```
kozlove-aws, line 235: Why is this was removed?

kozlove-aws (commented):
Great job, especially with making all tests work.
Just a couple moment I want to clarify.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/lib/storage_tier.py
ddneilson: The property was removed in CDKv2. `trust_account_identites=True`  is now the default and only option.

examples/deadline/All-In-AWS-Infrastructure-Basic/python/package/config.py
ddneilson: Good suggestion! I'll make that change.

edit: Change made in latest commit push.

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -338,17 +339,19 @@ test('repository installer iam permissions: installer get', () => {
   });
 
   // THEN
-  expectCDK(stack).to(haveResourceLike('AWS::IAM::Policy', {
+  //const res = Template.fromStack(stack).findResources('AWS::IAM::Policy', Match.anyValue);
+  //expect(res).toBe({ foo: 'bar' });
```
jericht: nit: commented out code

packages/aws-rfdk/lib/deadline/test/spot-event-plugin-fleet.test.ts
```diff
@@ -31,23 +33,17 @@ import {
...
-} from '@aws-cdk/aws-iam';
-import { ArtifactMetadataEntryType } from '@aws-cdk/cloud-assembly-schema';
-import {
-  App,
-  CfnElement,
-  Stack,
-  Tags,
-} from '@aws-cdk/core';
+} from 'aws-cdk-lib/aws-iam';
+//import { ArtifactMetadataEntryType } from '@aws-cdk/cloud-assembly-schema';
```
jericht: nit: commented out code

jericht (dismissed):
LGTM! Thanks for doing this migration. Just found a couple bits of commented out code, but not holding back the PR for it

packages/aws-rfdk/lib/deadline/test/repository.test.ts
ddneilson: Drat. Will fix. Thanks!

packages/aws-rfdk/lib/deadline/test/spot-event-plugin-fleet.test.ts
ddneilson: Drat. Will fix. Thanks!

kozlove-aws (approved):

jericht (approved):

PR #751, opened 2022-07-31
chore(all): remove cdk-build-tools

ddneilson:
* Removes the `cdk-build-tools` from the RFDK repository in favour of using idiomatic tooling directly; this simplifies the maintenance burden going forward by eliminating a lot of unnecessary code. 
* Removes unnecessary rules from `pkglint`
* Increases the version of `eslint` in use as a bonus.

Testing: Full build (`yarn build`) and packaging of build artifacts (`pack.sh`). 

----


packages/aws-rfdk/package.json
```diff
@@ -47,19 +43,20 @@
...
+    "build": "jsii --project-references --silence-warnings=reserved-word && yarn run lint",
     "build+test": "yarn run build && yarn test",
     "build+test+package": "yarn run build+test && yarn run package",
     "clean": "tsc --build --clean && bash ./clean.sh",
-    "lint": "cdk-lint",
-    "lintfix": "cdk-lint --fix",
-    "package": "cdk-package",
-    "pkglint": "pkglint -f",
+    "lint": "eslint . --ext=.ts --resolve-plugins-relative-to=. && yarn run pkglint",
+    "lintfix": "eslint . --ext=.ts --resolve-plugins-relative-to=.",
```
jusiskin: I don't see any difference between the the `eslint` portions of these commands. Is there an additional flag we should be passing to `eslint` to make it fix the source code?

jusiskin (dismissed):
Looks good. I notice one potential bug in `lintfix` indicated in my comments, but it's not a blocker since it is developer-facing and can be fixed in a follow-up.

jericht (dismissed):
LGTM

packages/aws-rfdk/package.json
ddneilson: Fixed in revision. Thanks for spotting that!

jusiskin (approved):

jericht (approved):

Issue #807, opened 2022-09-09
(core): Rate limit errors deploying CloudWatch Log groups


ddneilson:
When the RFDK creates a CloudWatch Log group it does it via the CDK's [LogRetention ](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs.LogRetention.html) construct. The CloudWatch Logs APIs that the LogRetention construct calls have a very low TPS limit (5 TPS). The result is that when deploying multiple RFDK constructs one can encounter rate limit errors in the deployment -- this is particularly seen, randomly, with the RFDK's integration tests.

CDK's LogRetention construct provides a way to configure the Javascript SDK that it's using to make the CloudWatch API calls. Using this, the RFDK should set higher max retries and retry delay. 

Specifically, the RFDK should set the [logRetentionRetryOptions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs.LogRetention.html#logretentionretryoptions) option in these two places:
1. https://github.com/aws/aws-rfdk/blob/c3f25fac2982836e168714ff6bbe43f8622f0817/packages/aws-rfdk/lib/core/lib/log-group-factory.ts#L64
2. https://github.com/aws/aws-rfdk/blob/c3f25fac2982836e168714ff6bbe43f8622f0817/packages/aws-rfdk/lib/core/lib/exporting-log-group.ts#L114

Probably go for something really high for `maxRetries` (20?) and a delay base of 1s.

edit: The same option exists on Lambda Function constructs, and should be set there as well.

--- 

This is :bug: Bug Report

PR #822, opened 2022-09-20
fix(core): fix rate limit errors when deploying cloudwatch log groups 

xuetinp:
### Problem https://github.com/aws/aws-rfdk/issues/807
When the RFDK creates a CloudWatch Log group it does it via the CDK's [LogRetention ](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs.LogRetention.html)construct. The CloudWatch Logs APIs that the LogRetention construct calls have a very low TPS limit (5 TPS). The result is that when deploying multiple RFDK constructs one can encounter rate limit errors in the deployment -- this is particularly seen, randomly, with the RFDK's integration tests.

### Solution 

CDK's LogRetention construct provides a way - [logRetentionRetryOptions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_logs.LogRetention.html#logretentionretryoptions) to configure the Javascript SDK that it's using to make the CloudWatch API calls. Using this, the RFDK should set higher max retries and retry delay. Currently default backoff retry base is 100ms and max retry count is 3, proposed value is 200ms and 7. This is a good combination as it achieves a lot of retries while the total function runtime is not too long. 

- Added retry option to all places where new LogRetention is created. 
- Extended log retention function timeout. Referencing https://github.com/aws/aws-cdk/blob/205e493e7bd6c5212f0ae374fdee28128ea49afe/packages/%40aws-cdk/aws-logs/lib/log-retention.ts#L122-L130

### Testing
* Created a CloudFormation app that iteratively initialize LogRetention constructs. Wrote a script to deploy constructs in parallel to reproduce the rate limit error. Verified proposed values for max retry count and retry delay prevent the error. 
* Added unit test to verify template is modified as expected.
* Build success.

### DIFF
--- a/packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
+++ b/packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
@@ -17,6 +17,7 @@ import {
   Code,
   Runtime,
   SingletonFunction,
+  CfnFunction,
 } from 'aws-cdk-lib/aws-lambda';
 import {
   ILogGroup,
@@ -24,6 +25,7 @@ import {
   LogRetention,
   RetentionDays,
 } from 'aws-cdk-lib/aws-logs';
+import { Stack } from 'aws-cdk-lib/core';
 import { Construct } from 'constructs';
 
 /**
@@ -111,10 +113,25 @@ export class ExportingLogGroup extends Construct {
       threshold: 1,
     });
 
+    // Define log retention retry options to reduce the risk of the rate exceed error
+    // as the default create log group TPS is only 5. Make sure to set the timeout of log retention function
+    // to be greater than total retry time. That's because if the function that is used for a custom resource
+    // doesn't exit properly, it'd end up in retries and may take cloud formation an hour to realize that
+    // the custom resource failed.
     const logRetention = new LogRetention(this, 'LogRetention', {
       logGroupName: props.logGroupName,
       retention: retentionInDays,
+      logRetentionRetryOptions: {
+        base: Duration.millis(200),
+        maxRetries: 7,
+      },
     });
+    const logRetentionFunctionLogicalId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
+    const logRetentionFunction = Stack.of(this).node.tryFindChild(logRetentionFunctionLogicalId);
+    if (logRetentionFunction) {
+      const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
+      cfnFunction.addPropertyOverride('Timeout', 30);
+    }
 
     this.logGroup = LogGroup.fromLogGroupArn(scope, `${props.logGroupName}LogGroup`, logRetention.logGroupArn);
     this.logGroup.grant(exportLogsFunction, 'logs:CreateExportTask');
--- a/packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
+++ b/packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
@@ -126,12 +126,11 @@ export class ExportingLogGroup extends Construct {
         maxRetries: 7,
       },
     });
-    const logRetentionFunctionLogicalId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
-    const logRetentionFunction = Stack.of(this).node.tryFindChild(logRetentionFunctionLogicalId);
-    if (logRetentionFunction) {
-      const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
-      cfnFunction.addPropertyOverride('Timeout', 30);
-    }
+    // referenced from cdk code: https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L123
+    const logRetentionFunctionConstructId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
+    const logRetentionFunction = Stack.of(this).node.tryFindChild(logRetentionFunctionConstructId)!;
+    const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
+    cfnFunction.addPropertyOverride('Timeout', 30);
 
     this.logGroup = LogGroup.fromLogGroupArn(scope, `${props.logGroupName}LogGroup`, logRetention.logGroupArn);
     this.logGroup.grant(exportLogsFunction, 'logs:CreateExportTask');
--- a/packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
+++ b/packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
@@ -126,9 +126,9 @@ export class ExportingLogGroup extends Construct {
         maxRetries: 7,
       },
     });
-    // referenced from cdk code: https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L123
+    // referenced from cdk code: https://github.com/aws/aws-cdk/blob/v2.33.0/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L116
     const logRetentionFunctionConstructId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
-    const logRetentionFunction = Stack.of(this).node.tryFindChild(logRetentionFunctionConstructId)!;
+    const logRetentionFunction = Stack.of(this).node.findChild(logRetentionFunctionConstructId);
     const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
     cfnFunction.addPropertyOverride('Timeout', 30);
 
--- a/packages/aws-rfdk/lib/core/lib/log-group-factory.ts
+++ b/packages/aws-rfdk/lib/core/lib/log-group-factory.ts
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { Duration, Stack } from 'aws-cdk-lib';
+import { CfnFunction } from 'aws-cdk-lib/aws-lambda';
 import {
   ILogGroup,
   LogGroup,
@@ -52,7 +54,12 @@ export class LogGroupFactory {
     const fullLogGroupName = props?.logGroupPrefix ? `${props.logGroupPrefix}${logGroupName}` : logGroupName;
     const retention = props?.retention ? props.retention : LogGroupFactory.DEFAULT_LOG_RETENTION;
 
-    return props?.bucketName
+    // Define log retention retry options to reduce the risk of the rate exceed error
+    // as the default create log group TPS is only 5. Make sure to set the timeout of log retention function
+    // to be greater than total retry time. That's because if the function that is used for a custom resource
+    // doesn't exit properly, it'd end up in retries and may take cloud formation an hour to realize that
+    // the custom resource failed.
+    const logGroup = props?.bucketName
       ? new ExportingLogGroup(scope, logWrapperId, {
         bucketName: props.bucketName,
         logGroupName: fullLogGroupName,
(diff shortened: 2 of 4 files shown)


packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
```diff
@@ -111,10 +113,25 @@ export class ExportingLogGroup extends Construct {
...
+    // as the default create log group TPS is only 5. Make sure to set the timeout of log retention function
+    // to be greater than total retry time. That's because if the function that is used for a custom resource
+    // doesn't exit properly, it'd end up in retries and may take cloud formation an hour to realize that
+    // the custom resource failed.
     const logRetention = new LogRetention(this, 'LogRetention', {
       logGroupName: props.logGroupName,
       retention: retentionInDays,
+      logRetentionRetryOptions: {
+        base: Duration.millis(200),
+        maxRetries: 7,
```
ddneilson, line 126: Checking the math... 

I think that the backoff code in the JS SDK is [here](https://github.com/aws/aws-sdk-js/blob/aa6049093165b84eac975aa50c3a44e23cbfe4af/lib/util.js#L866) ( called from [here](https://github.com/aws/aws-sdk-js/blob/aa6049093165b84eac975aa50c3a44e23cbfe4af/lib/service.js#L551) amongst other places ).

Looks like the total amount of delay time over all retries will be `(2^0 + 2^1 + ... + 2^(maxRetries-1} ) * 200ms`. With maxRetries of 6 that puts our max delay at 25.2s. So, we should be good with these settings and increasing the Lambda Function's timeout to 30s.

ddneilson (dismissed):
This is perfect. Thank you!

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
```diff
@@ -111,10 +113,25 @@ export class ExportingLogGroup extends Construct {
...
+      },
     });
+    const logRetentionFunctionLogicalId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
```
jericht: Looks like we've grabbed this from the CDK code linked in the PR description. Could we also add a short comment here saying this is taken directly from the underlying CDK code as well?

packages/aws-rfdk/lib/core/lib/log-group-factory.ts
```diff
@@ -64,7 +71,20 @@ export class LogGroupFactory {
         new LogRetention(scope, logWrapperId, {
           logGroupName: fullLogGroupName,
           retention,
+          logRetentionRetryOptions: {
+            base: Duration.millis(200),
+            maxRetries: 7,
+          },
         }).logGroupArn);
+
+    const logRetentionFunctionLogicalId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
```
jericht: Same thing here, let's add a short comment saying this is taken directly from the CDK code

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
ddneilson: Good point. Breadcrumbs are useful...

xuetinp: Will do

packages/aws-rfdk/lib/core/lib/log-group-factory.ts
xuetinp: Will do

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
```diff
@@ -111,10 +113,25 @@ export class ExportingLogGroup extends Construct {
...
+    const logRetentionFunction = Stack.of(this).node.tryFindChild(logRetentionFunctionLogicalId);
+    if (logRetentionFunction) {
+      const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
+      cfnFunction.addPropertyOverride('Timeout', 30);
+    }
```
jusiskin: Should we fail if we cannot find the log function? I'm wondering about the case where we upgrade CDK to a new version and the construct path changes.

This would be a breaking change in CDK, but we might not catch the integration break unless the developer doing the CDK upgrade knows to check here. Having this throw an exception would be a more explicit failure mode.

Same feedback applies to `LogGroupFactory` below.

What do you think?

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
jusiskin: Strictly speaking, I don't believe this is a [logical ID](https://docs.aws.amazon.com/cdk/v2/guide/identifiers.html#identifiers_logical_ids). The [API docs](https://docs.aws.amazon.com/cdk/api/v2/docs/constructs.Node.html#trywbrfindwbrchildid) for `Node.tryFindChild(...)` document the argument with:

>   Identifier of direct child

And I believe this refers to the [construct ID](https://docs.aws.amazon.com/cdk/v2/guide/identifiers.html#identifiers_construct_ids) which is a unique ID within the scope of its parent in the [construct tree](https://docs.aws.amazon.com/cdk/v2/guide/constructs.html#constructs_tree).

TLDR: Can we rename this variable to `logRetentionFunctionConstructId`?

Same feedback applies to the similar change in the `LogGroupFactory` class below.

jusiskin (changes requested):
Great work. I have two minor suggestions to improve this.

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
xuetinp: Good callout! I will update.

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
xuetinp: I referenced the variable name from https://github.com/aws/aws-cdk/blob/205e493e7bd6c5212f0ae374fdee28128ea49afe/packages/%40aws-cdk/aws-logs/lib/log-retention.ts#L122-L130. Actually I think this id may be both logical id and construct id. It's logical id because in my testing I created different stacks, but they all have the same logical id for the function in the output cloudformation template. Meanwhile, it's also construct id for LogRetentionFunction.

xuetinp: Updated it to `logRetentionFunctionConstructId` to reduce confusion.

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
```diff
@@ -126,12 +126,11 @@ export class ExportingLogGroup extends Construct {
         maxRetries: 7,
       },
     });
-    const logRetentionFunctionLogicalId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
-    const logRetentionFunction = Stack.of(this).node.tryFindChild(logRetentionFunctionLogicalId);
-    if (logRetentionFunction) {
-      const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
-      cfnFunction.addPropertyOverride('Timeout', 30);
-    }
+    // referenced from cdk code: https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L123
```
jusiskin: nit: this URL is not pinned to a commit, so the line number may drift when CDK's default branch pointer is modified.

The best approach is to browse the git tag for the version RFDK currently targets (`2.33.0`) and link to that.

```suggestion
    // referenced from cdk code: https://github.com/aws/aws-cdk/blob/v2.33.0/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L116
```

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
```diff
@@ -111,10 +113,24 @@ export class ExportingLogGroup extends Construct {
...
       retention: retentionInDays,
+      logRetentionRetryOptions: {
+        base: Duration.millis(200),
+        maxRetries: 7,
+      },
     });
+    // referenced from cdk code: https://github.com/aws/aws-cdk/blob/main/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L123
+    const logRetentionFunctionConstructId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
+    const logRetentionFunction = Stack.of(this).node.tryFindChild(logRetentionFunctionConstructId)!;
+    const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
```
jusiskin: Just to confirm that this change was made in response to my previous suggestion where we'd throw an exception if we couldn't find the lambda function.

The belief here is that if the node without the specified construct ID does not exist, then `logRetentionFunction` will be `undefined` and line `132` will throw a:

```txt
Uncaught TypeError: Cannot read property 'node' of undefined
```

There may be an opportunity to make the error more developer-friendly since we have all the context here. How about this:

```suggestion
    const logRetentionFunction = Stack.of(this).node.tryFindChild(logRetentionFunctionConstructId);
    if (!logRetentionFunction) {
      throw new Error(`Could not find log retention Lambda Function with construct ID ${logRetentionFunctionConstructId} in construct tree`);
    }
    const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
```

Same feedback applies for `LogGroupFactory` below.

jusiskin (changes requested):

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
xuetinp: Will do

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
jusiskin: Ah, I see their code uses similar naming. Oddly, in CDK's docs on [logical IDs](https://docs.aws.amazon.com/cdk/v2/guide/identifiers.html#identifiers_logical_ids):

>   Unique IDs serve as the logical identifiers

The [docs for unique IDs](https://docs.aws.amazon.com/cdk/v2/guide/identifiers.html#identifiers_unique_ids) say that a node's unique ID is derived from its construct path:

>   AWS CDK generates the necessary unique identifiers by concatenating the elements of the path and adding an 8-digit hash

The [path docs](https://docs.aws.amazon.com/cdk/v2/guide/identifiers.html#identifiers_paths) state:

>   We refer to the collection of IDs from a given construct, its parent construct, its grandparent, and so on to the root of the construct tree, as a path.

The root of the construct tree is the app. The stack construct ID should be a part of the path, so based on my understanding of the docs, you should get different logical IDs for different stacks within the same app. \*shrug\*

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
xuetinp: The reason I didn't explicitly throw new error is that it will cause insufficient branch coverage and fail the build, and it's not easy to test that error branch as the log retention function is defined in cdk library. 

I tried using `findChild` function instead and it throws more explicit error. The code will be `const logRetentionFunction = Stack.of(this).node.findChild(logRetentionFunctionConstructId);` and error message will be like `Error: No child with id: 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8'`. I think this error message is a little better than undefined message above. What do you think if we switch to this?

jusiskin: Yeah, this combined with the stack trace will be sufficient context for the developer use-case I'd envisioned. Let's go with it.

packages/aws-rfdk/lib/core/lib/exporting-log-group.ts
xuetinp: Yes, the LogRetention resource that I created in my TestStacks has a different logical ID per stack. The logical id is concatenation of construct ids of its parent construct, its grandparent etc. But the LogRetentionFunction resource has a fixed logical ID that is not related to any parent constructs. I'm not sure why, but it's interesting to know.

jericht (approved):
LGTM!

ddneilson (approved):

jusiskin (approved):

PR #827, opened 2022-09-27
fix(integ): fix rate limit errors when deploying cloudwatch log group…

xuetinp:
### Problem #807 

This is a continuous fix following https://github.com/aws/aws-rfdk/pull/822. The previous fix partially solved the problem, and integration test still got random failures from other lambdas where LogRetention is defined implicitly without retries (e.g. [certificate function](https://github.com/aws/aws-rfdk/blob/c7e6cea813b77c9d8d75057f538f24782ce1fbe9/packages/aws-rfdk/lib/core/lib/x509-certificate.ts#L205)). 

### Solution

Add `LogRetentionRetryOptions` to all lambdas and extend log retention function timeout for all integration test stacks via [Aspect](https://docs.aws.amazon.com/cdk/v2/guide/aspects.html). Similar to [SSMInstancePolicyAspect](https://github.com/aws/aws-rfdk/search?q=SSMInstancePolicyAspect), it is only for integration test and not being added to example stacks for now. 

### Testing

- Tested Aspect code via a custom app deployment
- Build success

----

### DIFF
--- a/integ/lib/log-retention-retry-aspect.ts
+++ b/integ/lib/log-retention-retry-aspect.ts
@@ -1,0 +1,29 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Duration, IAspect, Stack } from 'aws-cdk-lib';
+import { CfnFunction } from 'aws-cdk-lib/aws-lambda';
+import { IConstruct } from 'constructs';
+
+export class LogRetentionRetryAspect implements IAspect {
+  public visit(node: IConstruct): void {
+    // Define log retention retry options to reduce the risk of the rate exceed error
+    // as the default create log group TPS is only 5. Make sure to set the timeout of log retention function
+    // to be greater than total retry time. That's because if the function that is used for a custom resource
+    // doesn't exit properly, it'd end up in retries and may take cloud formation an hour to realize that
+    // the custom resource failed.
+    if (node instanceof CfnFunction) {
+      node.addPropertyOverride('logRetentionRetryOptions', {
+        base: Duration.millis(200),
+        maxRetries: 7,
+      });
+    }
+    // referenced from cdk code: https://github.com/aws/aws-cdk/blob/v2.33.0/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L116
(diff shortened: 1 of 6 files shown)


integ/lib/log-retention-retry-aspect.ts
```diff
@@ -0,0 +1,29 @@
...
+      node.addPropertyOverride('logRetentionRetryOptions', {
+        base: Duration.millis(200),
+        maxRetries: 7,
+      });
+    }
+    // referenced from cdk code: https://github.com/aws/aws-cdk/blob/v2.33.0/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L116
+    const logRetentionFunctionConstructId = 'LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a';
+    const logRetentionFunction = Stack.of(node).node.findChild(logRetentionFunctionConstructId);
+    const cfnFunction = logRetentionFunction.node.defaultChild as CfnFunction;
+    cfnFunction.addPropertyOverride('Timeout', 30);
```
jusiskin: A few suggestions:

1.  From the [2.33.0 source code for `LogRetention`](https://github.com/aws/aws-cdk/blob/v2.33.0/packages/@aws-cdk/aws-logs/lib/log-retention.ts#L85-L97) it looks like it creates a `CfnResource` construct and not a `CfnFunction` for the `Custom::LogRetention` resource. Also, the resource CloudFormation property is named `SdkRetry` not `logRetentionRetryOptions` (which is actually the construct's input property name).
2.  I wonder if instead we should be looking for instances of [`LogRetention` constructs](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-logs.LogRetention.html) in the tree. Once we have a reference to that construct node, we can use CDK escape hatches to:
    1.  find the `CfnResource` child construct and override its properties.
    2.  find the stack singleton and override its timeout

What do you think?

jusiskin (changes requested):

integ/lib/log-retention-retry-aspect.ts
xuetinp: Thanks for the feedback! Will update.

jusiskin (approved):

integ/lib/log-retention-retry-aspect.ts
jusiskin: Thanks for applying this. Looks good now!

ddneilson (approved):
Perfect! Thanks!

PR #895, opened 2022-12-01
fix(deadline): use new repository installer log path for Deadline 10.2.*

AWS-Samuel:
## Problem

With the release of Deadline `10.2.0.9`, the deadline repository installer log file location changed from `/tmp/bitrock_installer.log` to `/tmp/installbuilder_installer.log`. 

This introduced a bug which made it so that the log stream in the repository component, meant to stream the repository installation logs, would no longer stream to cloudwatch.

## Solution

Add a conditional check in the repository component which changes the path of the installer logs depending on the deadline version.

## Testing

- Build Succeeds
- Modified unit tests pass
- Ran e2e tests on multiple versions

### E2E Testing Details

Ran the end to end tests in the `integ` folder by running

`yarn e2e` in the `aws-rfdk/integ` directory.

These end to end tests were run using both deadline `10.1.23.6` (max version before installer log change) and `10.2.0.9` (min version after installer log change).

When running for `10.1.23.6`
  - Verified that the cloud formation stack template had the following line
     ```
     "\",\"log_stream_name\":\"deadlineRepositoryInstallationLogs-{instance_id}\",\"file_path\":\"/tmp/bitrock_installer.log\",\"timezone\":\"Local\"}]}},\"log_stream_name\":\"DefaultLogStream-{instance_id}\",\"force_flush_interval\":15}}"
     ```
  - Verified that the Repository log group has both of the following streams:
    - `deadlineRepositoryInstallationLogs-<instance-id>`
    - `cloud-init-output-<instance-id>`
  - Verified that the `deadlineRepositoryInstallationLogs-<instance-id>` log stream was streaming installation logs
  
  When running for `10.2.0.9`
  - Verified that the cloud formation stack template had the following line
     ```
     "\",\"log_stream_name\":\"deadlineRepositoryInstallationLogs-{instance_id}\",\"file_path\":\"/tmp/installbuilder_installer.log\",\"timezone\":\"Local\"}]}},\"log_stream_name\":\"DefaultLogStream-{instance_id}\",\"force_flush_interval\":15}}"
     ```
  - Verified that the Repository log group has both of the following streams:
    - `deadlineRepositoryInstallationLogs-<instance-id>`
    - `cloud-init-output-<instance-id>`
  - Verified that the `deadlineRepositoryInstallationLogs-<instance-id>` log stream was streaming installation logs

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/version.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/version.ts
@@ -23,6 +23,11 @@ export class Version implements IPatchVersion {
   public static readonly MINIMUM_SECRETS_MANAGEMENT_VERSION = new Version([10, 1, 19, 0]);
 
   /**
+   * The minimum Deadline version which uses Install Builder instead of Bitrock Installer.
+   */
+  public static readonly MINIMUM_VERSION_USING_NEW_INSTALLBUILDER_LOG = new Version([10, 2, 0, 0]);
+
+  /**
    * This method parses the input string and returns the version object.
    *
    * @param version version string to parse
--- a/packages/aws-rfdk/lib/deadline/test/repository.test.ts
+++ b/packages/aws-rfdk/lib/deadline/test/repository.test.ts
@@ -72,12 +72,25 @@ let app: App;
 let stack: Stack;
 let vpc: IVpc;
 let version: IVersion;
+let installers: PlatformInstallers;
 
 function escapeTokenRegex(s: string): string {
   // A CDK Token looks like: ${Token[TOKEN.12]}
   // This contains the regex special characters: ., $, {, }, [, and ]
   // Escape those for use in a regex.
   return s.replace(/[.${}[\]]/g, '\\$&');
+}
+
+function create_version(version_array: number[]): IVersion {
+  class MockVersion extends Version implements IVersion {
+    readonly linuxInstallers: PlatformInstallers = installers;
+
+    public linuxFullVersionString() {
+      return this.toString();
+    }
+  }
+
+  return new MockVersion(version_array);
 }
 
 beforeEach(() => {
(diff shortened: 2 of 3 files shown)


ddneilson (dismissed):
Thanks for the fix, Sam!

packages/aws-rfdk/lib/deadline/lib/version.ts
```diff
@@ -22,6 +22,11 @@ export class Version implements IPatchVersion {
    */
   public static readonly MINIMUM_SECRETS_MANAGEMENT_VERSION = new Version([10, 1, 19, 0]);
 
+  /**
+   * The minimum Deadline version which uses Install Builder instead of Bitrock Installer.
+   */
+  public static readonly MINIMUM_VERSION_USING_INSTALLBUILDER = new Version([10, 2, 0, 0]);
```
jusiskin: nit: Bitrock is the company that makes InstallBuilder. Both versions of Deadline use InstallBuilder to make installers, but I suspect that the version of InstallBuilder was changed between Deadline releases and that InstallBuilder changed its (default?) log file path.

Can we instead rename to `MINIMUM_VERSION_USING_NEW_INSTALLBUILDER_LOG` and adjust the code comment accordingly?

packages/aws-rfdk/lib/deadline/test/repository.test.ts
```diff
@@ -424,7 +424,9 @@ test('repository installer logs all required files', () => {
           {}, // log group name. checked in another test.
           '\",\"log_stream_name\":\"cloud-init-output-{instance_id}\",\"file_path\":\"/var/log/cloud-init-output.log\",\"timezone\":\"Local\"},{\"log_group_name\":\"',
           {}, // log group name again.
-          '\",\"log_stream_name\":\"deadlineRepositoryInstallationLogs-{instance_id}\",\"file_path\":\"/tmp/bitrock_installer.log\",\"timezone\":\"Local\"}]}},\"log_stream_name\":\"DefaultLogStream-{instance_id}\",\"force_flush_interval\":15}}',
+          '\",\"log_stream_name\":\"deadlineRepositoryInstallationLogs-{instance_id}\",\"file_path\":\"/tmp/'+
+          (version.isLessThan(Version.MINIMUM_VERSION_USING_INSTALLBUILDER) ? 'bitrock' : 'installbuilder') +
```
jusiskin: I don't think we actually triggering both code paths here, so there's some missed coverage.

I think the value comes from 

https://github.com/aws/aws-rfdk/blob/60a1627376b256a18083fb0914906ec7e87fc2b4/packages/aws-rfdk/lib/deadline/test/repository.test.ts#L121

jusiskin (changes requested):
Great work, @AWS-Samuel!

Just two non-functional suggestions, so if you can incorporate them you shouldn't need to run integ tests again.

packages/aws-rfdk/lib/deadline/lib/version.ts
AWS-Samuel: Implemented in latest commit

packages/aws-rfdk/lib/deadline/test/repository.test.ts
AWS-Samuel: Good catch. I've done a bit of refactoring to support testing with multiple versions and can confirm that the test covers both cases now.

ddneilson (approved):

jusiskin (approved):

PR #899, opened 2022-12-15
chore: rename root project

ddneilson:
Just renaming the root package to match the internal package.json that we publish to reduce confusion by those not familiar with typescript `package.json` files.

----

### DIFF
--- a/scripts/rfdk_build_environment.sh
+++ b/scripts/rfdk_build_environment.sh
@@ -8,7 +8,7 @@
 # repository.
 
 # Make sure we're running from the root of the CDK repo
-if ! test -f package.json || ! grep '"name": "aws-rfdk-project",' package.json > /dev/null || ! grep '"private": true,' package.json > /dev/null
+if ! test -f package.json || ! grep '"name": "aws-rfdk",' package.json > /dev/null || ! grep '"private": true,' package.json > /dev/null
 then
     echo "Usage: Run from the root of the RFDK repository"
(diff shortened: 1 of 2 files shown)


scripts/rfdk_build_environment.sh
```diff
@@ -8,7 +8,7 @@
 # repository.
 
 # Make sure we're running from the root of the CDK repo
-if ! test -f package.json || ! grep '"name": "aws-rfdk-project",' package.json > /dev/null || ! grep '"private": true,' package.json > /dev/null
+if ! test -f package.json || ! grep '"name": "aws-rfdk",' package.json > /dev/null || ! grep '"private": true,' package.json > /dev/null
```
jusiskin, line 11: This check is not conclusive anymore (the comment says it checks that we are at the repo root). Do we care? What if we instead look for the `.git` directory as well?

jusiskin (changes requested):

scripts/rfdk_build_environment.sh
ddneilson: It's still conclusive due to the next check in the chain -- looking for `private: true`. That line does not exist in the `packages/aws-rfdk/package.json`

jusiskin: Aha, you're correct. All good then!

jusiskin (approved):

jericht (approved):

PR #917, opened 2023-01-03
chore: update copyright year to 2023 in license and notice files

crowecawcaw:
Previously, the RFDK builds were failing in a linting step because the notice and license files use an out of date year. The fix is to run `yarn pkglint -f` in a few packages which updated the current year in these files. The build now succeeds.

----


grbartel (approved):
Changes all look good to me.

jusiskin (approved):

PR #1075, opened 2023-08-07
chore(deps): bumping jsii from 1.77.0 to 5.1.9

edwards-aws:
Updated jsii and fixed issues that came up from that.

jsii got more strict about typing during compilation mostly about using errors in catches without first checking its type. 

To test this change I ran the ./build.sh script, I also ran yarn run build in each of the example directories to make sure those built as well. I also deployed ./examples/deadline/All-In-AWS-Infrastructure-Basic for TypeScript and Python and ensured that it was deployed correctly.

The new version of jsii also dropped support for Node 14. So I had to make changes to the integ tests to support Node >= 16. The main issues were related to npm changing the working directory to the directory with `package.json` in it. This broke some scripting that used relative paths.

To test these changes I ran the full integ test suite and ensure all tests passed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [18.x, 20.x]
 
     steps:
     - uses: actions/checkout@v3
--- a/integ/components/deadline/common/scripts/bash/deploy-utils.sh
+++ b/integ/components/deadline/common/scripts/bash/deploy-utils.sh
@@ -23,7 +23,7 @@ function deploy_component_stacks () {
   # (https://docs.aws.amazon.com/cdk/latest/guide/context.html#context_methods)
   # in our case this is for stack.availabilityZones
 
-  npx cdk synth &> "${INTEG_TEMP_DIR}/${COMPONENT_NAME}/synth.log"
+  npx cdk synth --output "$(pwd)/cdk.out" --app "npx ts-node --cwd $(pwd) bin/${COMPONENT_NAME}.ts" &> "${INTEG_TEMP_DIR}/${COMPONENT_NAME}/synth.log"
   SYNTH_EXIT_CODE=$?
   echo $SYNTH_EXIT_CODE > "${INTEG_TEMP_DIR}/${COMPONENT_NAME}/synth-exit-code"
   if [[ $SYNTH_EXIT_CODE -ne 0 ]]
@@ -42,7 +42,7 @@ function deploy_component_stacks () {
 
   for stack in $(cdk_stack_deploy_order); do
     echo "$(timestamp) [${COMPONENT_NAME}] -> [${stack}] stack deployment started"
-    npx cdk deploy --app cdk.out --require-approval=never -e "${stack}" &>> "${deploy_log_path}"
+    npx cdk deploy --concurrency 10 --app $(pwd)/cdk.out --require-approval=never -e "${stack}" &>> "${deploy_log_path}"
     STACK_DEPLOY_EXIT_CODE=$?
     if [[ $STACK_DEPLOY_EXIT_CODE -ne 0 ]]
     then
@@ -93,7 +93,7 @@ function destroy_component_stacks () {
   cp /dev/null "${destroy_log_path}"
   for stack in $(cdk_stack_destroy_order); do
     echo "$(timestamp) [${COMPONENT_NAME}] -> [${stack}] stack destroy started"
-    npx cdk destroy --app cdk.out -e -f "${stack}" &>> "${destroy_log_path}"
+    npx cdk destroy --app $(pwd)/cdk.out -e -f "${stack}" &>> "${destroy_log_path}"
     STACK_DESTROY_EXIT_CODE=$?
     if [[ $STACK_DESTROY_EXIT_CODE -ne 0 ]]
     then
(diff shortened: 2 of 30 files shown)


.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [18.x]
```
ddneilson: We should be adding node 20.x -- the current revision, and next LTS ( https://github.com/nodejs/release#release-schedule )

integ/components/deadline/deadline_05_secretsManagement/test/deadline_05_secretsManagement.test.ts
```diff
@@ -3,12 +3,13 @@
...
  */
 
+import { BlockList } from 'net';
 import { CloudFormation } from '@aws-sdk/client-cloudformation';
 import { SecretsManager } from '@aws-sdk/client-secrets-manager';
-import {
-  CidrBlock,
-  NetworkUtils,
-} from 'aws-cdk-lib/aws-ec2/lib/network-util';
+// import {
```
ddneilson: I guess we can remove these instead of comment-out?

ddneilson (changes requested):

.github/workflows/ci.yml
edwards-aws: Sure! Hopefully nothing breaks 🤞

integ/components/deadline/deadline_05_secretsManagement/test/deadline_05_secretsManagement.test.ts
edwards-aws: Ahhhh, thanks for catching that!

integ/components/deadline/common/scripts/bash/deploy-utils.sh
```diff
@@ -42,7 +42,7 @@ function deploy_component_stacks () {
 
   for stack in $(cdk_stack_deploy_order); do
     echo "$(timestamp) [${COMPONENT_NAME}] -> [${stack}] stack deployment started"
-    npx cdk deploy --app cdk.out --require-approval=never -e "${stack}" &>> "${deploy_log_path}"
+    npx cdk deploy --concurrency 10 --app $(pwd)/cdk.out --require-approval=never -e "${stack}" &>> "${deploy_log_path}"
```
jusiskin, line 45: I wonder if we want concurrency used by default. There is a similar `RUN_TESTS_IN_PARALLEL` configuration here

https://github.com/aws/aws-rfdk/blob/638e82086160a01e04c063e7b8f9368dadde2390/integ/test-config.sh#L63-L64

I wonder if we could introduce a `DEPLOY_TEST_STACKS_IN_PARALLEL` configuration for this.

The reason I push for this is because sometimes these tests can fail to deploy when there are many resources deploying in parallel. These are often because of API throttling and timeout/retries being exhausted.

integ/scripts/bash/deploy-infrastructure.sh
```diff
@@ -16,7 +16,7 @@ echo "$(date "+%Y-%m-%dT%H:%M:%S") [infrastructure] deployment started"
 set +e
 
 # Hide the deploy log unless something goes wrong (save the scrollback buffer)
-npx cdk deploy "*" --require-approval=never &> "${INTEG_TEMP_DIR}/infrastructure/deploy.txt"
+npx cdk deploy "*" --concurrency 10 --require-approval=never --output "$(pwd)/cdk.out" --app "npx ts-node --cwd $(pwd) bin/_infrastructure.ts" &> "${INTEG_TEMP_DIR}/infrastructure/deploy.txt"
```
jusiskin, line 19: Same as the other comment. Maybe concurrent deployment should be behind a configuration setting.

What do you think?

packages/aws-rfdk/lib/lambdas/nodejs/asg-attach-eni/test/test.ts
```diff
@@ -220,3 +220,49 @@ test('continues when complete lifecycle errors', async () => {
...
+          }),
+        },
+      },
+    ],
+  };
+
+  attachSpy = jest.fn( (request) => successRequestMock(request) );
+  mock('EC2', 'attachNetworkInterface', attachSpy);
+
+  JSON.parse = jest.fn( () => {throw 47;} );
```
jusiskin: Unless I'm missing something, I think we may not be restoring `JSON.parse` after the test. According to [this SO answer](https://stackoverflow.com/a/76292208), we should be able to do:


```suggestion
  jest.spyOn(JSON, 'parse').mockImplementationOnce(jest.fn( () => {throw 47;} ));
```

jusiskin (changes requested):
Great work here, @edwards-aws! Just a few minor suggestions

integ/components/deadline/common/scripts/bash/deploy-utils.sh
edwards-aws: I ran these tests a lot and never ran into any issues. I'd bias towards making the faster thing the default unless we run into issues.

packages/aws-rfdk/lib/lambdas/nodejs/asg-attach-eni/test/test.ts
edwards-aws: Gotcha, I'm not super familiar with jest. But this makes sense to me.

integ/components/deadline/common/scripts/bash/deploy-utils.sh
jusiskin: This is easy to change back, so I'm okay with doing this for now. I think each app deploys two stacks at most, so this should only become a big problem when running the tests in parallel.

jusiskin (approved):
Great job with this big migration and keeping the RFDK up-to-date. Thanks for all the effort here, @edwards-aws!

ddneilson (approved):

PR #1110, opened 2023-09-29
chore: use yarn to build in pipeline

edwards-aws:
Switched to using yarn for builds and for the pack command. This is needed for the upgrade to Node18.

Tested by running in the pipelines and making sure all steps passed.

### DIFF
--- a/buildspec.yaml
+++ b/buildspec.yaml
@@ -10,10 +10,10 @@ phases:
       - yarn install --frozen-lockfile
   build:
     commands:
-      - /bin/bash ./build.sh
+      - yarn build
   post_build:
     commands:
-      - "[ -f .BUILD_COMPLETED ] && /bin/bash ./pack.sh"
+      - "[ -f .BUILD_COMPLETED ] && yarn pack"
 artifacts:
   files:
     - "**/*"
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "private": true,
   "scripts": {
     "build": "./build.sh",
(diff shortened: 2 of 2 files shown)


jusiskin (approved):

crowecawcaw (approved):

epmog (approved):

buildspec.yaml
```diff
@@ -10,10 +10,10 @@ phases:
       - yarn install --frozen-lockfile
   build:
     commands:
-      - /bin/bash ./build.sh
+      - yarn build
```
ddneilson, line 13: I believe that this loses the [git secret scan](https://github.com/aws/aws-rfdk/blob/a875bc0804c74479e717aa3846c5d290b2ec416a/build.sh#L41C1-L41C32) that we have in the `build.sh`

ddneilson (changes requested):

buildspec.yaml
jusiskin: `yarn build` is an alias for `yarn run build` ([ref](https://classic.yarnpkg.com/lang/en/docs/cli/#toc-user-defined-scripts)). And our script just runs `build.sh`:

https://github.com/aws/aws-rfdk/blob/a875bc0804c74479e717aa3846c5d290b2ec416a/package.json#L17

The reason this change was made is because if we run the script directly, we don't get any of the environment setup that yarn does when running scripts.

ddneilson (approved):

PR #1155, opened 2023-12-13
feat(deadline): DocumentDB engine version support upgraded upto v5.0

sakshie95:
**Problem**
Only documentDB engine version 3.6 was supported before current Deadline release 10.3.1.3. With the MongoDB and DocumentDB version upgrades to 5.0 all versions (3.6, 4.0, 5.0) are supported with Deadline. We need to relax constraints on engine version 3.6 that exist in RFDK 

**Solution**
Upgraded engine version support to all versions prefix with 3.6, 4.0 or 5.0 as per the compatibility now. Updated relevant documentation in the code as well. Removed a Unit Test case testing the engine version constraint to 3.6. Since there are no other versions of DocumentDB available we need not add this test case with a different parameter.

**Testing**
1. yarn build:

```
aws-rfdk: =============================== Coverage summary ===============================
aws-rfdk: Statements   : 97.93% ( 3316/3386 )
aws-rfdk: Branches     : 95.08% ( 1045/1099 )
aws-rfdk: Functions    : 96.98% ( 514/530 )
aws-rfdk: Lines        : 98.09% ( 3196/3258 )
aws-rfdk: ================================================================================
aws-rfdk: Test Suites: 62 passed, 62 total
aws-rfdk: Tests:       1277 passed, 1277 total
aws-rfdk: Snapshots:   0 total
aws-rfdk: Time:        240.366 s
```

2. Deployed the All-In-Infrastructure-SEP stack in my AWS account with default engine version set to 3.6, confirmed it spun up and tore down successfully.

3. Manually updated the default version of DocDB to 5.0 and deployed the All-In-Infrastructure-SEP stack with the updated engine version to validate it works fine.

4. Ran the integration tests using my AWS account as well. Validated that they run fine.

```
Test Suites: 1 passed, 1 total
Tests:       21 passed, 21 total
Snapshots:   0 total
Time:        35.8 s, estimated 44 s
Ran all test suites matching /deadline_01_repository.test/i.
Test results written to: .e2etemp/deadline_01_repository/test-report.json
```

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/database-connection.ts
@@ -46,7 +46,7 @@ export interface DocDBConnectionOptions {
 
   /**
    * The Document DB Cluster this connection is for.
-   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6.
+   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6, 4.0, 5.0.
    */
   readonly database: IDatabaseCluster;
 
@@ -63,7 +63,7 @@ export interface DocDBConnectionOptions {
 export interface MongoDbInstanceConnectionOptions {
   /**
    * The MongoDB database to connect to.
-   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6.
+   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6, 4.0, 5.0.
    */
   readonly database: IMongoDb;
 
@@ -85,7 +85,7 @@ export interface MongoDbInstanceConnectionOptions {
 export abstract class DatabaseConnection {
   /**
    * Creates a DatabaseConnection which allows Deadline to connect to Amazon DocumentDB.
-   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6.
+   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6, 4.0, 5.0.
    *
    * Resources Deployed
    * ------------------------
@@ -97,7 +97,7 @@ export abstract class DatabaseConnection {
 
   /**
    * Creates a DatabaseConnection which allows Deadline to connect to MongoDB.
-   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6.
+   * Note: Deadline officially supports only databases that are compatible with MongoDB 3.6, 4.0, 5.0.
    *
    * Resources Deployed
    * ------------------------
@@ -188,7 +188,7 @@ class DocDBDatabaseConnection extends DatabaseConnection {
     super();
 
     if (!this.isCompatibleDocDBVersion()) {
-      Annotations.of(props.database).addError('engineVersion must be 3.6.0 to be compatible with Deadline');
+      Annotations.of(props.database).addError('engineVersion must be one of 3.6.0, 4.0.0 or 5.0.0 to be compatible with Deadline');
     }
 
     this.containerEnvironment = {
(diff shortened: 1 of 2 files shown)


packages/aws-rfdk/lib/deadline/lib/database-connection.ts
```diff
@@ -335,7 +335,7 @@ class DocDBDatabaseConnection extends DatabaseConnection {
     // checking the value of the engineVersion property of that object.
     if (this.props.database.node.defaultChild) {
       const cluster = this.props.database.node.defaultChild! as CfnDBCluster;
-      return cluster.engineVersion?.startsWith('3.6') ?? false;
+      return ( cluster.engineVersion?.startsWith('3.6') || cluster.engineVersion?.startsWith('4.0') || cluster.engineVersion?.startsWith('5.0') )?? false;
```
ddneilson, line 338: Should this compatibility be tied to Deadline version? Put another way, Deadline 10.3.1.3 [added support for DocumentDB 4/5](https://docs.thinkboxsoftware.com/products/deadline/10.3/1_User%20Manual/manual/release-notes.html#deadline-release); can version 10.3.0 or 10.2.0 use DocumentDB 4/5 successfully?

ddneilson:
Aside, we should also add the ability for the user to select their DocumentDB engine version when using the [defaults in the Repository construct](https://github.com/aws/aws-rfdk/blob/b84d37353663381c94986b3f0431f2fb02f992e9/packages/aws-rfdk/lib/deadline/lib/repository.ts#L680-L689).

jericht (commented):
I agree with Daniel's comments. Ready to approve once both of those are addressed.

packages/aws-rfdk/lib/deadline/lib/database-connection.ts
sakshie95: This is a great callout! While deadline versions 10.3.0 & 10.2.0 will support DocumentDB 4/5 since the drivers were updated with the 10.2.0 release, 10.1.0 will not since the drivers weren't updated then. However, we haven't been updating RFDK with the releases 10.2.0 & 10.3.0 and it's difficult to communicate this partial compatibilities with older versions. So, to make it easier for the customer to comprehend this change we can tie it to the latest version & beyond as per https://docs.thinkboxsoftware.com/products/deadline/10.3/1_User%20Manual/manual/release-notes.html#deadline-release.

sakshie95: Looked into the code and discussed this offline with @ddneilson. There isn't a direct way for us to tie this to the deadline version via code since that's not available to us at synthesis time. We should communicate this to the customers by adding a line to the Changelog that the version constraints are updated for the latest version of Deadline 10.3.1 and beyond. @marofke since you're owning this release, this line will need to be added manually to the Changelog, that should address this comment. We should double check with a PM for the wording.

sakshie95:
> Aside, we should also add the ability for the user to select their DocumentDB engine version when using the [defaults in the Repository construct](https://github.com/aws/aws-rfdk/blob/b84d37353663381c94986b3f0431f2fb02f992e9/packages/aws-rfdk/lib/deadline/lib/repository.ts#L680-L689).

Discussed this offline with @ddneilson. This is a feature that would need a data migration strategy for the customers to migrate their data from older version to a newer version which they can choose. Logged a backlog item for this

sakshie95:
> I agree with Daniel's comments. Ready to approve once both of those are addressed.

Addressed & resolved both comments!

jericht (approved):

ddneilson (approved):

PR #1165, opened 2023-12-20
test: set feature flag for unique launch template names to fix integ tests

marofke:
### Problem
Integration tests were failing to deploy some stacks due to Launch template name collisions. This is due to the IMSDv2 aspect requiring a feature flag for a launch template name to be unique.

### Solution
Set the feature flag in the `cdk.json` files for the integration tests

### Testing
Synthesized a simple CDK app with and without the feature flag set, confirming a LaunchTemplate in the template receive a unique name with the feature flag.

----

### DIFF
--- a/integ/components/deadline/deadline_01_repository/cdk.json
+++ b/integ/components/deadline/deadline_01_repository/cdk.json
@@ -1,5 +1,6 @@
 {
   "app": "npx ts-node bin/deadline_01_repository.ts",
   "context": {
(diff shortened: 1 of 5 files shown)


ddneilson (approved):

jusiskin (approved):
Looks good.

Only thing is the commit's category was `fix: ...` which should instead be `test: ...`. I've renamed the PR so that should become the squashed commit (if the person merging reloads the page since I've renamed it).

PR #1211, opened 2024-07-08
feat(deadline): Allow passing Context to SEP config

mgway:
This change allows the reserved spot fleet `context` field to be passed to the spot event plugin configuration custom resource. One caveat that was introduced is the launch template must specify an explicit version instead of `$Latest` because of the launch template parameter constraint `Context cannot be used with Launch Template version '$Latest' or '$Default'.`

Added a test for this new parameter and updated the test for the launch template version number. The documentation for these fields is intentionally vague as the CFN construct for spot fleets does not document this field https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata.html#cfn-ec2-spotfleet-spotfleetrequestconfigdata-context

----

### DIFF
--- a/packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
+++ b/packages/aws-rfdk/lib/deadline/lib/configure-spot-event-plugin.ts
@@ -568,6 +568,7 @@ export class ConfigureSpotEventPlugin extends Construct {
       ValidUntil: fleet.validUntil?.date.toISOString(),
       // Need to convert from IResolvable to bypass TypeScript
       TagSpecifications: (spotFleetRequestTagsToken as unknown) as SpotFleetTagSpecification[],
+      Context: fleet.context,
     };
 
     const spotFleetRequestConfigurations = fleet.deadlineGroups.map(group => {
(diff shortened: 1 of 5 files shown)


marofke:
Hi there, thanks for your submission! Sorry for the delayed response.

These changes look good to me, I'd just like a little more detail on your testing. Were you able to deploy your changes and confirm they worked as expected?

mgway:
I haven't yet tested these changes with rfdk 1.4.0 but a fork of 1.3.0 with these changes has been running a production farm for the past month without any ill effect and with the context string set in the spot requests as expected. The one thing I wasn't able to test was the behavior when an invalid context string is provided. However, I assume that it would work because configuring the spot fleet plugin via the UI will accept invalid context strings and only throw an error when the spot fleet request is eventually requested by the plugin.

marofke (approved):
Yeah that sounds reasonable to me, thank you for expanding on your testing, and for taking the time to put this change together!

We have a 2-person review process, so I'm having one of my colleagues to take a second look at this as well.

rondeau-aws (approved):

PR #1224, opened 2024-08-06
feat(aws-rfdk): update Lambda functions to use Node.js 18

rondeau-aws:
## Problem

The RFDK's Lambda functions use Node.js 16, which was deprecated on June 12, 2024.

## Solution

Update the Lambda functions to use Node.js 18.

Also:
- Previously, the Lambda functions used the AWS SDK for JavaScript v2. This is no longer included in the runtime environment for Node.js 18, so we updated the Lambdas to use the v3 SDK.
- Previously, we tested the Lambda's AWS SDK calls using `aws-sdk-mock`. However, this library doesn't support the AWS SDK for JavaScript v3, so we changed our tests to use `aws-sdk-client-mock` which does.
- The ExportingLogGroup Lambda was failing with errors such as "InvalidParameterException: Specified time range of '1723036552256' and '1722953812215' is not valid. Please make sure the values are within the retention period of the log groups and from value is lesser than the to value".  To fix this, I changed the implementation to calculate the time range using milliseconds instead of using `Date`.

## Testing

I tested the Lambdas by deploying the All-In-AWS-Infrastructure-Basic example, and by running the integration tests.  Here is a list of Node.js Lambdas that I found in this project, and how I tested them:

- LogExporterFunction
  - Tested using: All-In-AWS-Infrastructure-Basic
  - Not used by default, so I added a `logGroupProps` to `MongoDbInstance` and `RenderQueue`.
  - "Task has completed successfully!" in logs
- UnhealthyFleetAction
  - Tested using: All-In-AWS-Infrastructure-Basic
  - Created an unhealthy fleet by removing all inbound and output rules in the Worker's security group.
  - "Successfully suspended the fleet ..." in logs
- AcmImporter / x509-certificate.importCert
  - Tested using: deadline_05_secretsManagement
  - "SUCCESS" in logs
- mongodb.configureMongo
  - Tested using: deadline_01_repository
  - "SUCCESS" in logs
- DiskUsage / pad-efs-storage.getDiskUsage
  - Tested using: All-In-AWS-Infrastructure-Basic
  - called in concert with padFilesystem, which completed successfully
- PadFilesystem / pad-efs-storage.padFilesystem
  - Tested using: All-In-AWS-Infrastructure-Basic
  - "Adding 20GiB of files to the filesystem to grow size. " / "Writing 1024MiB to ..." and no errors in the logs
- AttachEniToInstance
  - Tested using: deadline_01_repository
  - Confirmed that the `StorageStruct3/MongoDB/Server` instance had a static ENI attached.  Terminated the instance, and confirmed that its replacement had the same static ENI attached.
- x509-certificate.generate
  - Tested using: deadline_05_secretsManagement
  - "SUCCESS" in logs
- x509-certificate.convert
  - Tested using: deadline_05_secretsManagement
  - "SUCCESS" in logs
- Configurator / configure-spot-event-plugin.configureSEP
  - Tested using: deadline_05_secretsManagement
  - "SUCCESS" in logs
- VersionProviderFunction / ecr-provider.handler
  - Tested using: All-In-AWS-Infrastructure-Basic
  - "SUCCESS" in logs
- VersionProviderFunction / version-provider.handler
  - Tested using: deadline_05_secretsManagement
  - "SUCCESS" in logs
- ECSWait / wait-for-stable-service.wait
  - Tested using: deadline_05_secretsManagement
  - "SUCCESS" in logs

I confirmed that all of the Lambdas listed above now use the Node.js 18 runtime.

Tested the lambda-layers by following the steps in the [README](https://github.com/aws/aws-rfdk/blob/mainline/lambda-layers/README.md).

----


loachri (dismissed):

lambda-layers/package.json
```diff
@@ -31,8 +31,5 @@
   "devDependencies": {
     "@types/node": "18.11.19",
     "typescript": "~5.1.6"
-  },
-  "dependencies": {
```
marofke, line 35: We don't need any Lambda SDK or anything for this package?  Just double checking we tested building and publishing as per https://github.com/aws/aws-rfdk/blob/mainline/lambda-layers/README.md

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/acm-handlers.ts
```diff
@@ -234,9 +251,9 @@ export class AcmCertificateImporter extends DynamoBackedCustomResource {
 /* istanbul ignore next */
 export async function importCert(event: CfnRequestEvent, context: LambdaContext): Promise<string> {
   const handler = new AcmCertificateImporter(
-    new ACM({ apiVersion: ACM_VERSION }),
```
marofke, line 237: Why'd we remove these specific versions? Are they no longer supported? Just wondering what the impact would be of this change

lambda-layers/package.json
rondeau-aws: Good catch, thanks!

I updated `lambda-layers` to work with the new v3 SDK, and I was able to run the steps in that README without any errors.

packages/aws-rfdk/lib/lambdas/nodejs/x509-certificate/acm-handlers.ts
rondeau-aws: Yes, `apiVersion` is no longer supported.  I'm having trouble finding a definitive reference for this, but see for example this ticket: [Remove apiVersion from configuration](https://github.com/aws/aws-sdk-js-v3/issues/1869).

Also, I ran the code through the [aws-sdk-js-codemod](https://github.com/aws/aws-sdk-js-codemod) tool, and it added annotations such as:

```
// The key apiVersion is no longer supported in v3, and can be removed.
// @deprecated The client uses the "latest" apiVersion.
apiVersion: '2014-11-06',
```

marofke (approved):
Thanks for the effort here Paul! Ship it! :ship: 🐋

ddneilson (approved):
Amazing effort sorting through all of those updates. Thank you Paul!

loachri (approved):

PR #1366, opened 2024-11-05
chore: Update github actions label to ubuntu 24

marofke:
## Summary
Github announced breaking changes to actions https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/

From what I can tell, the only change of interest is `ubuntu-latest` label being migrated to `ubuntu 24`. This draft PR will test that our actions work with the new label. For other changes in the announcement, we don't appear to be on github actions version 3, which is being removed so I think we're safe there.

----

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   auto-approve:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


marofke:
The CI has passed using the `ubuntu-24.04` label, which will be the new `ubuntu-latest`, so we should be good when the switch happens. Closing this PR.

PR #1389, opened 2024-11-22
feat(core): update to work with Deadline 10.4.0

rondeau-aws:
## Problem

We want to use the new Deadline 10.4.0 release with the RFDK.

Most of the issues we encountered are related to OS support in Deadline.  Previously we installed Deadline on Amazon Linux 2, but this is no longer supported in Deadline 10.4.0.

## Solution

- We updated the OS to Amazon Linux 2023 (AL2023).
- Previously our scripts used Python 2.  This is not installed on AL2023 by default, so we changed our scripts to use Python 3 instead.
- AL2023 includes a new version of GnuPG with different output.  We changed our invocation to use a machine-readable format that works on both the newer and older version.
- The `WF-submit-jobs-to-sets.sh` scripts in our integration tests submitted jobs that run `/usr/bin/sleep`.  This command is not available on Windows, causing occasional test failures.  We changed the tests to run `timeout.exe` on Windows instead.

## Testing

I ran the integration tests with Deadline 10.4.0.10 and confirmed that they succeed.

I also ran the integration tests on an older revision of this change, and confirmed that they succeeded with Deadline 10.3.2.1.

----

### DIFF
--- a/integ/lib/testing-tier.ts
+++ b/integ/lib/testing-tier.ts
@@ -7,7 +7,9 @@ import * as path from 'path';
 import { CfnOutput, Duration, Stack, StackProps } from 'aws-cdk-lib';
 import {
   BastionHostLinux,
+  IMachineImage,
   InstanceType,
+  MachineImage,
   Port,
   Vpc,
 } from 'aws-cdk-lib/aws-ec2';
@@ -16,7 +18,7 @@ import {
   SessionManagerHelper,
   X509CertificatePem,
 } from 'aws-rfdk';
-import { RenderQueue } from 'aws-rfdk/deadline';
+import { RenderQueue, Version } from 'aws-rfdk/deadline';
 import { Construct } from 'constructs';
 import { NetworkTier } from '../components/_infrastructure/lib/network-tier';
 import { IRenderFarmDb } from './storage-struct';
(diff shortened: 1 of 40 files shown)


integ/lib/testing-tier.ts
```diff
@@ -79,6 +81,7 @@ export abstract class TestingTier extends Stack {
       vpc: this.vpc,
       subnetSelection: { subnetGroupName: NetworkTier.subnetConfig.testRunner.name },
       instanceType: new InstanceType('t3.small'),
+      machineImage: this.getMachineImageForDeadlineVersion(this.deadlineVersion),
```
marofke: Just a question, but is this a change we just want to do in tests, or should we be putting this in the actual RFDK constructs as well?

rondeau-aws: Right, we should check to see if there are any constructs that I missed.

I was surprised that I didn't find any, but it makes some sense: we're using an AMI for the Worker, and containers for the RCS and License Forwarder, so none of those are running directly on the `machineImage`.

I think we'll need something similar when we update MongoDB though -- the v3.6 that we're running now doesn't work on Amazon Linux 2023.

marofke (dismissed):

marofke (approved):

ariemary-amz (approved):

PR #1392, opened 2024-12-03
feat!: Update MongoDB version to 8.0

marofke:
## Problem
The default version of MongoDB installed by RFDK is 3.6, which became end of life quite a long time ago https://www.mongodb.com/legal/support-policy/lifecycles

## Solution
Remove support for MongoDB 3.6 and add support for 8.0 by default (this is the latest version of MongoDB at this time). This required a few tweaks since the API changed. Also, I noticed the Node JS mongodb driver was quite old as well, so that's been bumped to the latest version.

Finally, I added `ap-southeast-5` to our denylist of regions for the lambda-layers package, as it's not enabled by default and is a new region since the last time we updated it.

## Testing
- ran integration tests, confirmed they all passed, also ran with the parallel flag set to true and confirmed they passed
  - I noticed the integration test directory wasn't being cleaned up properly, as I was left with a stale `staged` directory when running `yarn clean`. I've updated the cleanup script to remove all generated directories for integration tests (just run `yarn clean` from the root of the package)
- I did an upgrade test simulating an existing customer setup before these changes:
  - deployed the `All-In-AWS-Infrastructure-Basic` example with the current mainline changes
    - I did a `yarn clean && git clean -fxd` to ensure I had a clean environment
  - I then SSM'd onto the MongoDB instance and confirmed it was on version 3.6
  - I SSM'd onto the Worker instance and submitted a job:
    - `./deadlinecommand.exe -SubmitCommandLineJob -executable "/bin/sleep" -arguments "5" -frames 1-10 -priority 50 -name "Test Job2"`
    - confirmed it showed up with `./deadlinecommand.exe -GetJobs`
  - then, I checked out the changes from this PR and deployed them overtop my existing deployment
    - I did a `yarn clean && git clean -fxd` to ensure I had a clean environment
  - I SSM'd onto the DB instance and confirmed he existing Mongo 3.6 database was retained and it was not upgraded to 8.0. I confirmed my existing jobs were in the DB and I could submit new jobs from the Worker instance still
    - Users will have to manually update their MongoDB install (which will at least retain their data). There's a blog post about how one can do that here: https://awsthinkbox.zendesk.com/hc/en-us/articles/1500008016222-How-Do-I-Upgrade-MongoDB-to-version-4-2-from-3-2
- followed [the readme](https://github.com/aws/aws-rfdk/blob/mainline/lambda-layers/README.md) to publish lambda-layers to my account, confirming it succeeded

----


jgaudet-aws (dismissed):

examples/deadline/Local-Zone/ts/bin/config.ts
```diff
@@ -38,14 +38,14 @@ class AppConfig {
    * "10.1.16"
    * @default 10.3.2 is used, to match the worker AMI ID provided below
```
rondeau-aws:
```suggestion
   * @default 10.4.0 is used, to match the worker AMI ID provided below
```
Nitpick: looks like the "default" version in this comment should be updated from 10.3.2 to 10.4.0

rondeau-aws (approved):

jgaudet-aws (approved):

PR #1419, opened 2025-01-07
fix: Set Amazon Linux version based on Deadline Repository version

marofke:
## Problem
Recent changes to move RFDK infrastructure to Amazon Linux 2023 were causing integration tests to fail when installing the Deadline repository for Deadline versions before 10.4.0.10.

## Solution
When we install the Deadline Repository, we check the Deadline version, and fall back to AL2 instances if they are versions older than 10.4.0.

## Testing
- in the test-config.sh, set the `DEADLINE_VERSION="10.3.2.1"`
- ran `yarn e2e` from the integ directory, confirmed integration tests passed
- reverted test-config changes, reran integration tests, confirmed they passed with the latest Deadline version

----


ariemary-amz (approved):

gahyusuh (approved):
Thanks, looks good to me!

PR #1662, opened 2025-10-21
ci: scope down GitHub Token permissions

AdnaneKhan:
## Scope Down GitHub Token Permissions

This PR updates GitHub Actions workflows to use minimal required permissions instead of the default elevated permissions.

### Why This Matters

Following the principle of least privilege, workflows should only have the specific permissions they need to function.

### Changes

This PR adds explicit `permissions:` blocks to workflows that currently rely on default permissions, scoping them down to only what's required for their operations.

Please review the changes to ensure the specified permissions match your workflow requirements.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -9,6 +9,7 @@ on:
 
 permissions:
   issues: write
+  pull-requests: write
 
 jobs:
   modify-labels:
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -7,6 +7,9 @@ on:
     paths:
       - .github/config/labels.yml
 
+permissions:
+  issues: write
+
 jobs:
   modify-labels:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/labeler.yml
```diff
@@ -7,6 +7,9 @@ on:
     paths:
       - .github/config/labels.yml
 
+permissions:
```
marofke, line 10: I think we'll also need `pull-request:write` since our labels also need to be applied to PRs https://github.com/aws/aws-rfdk/blob/mainline/.github/config/labels.yml

marofke (dismissed):
Thank you for your submission! I just have one change I think we'll need to make to ensure we preserve our existing behaviour.

.github/workflows/labeler.yml
AdnaneKhan: done!

hdarren-aws (approved):

rondeau-aws (approved):

jgaudet-aws (approved):

PR #1714, opened 2026-07-24
feat: upgrade aws-cdk-lib to 2.262.1 and jsii toolchain to 5.9 (requires Node.js >= 20)

kenyrish-amazon:
### Why

1. `Runtime.NODEJS_22_X` does not exist in the pinned aws-cdk-lib 2.163.0; it is required to migrate the RFDK Lambda functions off the deprecated `nodejs18.x` runtime (follow-up PR).
2. aws-cdk-lib versions < 2.246.0 / < 2.260.0 are affected by two HIGH severity advisories (GHSA-999r-qq7v-r334, GHSA-vcrf-j523-4mrf — OS command injection in `NodejsFunction` bundling). RFDK does not use `NodejsFunction`, but our exact peerDependency pin pressures consumer applications below the fixed versions. This supersedes the dependabot PRs #1680, #1676, #1643, #1524.
3. The jsii 5.4 toolchain is end-of-support and cannot consume current aws-cdk-lib releases. This PR moves to jsii/jsii-rosetta 5.9, jsii-pacmak/jsii-reflect 1.139.0, and TypeScript 5.9.

### Notable changes

- `aws-cdk-lib` 2.163.0 → 2.262.1; `constructs` floor `^10.0.0` → `^10.5.0` (required by aws-cdk-lib >= 2.257), including the pkglint rule constant. `aws-cdk` CLI pin (integ) → 2.1132.0 (the CLI is versioned independently of the library now).
- `ImportedAcmCertificate` implements the `certificateRef` property required by `ICertificate` since the aws-cdk-lib resource-reference system; the implementation mirrors upstream `CertificateBase` (`{ certificateId: certificateArn }`).
- Test updates for CDK template drift: DynamoDB grants now synthesize table-stream actions as a separate IAM statement; KMS keys for SNS encryption gain a third key policy statement; `deployment-instance.test.ts` no longer mutates the construct tree after first synthesis (an error since aws-cdk-lib 2.167).
- `Buffer` → `Uint8Array` at AWS SDK call boundaries where SDK types tightened under TypeScript 5.9; test fakes updated for new interface members (`dbClusterRef`) and constructs 10.5.
- integ tsconfig: target/lib ES2022 (+`esnext.disposable`), `skipLibCheck` to match `packages/aws-rfdk`.

### Verification

- Full workspace build green: 1281/1281 unit tests across 62 suites; lint 0 errors; pkglint clean.
- Integration tests (scoped: repository, renderQueue, workerFleetHttps) run against a test sandbox account: all stack types deploy from 2.262.1-synthesized templates; the renderQueue functional suite passes 10/10 including TLS certificate generation.
- Two pre-existing issues were observed and excluded as unrelated (they reproduce identically on runs without this change): (1) 9 of 21 repository log-content tests fail due to an assertion expecting an `eventId` field that the CloudWatch Logs `GetLogEvents` API does not return; (2) worker fleet ASGs fail to launch the published basic worker AMI `ami-00667b691422ebb39` with a 403 (launch permission revoked/lapsed). Both will be filed as separate issues.

### Customer impact

The exact `aws-cdk-lib` peerDependency moves 2.163.0 → 2.262.1. Consumers must update the `aws-cdk-lib` dependency of their applications when adopting the RFDK release containing this change.

Additionally, aws-cdk-lib 2.262.1 declares `engines.node >= 20.0.0` (Node 18 is end-of-life), so building applications with this RFDK release requires Node.js >= 20. The CI matrix and `engines` fields have been updated accordingly.

### Decision for reviewers: dropping Node.js 18 build support

This release requires Node.js >= 20 to build RFDK applications. This is entailed, not optional: `aws-cdk-lib` declares `engines.node >= 20.0.0` as of 2.246.0, and **both HIGH security fixes are only available at >= 2.246.0** — a security-current CDK pin and Node 18 build support are mutually exclusive. Node.js 18 has been end-of-life upstream since April 2025. Deployed infrastructure (render farms, workers, Lambda functions) is unaffected; this only concerns environments that build/synthesize CDK apps. Python users are also affected via the jsii runtime's Node dependency.

**Pre-merge checklist item:** confirm the internal build/release pipeline and nightly canary CodeBuild environments run Node >= 20, or update them in lockstep with this merge.


kenyrish-amazon:
Superseded by #1716. Scaling this back to the minimal aws-cdk-lib 2.168.0 bump that unblocks the Node.js 22 runtime migration, without the jsii 5.9 toolchain upgrade (which would require a coordinated AWS-RFDK-pipeline change) or dropping Node.js 18 build support. The security-current CDK + jsii 5.9 upgrade will be tracked as a separate follow-up so it can be prioritized deliberately rather than blocking the runtime fix.

PR #1719, opened 2026-07-31
wip: migrate Lambda functions to nodejs22.x (openssl-al2023 layer)

kenyrish-amazon:
> [!WARNING]
> **DRAFT — DO NOT MERGE.** `packages/aws-rfdk/lib/lambdas/lambdaLayerVersionArns.ts` currently contains **placeholder ARNs** (assumed `224375009292` / `openssl-al2023` / version `:1`). These MUST be replaced with the real values from the production `openssl-al2023` layer publish (`node lambda-layers/bin/write-ts.js`) and verified before this PR is marked ready. This PR is opened early for visibility/CI while the production layer publish is arranged.

### Why

Completes the migration off the deprecated `nodejs18.x` Lambda runtime. Depends on two merged prerequisites: #1716 (aws-cdk-lib 2.168.0, which provides `Runtime.NODEJS_22_X`) and #1715 (the `openssl-al2023` layer definition).

### What

- Switch all **11 RFDK Lambda functions** from `Runtime.NODEJS_18_X` to `Runtime.NODEJS_22_X` (core + deadline).
- Point the three `openssl`-dependent constructs — `X509Certificate`, `MongoDbPostInstall`, `ImportedAcmCertificate` — at the **`openssl-al2023`** layer instead of `openssl-al2`. `nodejs22.x` is Amazon Linux 2023 based, on which the old `openssl-al2` binary fails (`symbol lookup error: SSL_get_srp_g, version OPENSSL_3.0.0`).
- Add the handler openssl-env wiring (`opensslProcessEnv`) that prepends `/opt/bin` and `/opt/lib` so the layer-provided openssl + its bundled libs take precedence over the AL2023 system openssl.
- Regenerate `lambdaLayerVersionArns.ts` for `openssl-al2023` (⚠️ placeholder — see warning above; this also drops the now-unused `openssl-al2` entries, which is intended since nothing references that layer after this change).
- Update unit-test assertions for the new runtime and layer name.

### Verification

- Full workspace build + **1281/1281 unit tests** pass (against the placeholder ARN map).
- **Pending before ready-for-review:** real ARNs from the production publish, and a live integration-test run on the `nodejs22.x` runtime.

### Consumer impact

RFDK's Lambda-backed custom resources move to `nodejs22.x`. No public API change.


kenyrish-amazon:
Closing this, we have merged this https://github.com/aws/aws-rfdk/pull/1722 to upgrade to nodejs 24.x

PR #1750, opened 2026-09-15
test(integ): fix DL-x-7 CloudWatch log-read flakiness

kenyrish-amazon:
### Problem
The `deadline_01_repository` tests `DL-x-5/6` (cloud-init-output) and `DL-x-7` (DeadlineRepositoryInstallationLogs) read a CloudWatch log stream **once** in `beforeAll`:
```ts
const data = await logs.getLogEvents(params);
logEvents = data.events!;
```
CloudWatch Logs ingestion is eventually consistent — the instance's log agent flushes asynchronously — so this one-shot read races ingestion and intermittently returns `[]`, producing `expect(logEvents).toContainEqual(...)` failures against an empty array.

### Impact
This is the **leading cause of recent mainline canary flakiness**. Confirmed instances: 2026-09-02 (`DL-1-7`) and 2026-09-04 (`DL-x-7`) prod-canary failures, plus repeated hits on branch canary runs. The same test previously failed for an unrelated reason (#1722), so this stream-read has a history of flaking.

### Fix
Add `common/functions/getLogEventsWithRetry.ts` — polls `getLogEvents` until a caller-supplied predicate (the expected log line is present) is satisfied, bounded to ~4 min (16 × 15s), which stays within the integ jest `testTimeout` of 5 min. Wire both `beforeAll` reads to it. When the content is present (the normal case) it returns on the first poll, so healthy runs are unaffected.

### Testing
- `integ` package builds clean (`tsc` + `pkglint` + `eslint`).
- Behavior verified against the failure signature (empty-array race) — the poll now waits for ingestion instead of asserting on an empty read.


integ/components/deadline/deadline_01_repository/test/deadline_01_repository.test.ts
```diff
@@ -242,8 +260,11 @@ describe.each(testCases)('Deadline Repository tests (%s)', (_, id) => {
           logGroupName: logGroupNames[id],
           logStreamName: cloudInitLogName,
         };
-        var data = await logs.getLogEvents(params);
-        logEvents = data.events!;
+        // CloudWatch ingestion is eventually consistent; poll until cloud-init output is present.
```
andrew-fenton: nit: these inline comments are redundant given you have a function-level comment for `getLogEventsWithRetry`

andrew-fenton (approved):

ttblanchard (approved):

jairaws (approved):

